PR #118, opened 2015-01-28
Authentication fix plus identity refactoring

Dantemss:
I might have gotten a bit carried away with the refactoring... but hopefully for the best?

We should refactor the SessionsCallback eventually. It's honestly still too complex.


spec/handlers/sessions_callback_spec.rb
```diff
@@ -4,237 +4,308 @@
...
-      it "transfers temp user auths to signed in user, destroys temp user, returns to app" do
-        result = SessionsCallback.handle(
-          user_state: user_state,
-          request: MockOmniauthRequest.new(authentication.provider, authentication.uid, [])
-        )        
-        expect(authentication.reload.user).to eq signed_in_user
-        expect(result.outputs[:status]).to eq :returning_user
-        expect(User.exists?(other_temp_user.id)).to be_false
+      context "with an email that matches existing emails" do
+        context "for one user" do
```
Dantemss: This test case should cover Ross's case. But honestly it's hard to even think about these tests.

app/models/identity.rb
```diff
@@ -1,21 +1,22 @@
...
   attr_accessible :password, :password_confirmation
 
-  # We need these validations because omniauth-identity does not
-  # provide them by default...
-  validates :password, presence: true,
-                       length: {minimum: 8, maximum: 40}
+  # We need these validations because
+  # omniauth-identity does not provide them by default
+  # These attributes are (obviously) not saved in the database
+  validates :password, presence: true, length: { minimum: 8 }
```
Dantemss: I can readd this if needed, but honestly there's no reason to limit password length I believe?

jpslav: fine to not have maximum on pw length

app/handlers/sessions_callback.rb
```diff
@@ -23,14 +23,14 @@ class SessionsCallback
...
+  uses_routine CreateUserFromOmniauthData
+  uses_routine AddEmailFromOmniauthData
   uses_routine DestroyUser
-  uses_routine TransferOmniauthInformation
 
   protected
 
   def setup
-    @auth_data = request.env['omniauth.auth']
+    @data = OmniauthData.new(request.env['omniauth.auth'])
```
jpslav: Yay :-)

app/handlers/sessions_callback.rb
```diff
@@ -39,11 +39,11 @@ def authorized?
   end
 
   def handle
-    # Get an authentication object for the incoming data, tracking if we have
+    # Get an authentication object for the incoming data, tracking if
     # the object didn't yet exist and we had to create it.
 
-    authentication_data = { provider: @auth_data[:provider],
-                            uid: @auth_data[:uid]}
+    authentication_data = { provider: @data.provider,
```
jpslav: actually don't need this `authentication_data` right?  Maybe we used to use it, but I don't see it below anywhere, so why not just pass these values into the `where` call?

app/handlers/sessions_callback.rb
```diff
@@ -102,24 +102,24 @@ def handle
...
-        sign_in(new_user)
+        sign_in!(new_user)
         status = :new_user
       end
 
     end
 
     if this_authentication_is_new
-      run(TransferOmniauthInformation, @auth_data, current_user)
+      run(AddEmailFromOmniauthData, @data, current_user)
```
jpslav: My first question was going to be "Can this just be `run(AddEmailToUser @data.email, current_user)`?", but...

Can we leave `TransferOmniauthInformation` so that later if we want to pull more info out we can?  It could be renamed `TransferOmniauthData` if that makes more sense w.r.t. other new names.

spec/models/authentication_spec.rb
```diff
@@ -6,13 +6,15 @@
 
   context "when an authentication exists" do
     it "is returned by by_provider_and_uid" do
-      value = Authentication.by_provider_and_uid(authentication.provider, authentication.uid)
+      value = Authentication.find_by_provider_and_uid(
```
jpslav: I always forget that these finders didn't get deprecated.

app/models/authentication.rb
```diff
@@ -4,12 +4,4 @@ class Authentication < ActiveRecord::Base
   belongs_to :user, inverse_of: :authentications
 
   validates_uniqueness_of :uid, scope: :provider
```
jpslav: I love me some tiny models!

app/models/contact_info.rb
```diff
@@ -12,7 +12,7 @@ class ContactInfo < ActiveRecord::Base
   validates :type, presence: true, inclusion: { in: VALID_TYPES }
   validates :value,
             presence: true,
-            uniqueness: {scope: [:type]}
+            uniqueness: {scope: [:user_id, :type]}
```
jpslav: Whoa is this really right?  Can two users really have the same contact info?

app/models/identity.rb
```diff
@@ -1,21 +1,22 @@
...
-  # provide them by default...
-  validates :password, presence: true,
-                       length: {minimum: 8, maximum: 40}
+  # We need these validations because
+  # omniauth-identity does not provide them by default
+  # These attributes are (obviously) not saved in the database
+  validates :password, presence: true, length: { minimum: 8, maximum: 50 }
   validates :password_confirmation, presence: true
 
-  before_save :update_password_expires_at, if: :password_changed?
```
jpslav: Not saying this is bad, but can you tell me where this functionality went?

app/models/identity.rb
```diff
@@ -23,68 +24,38 @@ def authenticate unencrypted_password
...
-  # Sets the password (and saves the identity)
-  def set_password!(password, password_confirmation)
-    # Always invalidate reset codes when the password changes
-    self.reset_code = nil
-    self.reset_code_expires_at = nil
-
+  # Sets the password and the password expiration date; Returns nil
+  def set_password(password,
+                   password_confirmation = nil,
+                   expiration_period = DEFAULT_PASSWORD_EXPIRATION_PERIOD)
```
jpslav: this parameter is unused

jpslav: actually since it is unused the comment for this method is not valid.

jpslav: nvm, confusing Github diff coloring.

app/models/identity.rb
```diff
@@ -23,68 +24,38 @@ def authenticate unencrypted_password
...
+    self.password_expires_at = \
+      expiration_period.nil? ? nil : DateTime.now + expiration_period
 
-  protected
-
-  def reset_code=(code)
-    super
-  end
+    # Invalidate reset code if it exists
+    reset_code.try(:destroy)
```
jpslav: is it weird that this method doesn't save this record but it does destroy one of the record's children?

app/models/identity.rb
```diff
@@ -23,68 +24,38 @@ def authenticate unencrypted_password
...
-  end
-
-  # Sets the password (and saves the identity)
-  def set_password!(password, password_confirmation)
-    # Always invalidate reset codes when the password changes
-    self.reset_code = nil
-    self.reset_code_expires_at = nil
-
+  # Sets the password and the password expiration date; Returns nil
+  def set_password(password,
```
jpslav: With the destroying of `reset_code` and this method's sole use from a handler, `set_password` is kind of looking like it should be a `SetPassword` routine?

app/models/identity.rb
```diff
@@ -23,68 +24,38 @@ def authenticate unencrypted_password
...
 
-  def reset_code_expires_at=(expiration_period)
-    super
+    nil
   end
 
-  def password_changed?
-    password_digest_changed?
+  # Expires the password
+  def expire_password
```
jpslav: This method is not used.

app/models/omniauth_data.rb
```diff
@@ -1,12 +1,48 @@
...
+  end
+
+  def last_name
+    @info[:last_name]
+  end
+
+  def email
+    # Facebook only returns verified emails
+    # Twitter returns no emails
+    # Google can return unverified emails
```
jpslav: interesting about google and the gem

app/routines/generate_reset_code.rb
```diff
@@ -0,0 +1,18 @@
+class GenerateResetCode
+
+  lev_routine
+
+  protected
+
+  def exec(identity, expiration_period = ResetCode::DEFAULT_EXPIRATION_PERIOD)
+    identity.reset_code ||= identity.build_reset_code
```
jpslav: Doesn't this code need to save the `identity` for when `reset_code` has to be built?  I see tests that save the `identity` but don't see saving in the real code

app/models/reset_code.rb
```diff
@@ -0,0 +1,23 @@
...
+  validates :identity_id, uniqueness: true
+  validates :code, presence: true, uniqueness: true
+
+  def generate(expiration_period = DEFAULT_EXPIRATION_PERIOD)
+    self.code = SecureRandom.hex(16)
+    self.expires_at = expiration_period.nil? ? nil : \
+                                               DateTime.now + expiration_period
+  end
+
+  # Returns true if the code is present in the DB and hasn't expired
```
jpslav: this comment talks about the `code` but the method doesn't do anything with `code`

app/routines/generate_reset_code.rb
```diff
@@ -0,0 +1,18 @@
+class GenerateResetCode
```
jpslav: You are right that you get a good refactoring by making `ResetCode` its own class, but I think this might be a reasonable example of when a routine is overkill, since only internal state of ResetCode is involved.  

Also, `ResetCode` reads like a routine, where "reset" sounds like a verb.  Since these are only used for password resets, can we rename it "password reset code"?

I'd be in favor of 

```
class Identity
  ...
  def reset_password_reset_code!(blah)
    self.build_password_reset_code && self.save if password_reset_code.nil?
    self.password_reset_code.reset!(blah)
  end
end

class PasswordResetCode
  ...
  def reset!(blah)
    # set values
    save
  end
end
```

and then a call to `identity.reset_password_reset_code!(blah)`.

That seem reasonable?

spec/handlers/sessions_callback_spec.rb
```diff
@@ -4,237 +4,308 @@
   
   let(:user_state) { MockUserState.new }
 
-  context "when not signed in and no existing auth" do
```
jpslav: Are these specs that were removed still accounted for in the tests?

spec/models/contact_info_spec.rb
```diff
@@ -40,28 +40,21 @@
   end
 
   context 'user emails' do
-    let!(:user1) { FactoryGirl.create(:user_with_emails, emails_count: 2) }
-    let!(:user2) { FactoryGirl.create(:user_with_emails, emails_count: 2) }
 
-    it "finds one user when there is one match" do
-      users = EmailAddress.where(:value => user1.contact_infos.first.value)
```
jpslav: did you take this out because it is weird to have a spec that tested a specific query?

spec/models/omniauth_data_spec.rb
```diff
@@ -0,0 +1,155 @@
+require 'spec_helper'
+
+describe OmniauthData do
```
jpslav: thanks for this

jpslav:
Done with review; I gave specs a cursory look

app/handlers/sessions_callback.rb
Dantemss: oh sure :)

app/handlers/sessions_callback.rb
Dantemss: Only reason I renamed it is that it was not clear what it did from the name. It really transfers only the emails, not anything else (username, first_name, last_name, full_name). But maybe there's a better name?

spec/models/authentication_spec.rb
Dantemss: Those are deprecated. In Rails 4 you'd use `find_by(provider: blah, uid: blah)`.

Still don't need to redefine them though.

app/models/contact_info.rb
Dantemss: They can according to our database schema...

Are you saying that they shouldn't? :)

app/models/identity.rb
Dantemss: Went into `set_password`.

app/models/identity.rb
Dantemss: Sure, I can do that.

app/models/identity.rb
Dantemss: Oh I intended for it to be used it in the users importer, but I forgot to change it.

I could remove it if you feel the users importer is fine just setting the expiration date manually.

app/models/omniauth_data.rb
Dantemss: Yep. Line 93 in https://github.com/zquestz/omniauth-google-oauth2/blob/master/lib/omniauth/strategies/google_oauth2.rb

app/routines/generate_reset_code.rb
Dantemss: The identity should already be persisted when you pass it to this routine. I guess I should add that to the comments.

Dantemss: The reset code contains the foreign key, so saving the identity is not needed (I set identity.reset_code just so the active record and the database are consistent without reloading the identity).

app/models/reset_code.rb
Dantemss: I should be more explicit here. The ResetCode I suppose.

app/routines/generate_reset_code.rb
Dantemss: Renaming it is fine.

But in this case the Identity class would be calling the reset method in `PasswordResetCode`, so it has to know about the methods in `PasswordResetCode`. How is this much better than calling `destroy`, which is a standard method for all ActiveRecord? :)

Dantemss: I feel `generate_reset_code` and `set_password` should either both be routines or methods, but not one routine and one method.

Dantemss: I've also come to dislike methods that automatically save the object they are called on...

When you have those, that's an extra query you cannot avoid at times.

spec/handlers/sessions_callback_spec.rb
Dantemss: Yes. I tried to nest them somewhat so we know which specs are related.

This particular spec ended up becoming the very first spec in the file after the change.

All the code got copied, however... and I found a feel untested cases that I marked as pending with `xit`.

spec/models/contact_info_spec.rb
Dantemss: Yeah, it's weird to test an ActiveRecord query... Assuming the SessionsCallback specs pass, this query should be tested anyway, since it's in that handler.

spec/models/omniauth_data_spec.rb
Dantemss: :D

Dantemss: Karen actually made this data, I just copied it from another test... But now I see she forgot to change a couple of lines... I'll change it.

app/handlers/sessions_callback.rb
jpslav: I'm ok with the name -- in the future it could transfer mobile phones too.  It just moves selected information from the resulting omniauth to the user.  Just so happens that the only selected information so far is email.

Dantemss: OK

app/models/contact_info.rb
jpslav: I don't think they should.  We have implicit assumptions that when we find matching emails during SessionsCallback that refers to the same user.  I'll ask the group at large, copying you.

app/models/identity.rb
jpslav: Either way.  If you want to use it elsewhere that's fine.  If you don't use it, remove it.

app/routines/generate_reset_code.rb
jpslav: duh

app/handlers/sessions_callback.rb
Dantemss: Actually it's used in the `where` and `create` methods so I guess this assignment avoids 1 duplication.

app/handlers/sessions_callback.rb
Dantemss: Actually, `TransferOmniauthData` is misleading. `TransferAuthentications` transfers `Authentication` from one user to another, but `TransferOmniauthData` does not transfer an `OmniauthData` at all. Rather, it gets some info from `OmniauthData` and adds it as a new `ContactInformation` for the `User`.

But I'll still do it if you want.

app/models/identity.rb
Dantemss: I suppose this is somewhat weird. Not a huge fan of methods that automatically save things, though. Maybe I should expire the reset_code instead?

app/models/contact_info.rb
Dantemss: OK don't merge until we have a decision on this :)

Dantemss:
OK I either fixed or asked for clarification on all of your comments. Waiting for the replies.

app/models/identity.rb
Dantemss: Asked for some input below, under a comment in `generate_reset_code`.

app/routines/generate_reset_code.rb
jpslav: Ok, do the renaming and go with routines instead of methods.

spec/models/contact_info_spec.rb
jpslav: also funny that this spec passed but where this code is used in SessionsCallback is where the original bug for this issue was :-)

app/models/identity.rb
Dantemss: Moved to the SetPassword routine now.

app/models/contact_info.rb
Dantemss: Seems like we have to keep this.

Dantemss:
@jpslav done with the changes you requested. Ready to merge I believe.

PR #122, opened 2015-02-02
Finish fixing double authentication bug; Fixes admin user search bug

Dantemss:
Turns out my first fix was incomplete. The routine still tries to add a duplicate email for Ross and errors out.

Also fixed undefined method call when trying to do admin user search.

Improved the test that was supposed to cover Ross's case (apparently it didn't actually work)

@edwoodward @reedstrm Maybe Karen could review this since JP is gone? Should be quick, it's pretty small. If not, we could ask LN.

Btw the reason only Ross found this bug is that he's the only one that has the same email on FB and google (and attempted to login with both)

### DIFF
--- a/app/controllers/terms_controller.rb
+++ b/app/controllers/terms_controller.rb
@@ -20,7 +20,7 @@ class TermsController < ApplicationController
 
   def show
     # Hide old agreements (should never get them)
-    raise ActiveRecord::RecordNotFound if !@contract.is_latest? && !current_user.is_admin?
+    raise ActiveRecord::RecordNotFound if !@contract.is_latest? && !current_user.is_administrator?
   end
 
   def pose
--- a/app/handlers/admin/users_search.rb
+++ b/app/handlers/admin/users_search.rb
@@ -15,7 +15,7 @@ module Admin
     protected
 
     def authorized?
-      !Rails.env.production? || caller.is_admin?
+      !Rails.env.production? || caller.is_administrator?
     end
 
     def handle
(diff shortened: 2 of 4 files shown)


lakshmivyas:
@Dantemss - I see a couple of more `is_admin` entries:

```
app/views/admin/users/_form.html.erb
18:        <td><%= f.check_box :is_admin %></td>

app/views/dev/base/_index.html.erb
29:            <%= f.check_box :is_admin %>
```

Are they okay?

lakshmivyas:
Rest of the changes good to me.  I will merge this soon.

Dantemss:
Thanks, not sure how I missed those. I'll make a new PR for these.

lakshmivyas:
@Dantemss - you can just add to this PR.

Dantemss:
@lakshmivyas actually I looked at those `is_admin` calls in the code last night and they shouldn't need to be changed. They seem unrelated to the ones I changed.

karenc:
Looks good to me.

jpslav:
@dantemss can those unless checks be inside the AddEmailToUser routine?  In particular this routine now marks addresses as already verified in normal use. If an email address is coming in here but already unverified in user, this call will be skipped and the addresses won't be verified.

Dantemss:
@jpslav I don't understand the concern... this routine is only called for verified emails coming from FB/google oauth and it has always marked them as verified.

Dantemss:
@jpslav oh I think I get it... if the user already has the email, but it's unverified, this routine should mark them as verified.

I'll fix after Sprint 6.

PR #130, opened 2015-02-17
improved gracefulness of adding emails to user (put logic inside that co...

jpslav:
...mmon code, handle verifiying of existing emails), and took out a typo line from key setup

@lakshmivyas I took out that line that appeared errant from your contribution. 

@Dantemss to review.

### DIFF
--- a/app/routines/add_email_to_user.rb
+++ b/app/routines/add_email_to_user.rb
@@ -28,6 +28,6 @@ class AddEmailToUser
     transfer_errors_from(email_address, { scope: :email_address }, true)
 
     # The confirmation info won't be sent if already verified
-    run(SendContactInfoConfirmation, email_address) unless errors?
+    run(SendContactInfoConfirmation, email_address)
   end
 end
--- a/app/routines/add_email_to_user.rb
+++ b/app/routines/add_email_to_user.rb
@@ -27,6 +27,7 @@ class AddEmailToUser
     email_address.save
     transfer_errors_from(email_address, { scope: :email_address }, true)
 
-    run(SendContactInfoConfirmation, email_address)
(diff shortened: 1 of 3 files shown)


app/routines/add_email_to_user.rb
```diff
@@ -7,13 +7,28 @@ class AddEmailToUser
...
+    # If it is a brand new email address, make it and send confirmation
+    if email_address.nil?
+      email_address = EmailAddress.new(value: email_address_text)
+      email_address.user = user
+      email_address.verified = options[:already_verified] || false
+
+      email_address.save
+      transfer_errors_from(email_address, { scope: :email_address }, true)
+
+      run(SendContactInfoConfirmation, email_address)
```
Dantemss: Don't wanna send this if already verified.

app/routines/transfer_omniauth_data.rb
```diff
@@ -11,8 +11,6 @@ def exec(data, user)
     raise Unexpected if data.provider == 'identity'
 
     run(AddEmailToUser, data.email, user, {already_verified: true}) \
```
Dantemss: Remove the \

Dantemss:
Couple of comments, otherwise looks good.

app/routines/add_email_to_user.rb
jpslav: actually SendContactInfoConfirmation doesn't send if already verified.  But that means I should be able to simplify this a bit.

app/routines/transfer_omniauth_data.rb
jpslav: good catch

PR #133, opened 2015-02-20
Add suffix field to users

karenc:
I have a few questions about this change:

- I've included it in the registration form.  Should I remove it?
- I've included the suffix in users' full name.  Is it ok?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -41,7 +41,7 @@ class UsersController < ApplicationController
   def user_params
     up = params[:user]
     return {} unless up.is_a? Hash
-    up = up.slice(:title, :first_name, :last_name)
+    up = up.slice(:title, :first_name, :last_name, :suffix)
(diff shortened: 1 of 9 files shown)


Dantemss:
- That is probably fine. Is the title there too? Maybe one day the registration page and profile page could be the same...
- Full name is weird. Have to ask JP
- My preference would be to shorten the first/last name fields and leave the button on the same line, but it's up to JP.

karenc:
@Dantemss No, I've only added suffix to the registration form.  I'll add title too.

karenc:
I've added title to the registration form.  I've also shortened the first name and last name field on the profile form so now it looks like this:

profile

I checked that the first name and last name fields are still about 20 characters wide on my browser.  So may be it's ok?

Dantemss:
I like it, but JP should weigh in.

jpslav:
Looks good.  Suffix in full name fine too.

jpslav:
Reviewed code too.  Looks good.

PR #135, opened 2015-02-24
WIP: add :exact parameter for querying users without wildcard selection

nathanstitt:


nathanstitt:
It seems to me that the core part of selecting by using wildcards is going to be running the SearchUsers routine with and without wildcard searching.  I've came up with several different ways of doing this and wanted to throw it out there for discussion.

app/routines/search_users.rb
```diff
@@ -36,33 +40,51 @@ class SearchUsers
...
+      end
+
+      with.default_keyword :any
+      with.keyword :username do |usernames|
+        users = users.where{ | q | compose_clause[q, q.username, usernames] }
       end
 
       with.keyword :last_name do |last_names|
-        users = users.where{lower(last_name).like_any my{prep_names(last_names)}}
+        users = users.where{lower(last_name).method(match_method).call my{prep_names(last_names)}}
```
nathanstitt: This uses the method(<like_any,any>).call

app/routines/search_users.rb
```diff
@@ -36,33 +40,51 @@ class SearchUsers
...
       end
 
       with.keyword :full_name do |full_names|
-        users = users.where{lower(full_name).like_any my{prep_names(full_names)}}
+        users = users.where{
+          options[:exact] ?
+            lower(full_name).any( my{prep_names(full_names)} ) :
+            lower(full_name).like_any( my{prep_names(full_names)} )
+        }
       end
```
nathanstitt: This is the "naive" implementation where it would just check the options on every block.  I think it would get pretty messy

app/routines/search_users.rb
```diff
@@ -36,33 +40,51 @@ class SearchUsers
...
+                       else
+                         lambda{ |q,field,names| q.lower(field).like_any prep_names(names) }
+                       end
 
       with.keyword :first_name do |first_names|
-        users = users.where{lower(first_name).like_any my{prep_names(first_names)}}
+         users = users.where{
+           lower(first_name).send(match_method, my{prep_names(first_names)})
+         }
+      end
```
nathanstitt: This uses send to dynamically call either :any or :like_any.  It's really the same as the method() call below

app/routines/search_users.rb
```diff
@@ -36,33 +40,51 @@ class SearchUsers
...
+
+      with.default_keyword :any
+      with.keyword :username do |usernames|
+        users = users.where{ | q | compose_clause[q, q.username, usernames] }
       end
```
nathanstitt: This is my preferred way of doing it, but thought it might not be immediately obvious what I was doing. It sets up a lamba that's assigned to the `compose_clause` variable and then calls it for the query.

app/routines/search_users.rb
```diff
@@ -36,33 +40,51 @@ class SearchUsers
   SORT_DESCENDING = 'DESC'
   MAX_MATCHING_USERS = 10
 
+  def matches(field)
+    debugger
```
Dantemss: Forgot to remove this :)

nathanstitt: Opps I did indeed.  Actually the whole method needs removed.  That was an earlier thing I was noodling on

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -135,6 +135,32 @@ def index
...
+      email addresses in ContactInfos and the password hash in Identity.
+
+      Users have the following String attributes:
+      username, first_name, last_name, full_name, title
+
+    #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
+  EOS
+  def username
+    OSU::AccessPolicy.require_action_allowed!(:search, current_api_user, User)
+    outputs = SearchUsers.call("username:#{params[:username]}", {exact: true}).outputs
```
jpslav: I think we should be using `ApplicationUser` here instead of `User`.  An `ApplicationUser` is a wrapper around `User` dedicated for the use of one of our applications.  

You can probably delegate some of the work to the `standard_read` method in `openstax_api` (https://github.com/openstax/openstax_api/blob/master/lib/openstax/api/roar.rb#L38-L40):

```
def username
  standard_read(ApplicationUser.find_by_username(params[:username]))
end
```

You'll have to make sure the access policy for `ApplicationUser` has an appropriate check for the `:read` action.

app/routines/search_users.rb
```diff
@@ -39,32 +43,37 @@ class SearchUsers
   def exec(query, options={})
```
jpslav: Since you won't be using `SearchUsers` in the endpoint controller method, we can leave this code out for now (maybe we'll pull it in at some point, but maybe not for a while.

nathanstitt: Yes, if I use standard_read I can leave the search_users alone and not overly complicate it

app/controllers/api/v1/application_users_controller.rb
nathanstitt: Thanks.  I didn't know about standard_read and didn't see any other way to query a User.

nathanstitt: Looks like I can't use standard_read, since accounts are at version 2.2.4.  That standard_read's signature is: `standard_read(model_klass, id, represent_with=nil)`, and it can only lookup using `model_klass.find`

It's only a few lines difference though to do the same work in the method, I'll go that route

jpslav: ok, note that @karenc is likely going to be updating openstax_api soon for Accounts, so we'll have the new version of `standard_read` available in short order.

nathanstitt:
Closing PR in favor of #137   I'm going to cleanup this branch and leave it around for awhile in case the `:exact` option for the SearchUsers routine is needed in the future.

PR #136, opened 2015-02-25
Change SendMessage routine to use Action Mailer

### LINKED ISSUE #127: Investigate email exceptions
Most likely the SendMessage routine is ignoring the SES settings because it uses the Mail gem directly instead of using a Rails mailer.

karenc:
The aws-ses gem doesn't support using Mail directly, there also seems to
be no way of initializing a class for Mail.delivery_method.

Add a new mailer called "ApiMailer" for sending messages on behalf of
registered apps using the messages API.

Fix #127 

This works locally with smtp.


jpslav:
Looks good; thanks @karenc!

Issue #134, opened 2015-02-24
Allow for exact matching on user search


pumazi:
This comes out of Connexions/cnx-publishing#83. We identify the users by username (partly because they are human readable and partly because that's what legacy uses.) As part of publishing, we make a shadow record of user-info in accounts (for search and representation of data). To do this, we search accounts by username (the only piece of unique data we have available at this time). However, it is currently possible to match a username within another username. For example, searching for username 'moo' would match 'moo', 'smoo', 'wolfsmoon', etc. We currently work around this by iterating through the results to exactly match the username.

We would like to possibly submit a new parameter that makes the search criteria match exactly. For example, searching for `?q=first_name:"Smoo Ooms"&exact=1` should match only records with that exact name. If this general case parameter `exact=<boolean>` can not be done, we at the very least would like something that can do an exact match on the `username` (maybe something like `exact-username:<string>`.

jpslav:
@Dantemss can you comment on this, and then maybe @nathanstitt can give it a shot?

Dantemss:
Either way should be doable, JP. Just depends on your preference. exact-username would be a new block for KeywordSearch, while the exact=(boolean) param would have to be passed to the blocks.

Dantemss:
Oh and actually I believe CNX wants to use the application_user search (not the user search).
User search searches all users, but is limited to (10 or 20, I forgot) results. If more than that would be returned, you get nothing.
ApplicationUser search can search any number of users that have been associated with CNX.

Maybe what we REALLY want is to implement the GET /application_user endpoint (commented out in the application_users_controller) and make it take a username instead of id (or either username or id).

pumazi:
I should probably mention that the use case here isn't using the application_user search. We are using the global search (/api/users.json), because the user isn't a member of this application. Publishing uses a different set of api keys than what authoring.

Dantemss:
Hm... sounds like it would be better to make users members of both apps, so you don't have the limitations of the global user search.

pumazi:
@Dantemss I'm sure it would be better to use the application_user search, but the most (if not all) CNX users will never sign into the publishing application. Therefore, if I understand correctly, they won't be registered as an application user, which means they won't show up in the search. Correct?

karenc:
I think #75 is meant to allow different cnx apps to use the same app id and secret?  That way publishing will be able to use the application user search.

karenc:
[...]

PR #137, opened 2015-02-26
Add /application_users/find/username/:username endpoint

nathanstitt:
Intended for looking up a user's information when
all that is known is the username.  Will perform an
exact match against the username, without using
sub-string search.

Issue #134

### DIFF
--- a/app/access_policies/application_user_access_policy.rb
+++ b/app/access_policies/application_user_access_policy.rb
@@ -5,8 +5,8 @@ class ApplicationUserAccessPolicy
     # Human users are not allowed
     return false if requestor.is_human?
 
-    # Apps can only call search, updates and updated
-    return [:search, :updates, :updated].include?(action)
+    # Apps can only call read, search, updates and updated
+    return [:read, :search, :updates, :updated].include?(action)
   end
 
 end
--- a/app/access_policies/application_user_access_policy.rb
+++ b/app/access_policies/application_user_access_policy.rb
@@ -4,9 +4,15 @@ class ApplicationUserAccessPolicy
   def self.action_allowed?(action, requestor, application_user)
     # Human users are not allowed
     return false if requestor.is_human?
-
-    # Apps can only call read, search, updates and updated
-    return [:read, :search, :updates, :updated].include?(action)
+    case action
+    when :read
+      # Applications can only read their own User records
+      return application_user.application == requestor
+    when :search, :updates, :updated # Apps can only call read, search, updates and updated
+      return true
+    else
+      return false
+    end
   end
 
 end
--- a/app/controllers/api/v1/application_users_controller.rb
+++ b/app/controllers/api/v1/application_users_controller.rb
@@ -95,10 +95,10 @@ class Api::V1::ApplicationUsersController < OpenStax::Api::V1::ApiController
   # find_by_username
   ###############################################################
 
-  api :GET, '/application_users/username/:username',
+  api :GET, '/application_users/find/username/:username',
       'Gets a single ApplicationUser with the specified username.'
   description <<-EOS
-    #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
+    #{json_schema(Api::V1::ApplicationUserRepresenter, include: :readable)}
   EOS
   def find_by_username
     application_user = ApplicationUser.includes(:user).joins(:user).where({
@@ -106,7 +106,7 @@ class Api::V1::ApplicationUsersController < OpenStax::Api::V1::ApiController
       :application_id => current_api_user.application.id
     }).first!
     OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, application_user)
-    respond_with [application_user], represent_with: Api::V1::ApplicationUsersRepresenter
+    respond_with application_user, represent_with: Api::V1::ApplicationUserRepresenter
   end
 
   ###############################################################
--- a/app/controllers/api/v1/application_users_controller.rb
+++ b/app/controllers/api/v1/application_users_controller.rb
@@ -92,30 +92,21 @@ class Api::V1::ApplicationUsersController < OpenStax::Api::V1::ApiController
   end
 
   ###############################################################
-  # username
+  # find_by_username
   ###############################################################
 
   api :GET, '/application_users/username/:username',
-      'Gets a single User with the specified username.'
+      'Gets a single ApplicationUser with the specified username.'
   description <<-EOS
-      All users of OpenStax have an associated User object.  This endpoint
-      allows querying additional data on a user when all that is known is the
-      User's username.
-
-      Admins (for Accounts only) are identified by the is_administrator boolean.
-      Some additional user information can be found in associations, such as
-      email addresses in ContactInfos and the password hash in Identity.
-
-      Users have the following String attributes:
-      username, first_name, last_name, full_name, title
-
     #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
   EOS
-  def username
-    OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, ApplicationUser)
-    username = params[:username]
-    respond_with ApplicationUser.includes(:user).joins(:user).where{user.username.eq username},
-                 represent_with: Api::V1::ApplicationUsersRepresenter
+  def find_by_username
+    application_user = ApplicationUser.includes(:user).joins(:user).where({
+      :user           => { :username => params[:username] },
+      :application_id => current_api_user.application.id
+    }).first!
+    OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, application_user)
+    respond_with [application_user], represent_with: Api::V1::ApplicationUsersRepresenter
   end
 
   ###############################################################
--- a/app/controllers/api/v1/application_users_controller.rb
+++ b/app/controllers/api/v1/application_users_controller.rb
@@ -1,6 +1,6 @@
 class Api::V1::ApplicationUsersController < OpenStax::Api::V1::ApiController
   #before_filter :get_app_user, :only => [:show, :update, :destroy]
-  
+
   resource_description do
     api_versions "v1"
     short_description 'Records which users interact with which applications, as well the users'' preferences for each app.'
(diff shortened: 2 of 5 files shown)


app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,34 @@ def index
...
+    OSU::AccessPolicy.require_action_allowed!(:search, current_api_user, ApplicationUser)
+    username = params[:username]
+    # N.B. references{:user} must be added to the query when/if upgraded to Rails4
+    respond_with ApplicationUser.includes{user}.where{user.username.eq username},
+                 represent_with: Api::V1::ApplicationUsersRepresenter
+  end
+
+  ###############################################################
   # show
   ###############################################################
```
nathanstitt: Note - I used `:search` for the AccessPolicy.  It [lacks :read](https://github.com/openstax/accounts/blob/master/app/access_policies/application_user_access_policy.rb#L9).   Perhaps `:read` should be added and used here?

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,34 @@ def index
...
+      Users have the following String attributes:
+      username, first_name, last_name, full_name, title
+
+    #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
+  EOS
+  def username
+    OSU::AccessPolicy.require_action_allowed!(:search, current_api_user, ApplicationUser)
+    username = params[:username]
+    # N.B. references{:user} must be added to the query when/if upgraded to Rails4
+    respond_with ApplicationUser.includes{user}.where{user.username.eq username},
```
Dantemss: I would do a .includes(:user).joins(:user) to force it to be an inner join (brackets work too).

Not a huge problem here I don't think, but Rails sometimes feels like splitting includes into multiple queries, which would cause an error, since the where clause references the users table.

Dantemss:
1 comment above, otherwise looks good to me.

app/controllers/api/v1/application_users_controller.rb
Dantemss: I agree with adding :read

app/controllers/api/v1/application_users_controller.rb
nathanstitt: Yeah, good call on the `.includes(:user).joins(:user)`.  I forgot that Rails 3 does that sometimes.  Rails 4 introduced "references" to get around that and I even noted that would be needed, I just forgot why ...

app/access_policies/application_user_access_policy.rb
```diff
@@ -5,8 +5,8 @@ def self.action_allowed?(action, requestor, application_user)
     # Human users are not allowed
     return false if requestor.is_human?
 
-    # Apps can only call search, updates and updated
-    return [:search, :updates, :updated].include?(action)
+    # Apps can only call read, search, updates and updated
+    return [:read, :search, :updates, :updated].include?(action)
```
jpslav: Applications should only allowed to read their own `ApplicationUser`s, so this access policy needs to enforce that requirement.  Means `:read` should be checking that the requesting application actually owns the `application_user`.

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,33 @@ def index
   end
 
   ###############################################################
+  # username
+  ###############################################################
+
+  api :GET, '/application_users/username/:username',
+      'Gets a single User with the specified username.'
```
jpslav: Should be "Gets a single ApplicationUser..."

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,33 @@ def index
...
+  description <<-EOS
+      All users of OpenStax have an associated User object.  This endpoint
```
jpslav: This description should be changed to reflect that this endpoint deals with ApplicationUsers, not Users (reading it it sounds like it is searching for and returning Users).  Fine to say that the response will include information in the User record if helpful to a developer.

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,33 @@ def index
...
+      allows querying additional data on a user when all that is known is the
+      User's username.
+
+      Admins (for Accounts only) are identified by the is_administrator boolean.
```
jpslav: This paragraph doesn't seem like it should be here (sounds like information about models in general in Accounts -- think this info is not needed here?)

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,33 @@ def index
...
+      Some additional user information can be found in associations, such as
+      email addresses in ContactInfos and the password hash in Identity.
+
+      Users have the following String attributes:
+      username, first_name, last_name, full_name, title
+
+    #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
+  EOS
+  def username
+    OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, ApplicationUser)
```
jpslav: To enforce that the current_api_user (hopefully an application) can actually read the ApplicationUser being retrieved, that ApplicationUser should be retrieved first and then passed in to the `require_action_allowed!`, e.g.

```
def username
  application_user = ApplicationUser.includes(:user).joins(:user).where{user.username.eq username}
  OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, application_user)
  respond_with application_user, represent_with: Api::V1::ApplicationUserRepresenter
end
```

You might even be able to use `standard_read` (then again, Account's version of openstax_api may be too old).

app/access_policies/application_user_access_policy.rb
Dantemss: Yes, good catch.

config/routes.rb
```diff
@@ -64,6 +64,7 @@
 
     resources :application_users, only: [:index] do
       collection do
+        get 'username/:username', action: 'username'
```
jpslav: I either recommended or OK'd this route, but I no longer have a good reaction to it :-)  I think I'd be happier with `/api/application_users/find/:query`, which is like the search route except that it is only for exact matches and `query` has to be something that can achieve an exact match (an ID, a UUID, a username).  

* `.../find/jpslav`
* `.../find/234`
* `.../find/a2ae1a0c-ca60-46c3-926e-cafd674afcf4`

I think we should be able to automatically determine the type of query by the content of the query. If it has hyphens it is a UUID (usernames cannot have hyphens).  If it starts with a number it is an ID (this would require us preventing usernames from starting with numbers --  is this feasible for legacy CNX users?  I'll check with CNX folks).  Otherwise it is a username.

We can't find by email since multiple users can have the same email address.

Thoughts @nathanstitt and @Dantemss?

spec/controllers/api/v1/application_users_controller_spec.rb
```diff
@@ -46,6 +45,29 @@
     end
   end
 
+  describe "username returns" do
+    it "a single result when username matches" do
+      api_get :username, untrusted_application_token, parameters: { username: 'foo_bb' }
+      expect(response.code).to eq('200')
```
jpslav: This should fail based on the above because `foo_bb` is not an application user for the `untrusted_application`.

config/routes.rb
Dantemss: That seems like an odd route to me... it looks more like you want 3 different routes.

spec/controllers/api/v1/application_users_controller_spec.rb
nathanstitt: Actually `foo_bb` does belong to `untrusted_application` :) Your point is well made though, that only succeeded though dumb luck since I wasn't aware of that issue.  I've added more specs to check that an application can't search for users that they don't have.

config/routes.rb
nathanstitt: I agree with @Dantemss.  I like the `../find/` syntax, but don't like that you can't discern between a `id/uid/username` search.  I've just implemented a `../find/username/jpslav` endpoint.  We can add additional finders later if/when we need them.

app/controllers/api/v1/application_users_controller.rb
nathanstitt: Looks like `standard_read` won't work since accounts is locked to an older 2.2.4 and it only accepts an id for lookup, not a query or model:  https://github.com/openstax/openstax_api/blob/v2.2.4/lib/openstax/api/roar.rb#L20

I'm wondering if we should return `access_denied` if they don't have permissions though?  That'd allow evil minded applications to determine that a user with that username does exist but they aren't allowed to see it.  

Wouldn't it be better just to return 404?  I've taken that route in the latest commit by using `find!` which will raise `ActiveRecord::RecordNotFound` and be converted to a 404.

app/controllers/api/v1/application_users_controller.rb
nathanstitt: Agree.  I've removed it and am referring to the model as the proper `ApplicationUser` now.  I think I was still attempting to figure out the nomenclature when I wrote that code. And probably still am :)

app/access_policies/application_user_access_policy.rb
nathanstitt: Yep, I've implemented that check now.  Should the other `:updates` and `:updated` perform the same check though?  That's probably outside the scope of this issue, but we might want to file another one if that's the case.

nathanstitt:
Travis is a lazy sod and failed because he took to long to install a gem.  The branch does pass all specs

app/controllers/api/v1/application_users_controller.rb
Dantemss: I like 404 here. Though honestly we can probably make all  
SecurityTransgressions give 404's in production somewhere else.

Dante

Quoting Nathan Stitt <notifications@github.com>:

app/access_policies/application_user_access_policy.rb
Dantemss: Updates and updated only return users for the current app (and they  
return multiple users) so the check does not apply.

Quoting Nathan Stitt <notifications@github.com>:

Dantemss:
I replied to a couple of comments, but they are in the outdated diffs part.

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,24 @@ def index
   end
 
   ###############################################################
+  # find_by_username
+  ###############################################################
+
+  api :GET, '/application_users/username/:username',
```
jpslav: This is missing `/find/` right?

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,24 @@ def index
...
+      'Gets a single ApplicationUser with the specified username.'
+  description <<-EOS
+    #{json_schema(Api::V1::UserSearchRepresenter, include: :readable)}
```
jpslav: Also this is the wrong representer.

app/controllers/api/v1/application_users_controller.rb
nathanstitt: It should indeed be.  I must have missed that when we added the `/find` prefix.

app/controllers/api/v1/application_users_controller.rb
```diff
@@ -92,6 +92,24 @@ def index
...
+  EOS
+  def find_by_username
+    application_user = ApplicationUser.includes(:user).joins(:user).where({
+      :user           => { :username => params[:username] },
+      :application_id => current_api_user.application.id
+    }).first!
+    OSU::AccessPolicy.require_action_allowed!(:read, current_api_user, application_user)
+    respond_with [application_user], represent_with: Api::V1::ApplicationUsersRepresenter
```
jpslav: Should be the singular representer `ApplicationUserRepresenter`, then also don't need to wrap the `application_user` in an array on the left there.

spec/controllers/api/v1/application_users_controller_spec.rb
```diff
@@ -46,6 +45,34 @@
...
+      expect(response.body).to eq(expected_response)
+    end
+    it "raises not found when when not found" do
+      expect {
+        api_get :find_by_username, untrusted_application_token, parameters: { username: 'foo' }
+      }.to raise_error(ActiveRecord::RecordNotFound)
+    end
+    it "only finds users belonging to the requesting application" do
+      expect {
+        api_get :find_by_username, trusted_application_token, parameters: { username: 'billy_20' }
```
jpslav: Can this test one application looking for another application's user?  "billy_20" doesn't exist at all.

nathanstitt: Actually it does, it's created as part of the "billy_users" block at the top of the spec.  I just picked the 20th at random.  https://github.com/openstax/accounts/blob/application_users-username-search/spec/controllers/api/v1/application_users_controller_spec.rb#L23-L33

I'm re-working the spec to be more explicit though since it's non-obvious what it's testing :)

PR #139, opened 2015-03-03
Fix oauth application creation rake task.

lakshmivyas:
- Fix email and subject params
- Ensure that objects are created only if they are not found.

\cc @jpslav, @Dantemss

### DIFF
--- a/lib/tasks/add_ost_apps.rake
+++ b/lib/tasks/add_ost_apps.rake
@@ -11,32 +11,35 @@ namespace :accounts do
     task :create_apps, [:app_domain_suffix, :admin_password] => :environment do |t, args|
       ActiveRecord::Base.transaction do
         begin
-          # Create application owner if needed
-          app_owner_group = Group.find_or_create_by_name('ost_app_admin_group')
-          user = User.find_or_create_by_username('ost_app_admin')
-          app_owner_group.add_owner(user)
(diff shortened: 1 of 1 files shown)


lib/tasks/add_ost_apps.rake
```diff
@@ -11,32 +11,35 @@ namespace :accounts do
...
           # Create application owner if needed
           app_owner_group = Group.find_or_create_by_name('ost_app_admin_group')
-          user = User.find_or_create_by_username('ost_app_admin')
-          app_owner_group.add_owner(user)
-          password = args[:admin_password]
-          user.identity = Identity.create do |identity|
-            identity.password = password
-            identity.password_confirmation = password
-            identity.user_id = user.id
+          user = User.find_or_create_by_username('ost_app_admin') do |new_user|
```
jpslav: Can you add some comments (a few sentences is fine) to the top just to give a quick summary of what it is this file is doing?  I was wondering why we were making an admin user in Accounts, and it took me a while to realize this user is just the owner of the apps being created.  And just a sentence to say what the overall purpose of this file is.  Thanks!

jpslav:
Looks good.  I'm going to merge so it is available sooner.  Can you add another PR to add the capitalization I commented on and to add some comments?

Issue #89, opened 2014-09-12
Allow users in other apps to create temporary users in accounts


Dantemss:
The API call will work like this:

- It requires a (full) name and an email address.
- It will check if the email address already exists.
  - If the email does not exist, a new user marked as temporary is created. Any user who can verify the email address can claim that account. When the account is claimed, the user is asked to authorize his apps again (unless the apps are trusted).
  - If the email does exist and the application is trusted, the app is given automatic access to that user's account.
- In both cases the user is returned by the API call.

Dantemss:
This issue will likely be solved when we implement the API that lets tutor admins create accounts for teachers/students. This can just be an extra option on that API.

nathanstitt:
Users have a `is_temp` field, which is set to true when a user is first created. I'm not entirely sure what the purpose of the field is, but it looks like a way to create users and then have them accept terms and conditions before they can access systems.

I see that it's consulted in the sessions callback to decide if the account should be deleted and it's information transferred to a real account in [sessions_callback.rb#L78-L87](https://github.com/openstax/accounts/blob/master/app/handlers/sessions_callback.rb#L78-L87) The [user_access_policy.rb#L8-L16](https://github.com/openstax/accounts/blob/master/app/access_policies/user_access_policy.rb#L8-L16) also disallows access by temp users.

What I'm currently unclear on is whether the `is_temp` attribute is sufficient for the purposes of this request.  Most likely it's semantics are subtly different in some way?

If the current `is_temp` won't get us what we need, we could also replace the attribute with a `state` field that would allow `:temp` and `:pending` values.  We could then add a `is_temp?` method to Accounts to continue to provide the current behavior.

Dantemss:
Seems done to me...

PR #140, opened 2015-03-03
Create unclaimed users from email address

nathanstitt:


app/controllers/api/v1/users_controller.rb
```diff
@@ -120,4 +120,22 @@ def update
...
 
-end
\ No newline at end of file
+  ###############################################################
+  # find_pending
+  ###############################################################
+
+  api :POST, '/pending/:email', 'Finds or creates a pending user by email address.'
+  description <<-EOS
+    Creates a pending user with the given ID.  Only the user's id is returned.
```
Dantemss: given email :)

Dantemss:
I like it a lot. Just a few thoughts:

- Not sure if pending and temp are the best names for the states. They can be confusing.
- We might want to return true for is_temp? when the user is pending (i.e. pending is a state before temp). Hopefully that will allow us to do things like deleting pending users without messing up the rest of the logic?
- Would be nice to have a few tests that say that pending users can access the registration page and register (I think that will fail right now, because that part of the site expects is_temp? == true

I personally would go with `unclaimed`, `claimed` and `registered` and with the is_claimed? and is_registered? methods. But before you rename anything ask for JP's input :)

nathanstitt:
Yes, I like the `unclaimed` name better than `pending`, it expresses the meaning of the state much better.

I'm using `temp` just to keep the behavior identical to the previous, and could go either way on `claimed` vs `temp`.  Claimed may be marginally better named, but there's quite a few places in the app that refer to a "temp" user.  If we replace them all with a "claimed" status, it may be confusing to other developers who are familiar with the codebase as it is.  I'm not sure how valid of a reason this is though, perhaps we're the only ones we need to worry about?

I also like `registered` vs `active`.  I can see where an account could be signed up but never do anything, in which case it's not really "active".

I'll add some specs for the registration page and fix it up so that the accounts can be claimed from there. I also think you're correct, right now it will fail.

@jpslav Would you like to chime in on the names @Dantemss outlined?  I'm thinking they all make sense except possibly the `temp` state and `is_temp?` helper method should stay.

jpslav:
We can change the term "temp" if we can find a better name for it.  Only
other person who knows this code is @karenc.  If we change it we'd also
want to change the description in the gdocs Accounts document (and in the
flow chart).  Don't know what the right replacement term is.  Maybe
`unfinished` since `temp` users later have their profiles "finished".

`registered` might not be the best.  we sometimes use "register" for the
last step (including signing up for terms).

`active` also might not be the best, as it more connotes recent activity.
Eh, I dunno, `active` may be ok.

`unclaimed` is a great term to replace `pending`.

karenc:
I'm not good with naming either...  Is `activated` more one off than `active`?

nathanstitt:
@karenc I think `activated` would be a good fit.  It would indicate that the system has completed all the checks to make an account active and that it can now be used, but wouldn't give the any indication that the account is being used frequently or at all.

I really wanted to suggest `valid` or `validated` for the final status, but that ran into problems of being too similar to Rails validations.

After thinking on it a bit more, I like `temp` better than `claimed`.  Using `claimed` seems to indicate that the account was previously `unclaimed`.   In most cases that won't be true, it's just a temporary account that was created in response to an oauth login and will later be merged into a different account and then deleted via the awesomely named `first_user_lives_second_user_dies` routine.  Or perhaps I'm reading the code incorrectly and most `temp` accounts do indeed start as `unclaimed` ?

Unless anyone else has further suggestions, I think I'll go with states:
  * `temp`  An account that's created by a login attempt and will be either merged into an existing account, or upgraded to `activated` status.
  * `unclaimed`  Account has been created by a different user as a placeholder for activity.  It may or may not be converted to a "real" account later.
  * `activated`  A "real" account that can be logged into and used.

Helper methods `is_activated?` and `is_temp?`.  The `is_activated?` can be used wherever the code needs to check if the account can be logged into and used.  The `is_temp?` check will stay the same as it is now.  

I don't think I'll need a `is_unclaimed?` method, since that check will only need to be in a few places and I can just check `if authentication_user.state == 'unclaimed'` in that spot.  If the scope grows for it, then I'll create a `is_unclaimed?` method as well.

nathanstitt:
@Dantemss When you're available this is ready for another review

app/controllers/api/v1/users_controller.rb
```diff
@@ -120,4 +120,24 @@ def update
     standard_update(User, current_human_user.id)
   end
 
-end
\ No newline at end of file
+  ###############################################################
+  # find_unclaimed
+  ###############################################################
+
+  api :POST, '/unclaimed/:email', 'Finds or creates a user by email address.'
```
jpslav: The name of this endpoint is a little funky to me, in that it doesn't necessarily return an unclaimed user.

Do we instead want `application_users/find/email/:email?force=true` or something to that effect? (`force=true` is maybe not the best)

cc @Dantemss

jpslav: And won't the calling app want an application user? (to fit into the rest of its logic?)

app/representers/api/v1/unclaimed_user_representer.rb
```diff
@@ -0,0 +1,13 @@
+module Api::V1
+  class UnclaimedUserRepresenter < Roar::Decorator
```
jpslav: Like above, kind of think we should just be reusing existing `ApplicationUserRepresenter`. (since I think apps always want to talk in terms of application users -- unclaimed ones will just be missing most information)

app/models/user.rb
```diff
@@ -2,6 +2,7 @@ class User < ActiveRecord::Base
 
   USERNAME_DISCARDED_CHAR_REGEX = /[^A-Za-z\d_]/
   USERNAME_MAX_LENGTH = 50
+  VALID_STATES = ['temp', 'unclaimed', 'activated']
```
jpslav: @Dantemss does the unclaimed status need any related changes in accounts-rails?

app/models/user.rb
```diff
@@ -63,6 +67,27 @@ def is_application?
...
 
+  # State helpers.
+  #
+  # A User model begins life in the "temp" state, and can then be claimed by another user
+  # who originated from an OAuth login. Upon it being claimed it will be removed and it's
+  # data merged with the claimant.
+  #
+  # A User can also be created by a one of the consumer applications as a stand-in
+  # for a person who has not yet (or may never) created an account.  In this case
+  # the User model will be in the "unclaimed" state.  When the User does signup, they
```
jpslav: Can someone point me to the code where an "unclaimed" user is claimed?

db/migrate/20150227185631_add_users_state_flag.rb
```diff
@@ -0,0 +1,15 @@
+class AddUsersStateFlag < ActiveRecord::Migration
+  def up
+    add_column :users, :state, :string, default: 'temp'
+    execute "update users set state='temp' where is_temp='t'"
```
jpslav: @Dantemss I am no sql guy - can you make sure to review this line for working in sqlite and postgres?

spec/controllers/users_controller_spec.rb
```diff
@@ -47,6 +47,19 @@
       expect(FinePrint.signed_contract?(temp_user, contract_1)).to eq true
       expect(FinePrint.signed_contract?(temp_user, contract_2)).to eq true
     end
+
+    it "claims an unclaimed account" do
+      user = FactoryGirl.create :user, state: 'unclaimed'
+      expect(user.state).to eq 'unclaimed'
+      controller.sign_in! user
```
jpslav: How does an unclaimed user sign in (they have no authentications)

app/models/user.rb
nathanstitt: It's actually in the finish user creation routine: https://github.com/openstax/accounts/blob/temp-users/app/routines/finish_user_creation.rb#L8

All that routine does is check if the user is already activated at the top and bail if it is.  The remainder of it sets the user to "activated" and will be called regardless of whether their "temp" or "unclaimed"

db/migrate/20150227185631_add_users_state_flag.rb
Dantemss: Should be fine. Postgresql accepts 't' as a true boolean value.
However, the issue here is that ALL users will be marked as temp (because temp is the default).
I think what we really want is update users set state='activated' where is_temp='f'

app/representers/api/v1/unclaimed_user_representer.rb
nathanstitt: My understanding was that all we should report back was the id for a user, so I didn't want to return the entire dataset.  You're correct though - it doesn't hurt to return a JSON blob that has the fields all empty.

db/migrate/20150227185631_add_users_state_flag.rb
nathanstitt: Not quite, "temp" is the default but existing columns won't be set to that, they'll just be null.  But we probably don't want that either, they should be "activated" as you've pointed out.  Will correct shortly.  Thanks Dante

nathanstitt:
As @jpslav pointed out in the last review, the accounts weren't actually being claimed anywhere.  The latest commits add a `MergeUnclaimedUsers` routine and calls it whenever an email is confirmed in the `ContactInfosConfirm` handler.  I also considered calling it from the `MarkContactInfoVerified` routine, but decided it was best to do so from a handler.  Please advise if there's a better place to call the merge routine from.

It's a bit of a misnomer to call it a `merge` at this point, since all the handler does is call the `DestroyUser` routine on matching unclaimed users.  I don't think there's any additional behavior that we should call at this point.

app/representers/api/v1/unclaimed_user_representer.rb
nathanstitt: @Dantemss @jpslav 
I've just got through testing the use of `ApplicationUserRepresenter` vs the custom `UnclaimedUserRepresenter`.   If I use `ApplicationUserRepresenter`, the JSON that's returned is:
`{"id":4,"user":{"id":59,"username":"user379650"},"unread_updates":1}`

The original spec [issue #89](https://github.com/openstax/accounts/issues/89) for this said all that should be returned was the openstax uid.  Since that response gives out the randomly generated username, I'm not sure if we should use it.  Please advise :)

app/models/user.rb
nathanstitt: The implementation is in https://github.com/openstax/accounts/blob/temp-users/app/routines/merge_unclaimed_users.rb  It's called from the `ContactInfosConfirm` handler https://github.com/openstax/accounts/blob/temp-users/app/handlers/contact_infos_confirm.rb#L18

Feature spec for it: https://github.com/openstax/accounts/blob/temp-users/spec/features/user_claims_account_spec.rb

spec/controllers/users_controller_spec.rb
nathanstitt: The account is claimed whenever a user confirms a matching email. At that point it's merged with their existing account

app/models/user.rb
Dantemss: It seems accounts-rails does not know about the is_temp field. Unless we want it to know, no change is needed.

db/migrate/20150227185631_add_users_state_flag.rb
Dantemss: I believe that column should be null: false btw.

app/controllers/api/v1/users_controller.rb
Dantemss: I agree with a more generic endpoint, like /users/find/email
I don't think this is an application_user, since that would mean the user has authorized the app to read their info (which they haven't).

app/representers/api/v1/unclaimed_user_representer.rb
Dantemss: If you have to reuse the existing ApplicationUserRepresenter, you will need to add `:if` conditions on most fields to prevent them from showing for certain states.

I don't think we want the ApplicationUserRepresenter, though, since there is no ApplicationUser until you authorize the app to read your info. If anything, this could be the UserRepresenter (with extra `:if`'s on most fields.

app/models/user.rb
Dantemss: You will have to make sure this routine is called when people sign in through Google/FB/etc. By doing that, the emails are automatically approved, so the ContactInfosConfirm handler is not used.

Maybe we could just remove those special cases from the code and make the ContactInfosConfirm handler always be called... (and at that point, I guess it becomes a routine?)

app/representers/api/v1/unclaimed_user_representer.rb
nathanstitt: I see your point about the user not explicitly authorizing the application, but don't we have to assume they're trusting the application?  We're allowing the application to create data for them before they've even signed up, and then transferring that data to their account once they confirm their email address.

If that's an issue, perhaps we need to re-think how the claiming process works and present a message to the newly signed up user saying something like "Application Foo has entered some information on your behalf, do you want to authorize Foo to do so?"

But then users may click "no" out of reflex, not understanding that they'll lose important data by doing so.

Dantemss: That might be true. @jpslav should we assume that if a site creates an account for you and you login on that account, then the site is allowed to read your info? Or should we require explicit authorization (i.e. logging into that specific site)

db/migrate/20150227185631_add_users_state_flag.rb
```diff
@@ -0,0 +1,20 @@
...
+    execute "update users set state='activated' where is_temp='f'"
+    # SQLite lacks support for the alter column statement
+    unless "SQLite" == ActiveRecord::Base.connection.adapter_name
+      execute "alter table users alter column state set not null"
+    end
+    remove_column :users, :is_temp
```
nathanstitt: @Dantemss That's about the best we can do to enforce non-null.  You can't use `null: false` in the add_column statement if the table has existing rows.  If you do so, it fails because the newly created column will start out as null. SQLite will allow you to, but PG will fail with: `ERROR:  column "state" contains null values`

PG lets you alter a column to set it to not null later though, which is what I'm doing here after it's been updated to have a value.

Dantemss: I think if you specify a default, though, you're fine? It should just default all rows to the default value.

app/controllers/api/v1/users_controller.rb
nathanstitt: Do we need a `/users/find` endpoint?  Since the purpose of this is to allow applications to create an account using an email, shouldn't we name it `/users/create/email/foo@test.com`  The endpoint would continue to be implemented in terms of `find-or-create` where it would only return an existing id and not create a duplicate.

nathanstitt:
@jpslav @Dantemss I've changed the endpoint to be `/users/create/email/:email` now and am checking that the user is unclaimed before returning it's id (if it pre-existed).  Thoughts on that url?

jpslav:
will sometimes exist already and therefore won't be created, right?

nathanstitt:
That's true.  But most of the time it will be create only.  The spec and our description doesn't mention using it to query.

Dantemss:
It's more like a find_or_create (but find_or_create is too long)...

nathanstitt:
Yeah,  it is but I'd say that's just as a convience  to the FE.   If they have a need for a query only call, I'd say we should create a matching GET.

Dantemss:
This will be called by Exercises/Tutor BE so more like a convenience for the BE too. :)

nathanstitt:
This is ready for another round of reviews

app/controllers/api/v1/users_controller.rb
```diff
@@ -120,4 +120,43 @@ def update
     standard_update(User, current_human_user.id)
   end
 
-end
\ No newline at end of file
+  ###############################################################
+  # find_or_create
+  ###############################################################
+
+  api :POST, '/find-or-create', 'Finds or Creates a pending user a user.'
```
jpslav: Typo: "a pending user a user".

app/controllers/api/v1/users_controller.rb
```diff
@@ -120,4 +120,43 @@ def update
...
+  description <<-EOS
+    Either finds or creates a new user a user. The new user will be created
```
jpslav: Same typo: "a new user a user".

Might be clearer as:

```
Either finds or creates a new user.  If a new user is created, its
state will be "unclaimed" meaning...
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -120,4 +120,43 @@ def update
...
+
+    If the username or email is already in use by an unclaimed user,
+    a user will not be created and the existing the user's ID is returned
+    #{json_schema(Api::V1::UnclaimedUserRepresenter, include: :readable)}
+  EOS
+  param :email, String, required: false,
+        desc: "Email address to search by or assign to newly created user"
+  param :username, String, required: false,
+        desc: "Username to search by or assign to newly created user"
+  param :password, String, required: false,
```
jpslav: Are these actual URL parameters instead of posted JSON data?  I don't think we can do that because that means these fields will be sent in the clear outside of the encryption offered by SSL.  We of course can't send credentials in the clear like that.

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    if 'unclaimed' == user.state
+      # If a username and password was given, set the unclaimed user's identity
+      if options[:username] && options[:password]
+        set_or_create_password(user, options)
+      end
+      outputs[:user] = user
+    else
+      fatal_error(code: :account_already_claimed, message: "Account has already been claimed")
```
jpslav: This routine can return non-unclaimed (previously existing and claimed) users, right?  Reading this `if` statement at line 32 it looks like any non-unclaimed user will result in a fatal error which I don't think should be the case.  If I'm right, please add a spec to highlight this issue and then update the code.

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def exec(options)
+    user = nil
+
+    if options[:username]
+      user = find_or_create_by_username(options)
+    elsif options[:email]
+      user = find_or_create_by_email(options)
+    else
+      fatal_error(code: :invalid_input, message: "Must provide either email or username")
```
jpslav: The endpoint description says an email is always required, which disagrees with this error message.

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+      return
+    end
+
+    if 'unclaimed' == user.state
+      # If a username and password was given, set the unclaimed user's identity
+      if options[:username] && options[:password]
```
jpslav: Maybe there should be a `create_unclaimed_user(options)` private method in this routine that can handle the setting of the identity (instead of this code being disconnected from the user creation code down below.

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+      outputs[:user] = user
+    else
+      fatal_error(code: :account_already_claimed, message: "Account has already been claimed")
+      return
+    end
+  end
+
+
+  def set_or_create_password(user, options)
+    if user.identity
```
jpslav: If a user already has an identity, is this code resetting their password?  Should an application really have a right to do that?  I tend to think not.  But if an application sends password information, we'd need a way to tell the application that the password was not changed.

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+      user.reload # is needed in order to notice the newly created identity
+    end
+  end
+
+  def find_or_create_by_email(options)
+    email = EmailAddress.with_users.where(value: options[:email]).first
+    if email
+      return email.user
+    else
+      user = run(CreateUser,
```
jpslav: replace this block and the one very much like it below with a call to `create_unclaimed_user(options)`.  `AddEmailToUser` already deals gracefully with the case where the email is blank, so the code can be identical:

```ruby
def create_unclaimed_user(options)
  user = run(CreateUser,
             state: 'unclaimed', username: options[:username],
             ensure_no_errors: true).outputs.user
       
  run(AddEmailToUser, options[:email], user)
  user
end
```

app/models/user.rb
jpslav: @nathanstitt did you address this case?  If not, can you add a spec (maybe in the sessions callback spec, which already has some Google/FB stuff) to show that this fails and then fix?

app/representers/api/v1/unclaimed_user_representer.rb
jpslav: @nathanstitt I think this whole thread has been addressed right? (i.e. we're just returning an ID now so there's no data going back to the application?)

app/routines/find_or_create_unclaimed_user.rb
```diff
@@ -0,0 +1,90 @@
...
+                 state: 'unclaimed', username: options[:username],
+                 ensure_no_errors: true).outputs.user
+      run(AddEmailToUser, options[:email], user)
```
jpslav: Calling `AddEmailToUser` will send a confirmation message -- but the human attached to that email won't really be expecting it will he?   Seems like we should either not send a confirmation or we should be sending an email that says "hey someone invited you to join OpenStax blah, click here to create your account / sign up / something".

Sorry this is so complex @nathanstitt  :-)

app/routines/add_email_to_user.rb
```diff
@@ -27,7 +27,12 @@ def exec(email_address_text, user, options={})
     email_address.save
     transfer_errors_from(email_address, { scope: :email_address }, true)
 
-    # The confirmation info won't be sent if already verified
-    run(SendContactInfoConfirmation, email_address)
+    # Unclaimed accounts only receive a welcome email
+    if user.is_unclaimed?
+      UnclaimedUserMailer.welcome(email_address).deliver
```
jpslav: This is great and thanks for adding this.  Now the trick is that these user's email addresses will never be confirmed!  Can you think it through and figure out the best way to confirm the email address?

Haven't thought this all the way through, but maybe the best way is just use the contact info confirmation mailer but change the text for the unclaimed case (to have the invitation language you wrote but next to a link that confirms their email address and then drops them off in account finalization or something).  Thoughts?

nathanstitt:
And... it expands a bit more.

The latest commit changes the `UnclaimedUserMailer` to send a message asking the unclaimed account's email to visit `/confirm/unclaimed`, That confirm handler will then mark the email as confirmed and display a  page welcoming the user and asking them to visit either `/login` for users with a password, or '/register` for users who do not.

jpslav:
@nathanstitt is the man.

PR #144, opened 2015-03-06
Fix user creation error when sanitized downcased username is already taken

### LINKED ISSUE #143: Sign up is broken if username lowercased is taken
Log taken from when Ed tries to login using his google account:

```
  User Load (1.1ms)  SELECT "users".* FROM "users" WHERE "users"."username" = 'EdWoodward'
CreateUser#exec user: {"created_at":null,"first_name":"Ed","full_name":"Ed Woodward","id":null,"is_administrator":false,"is_temp":true,"last_name":"Woodward","person_id":null,"title":null,"updated_at":null,"username":"EdWoodward","uuid":null}
  User Exists (1.1ms)  SELECT 1 AS one FROM "users" WHERE "users"."username" = 'EdWoodward' LIMIT 1
  User Exists (1.7ms)  SELECT 1 AS one FROM "users" WHERE LOWER("users"."username") = LOWER('EdWoodward') LIMIT 1
CreateUser#exec user.errors: ["Username has already been taken"]
```

From #123

karenc:
Fix #143

### DIFF
--- a/app/routines/create_user.rb
+++ b/app/routines/create_user.rb
@@ -21,7 +21,7 @@ class CreateUser
 
       # If the number of attempts is exceeded, the user creation will fail
       for i in 1..USERNAME_ATTEMPTS do
-        break if User.where(username: username).none?
+        break if User.where('LOWER(username) = ?', username).none?
         username = get_valid_username(options[:username])
         username = randomify_username(username)
       end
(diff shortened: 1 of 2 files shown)


app/routines/create_user.rb
```diff
@@ -21,7 +21,7 @@ def exec(options = {})
 
       # If the number of attempts is exceeded, the user creation will fail
       for i in 1..USERNAME_ATTEMPTS do
-        break if User.where(username: username).none?
+        break if User.where('LOWER(username) = ?', username.downcase).none?
```
jpslav: Can username.downcase be in `get_valid_username`?

karenc: Yeah, good idea.

Issue #119, opened 2015-01-30
Allow one user to have multiple accounts from the same provider


Dantemss:
E.g. a user with multiple google accounts.

Expect this to error out currently.

nathanstitt:
The main roadblock to this seems to be that the `authentications` table has a unique composite index on the "user_id" and "provider" columns.  Since User's authentications association is already a has_many, removing the unique constraint **may** be all that is needed.

PR #145, opened 2015-03-06
Support multiple accounts per authentication provider

nathanstitt:
This allows a person to login using multiple accounts from
the same provider, as when a person has multiple gmail
accounts and wants to associate them all with a single
openstax one.

In response to issue #119

### DIFF
--- a/app/assets/stylesheets/sessions.css.scss
+++ b/app/assets/stylesheets/sessions.css.scss
@@ -13,12 +13,27 @@ $login-options-width: 300px;
   float: right;
   text-align: center;
 
-  .no-more-login-options {
-    font-style: italic;
-  }
-
   .login-button {
-    padding: 10px;
+      padding: 10px;
+      position: relative;
+      &.adding {
+          img {
+
+          }
+          &:before{
+              content: "+";
+              display: block;
+              border: 1px solid grey;
+              width: 25px;
+              height: 25px;
+              position: absolute;
+              bottom: 23px;
+              right: 10px;
+              background: darkblue;
+              border-bottom-left-radius: 5px;
+              opacity: 0.6;
+          }
+      }
   }
 
   .or-bar {
@@ -177,10 +192,6 @@ $login-options-width: 300px;
     }
 
     .login-options {
-      .no-more-login-options {
-        margin-top: 150px;
-      }
-      
       .or-text {
         background-color: whitesmoke;
       }
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -1,5 +1,17 @@
 module SessionsHelper
 
+
+  def oauth_login_link(service, options)
+    title = options[:title] || service.capitalize
+    has_login  = current_user.authentications.any?{|auth| auth.provider == service }
+    css_class  = has_login ? "login-button adding" : 'login-button'
+    link_title = has_login ? "Sign in with a different #{title} account" : "Sign in using your #{title} account"
+
+    link_to( "/auth/#{service}", { class: css_class, id: "#{service}-login-button" } ) do
+      image_tag( options[:icon], title: link_title )
+    end
+  end
+
   def authentications_list(authentications)
     providers = authentications.collect{|auth| auth.provider}
     provider_list(providers, 'and')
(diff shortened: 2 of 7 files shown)


jpslav:
Also, we need to address the UI to make this change apparent to users.  As it is now, once a user registers with a Google account, the button to login with Google disappears (as before we couldn't handle more than one Google authentication).  It disappearing was kind of nice because it helped drive our registering user to eventually say "finish setting up my account".  I kind of worry if we show a "sign in with Google button" after they just signed in with Google that they may get confused.  We can chat about this if not clear.

Dantemss:
OK so the validation needs to be gone from the model...
And we need some tests to make sure that the behavior is correct (i.e. that we can add 2 authentications to a user and loging in as either one will allow you to login).

Dantemss:
And from what JP said we need to not hide the "sign in with google" button after you sign in with google etc. And might need some UX design there.

nathanstitt:
@Dantemss & @jpslav  I'm confused on where you see the uniqueness on the model.  Are you referring to [Authentication's uid](https://github.com/openstax/accounts/blob/master/app/models/authentication.rb#L6) validation?

If so that only checks that the UID that's returned from google/twitter/fb etc is unique.  In the case of a person having two gmail accounts, they'll have two different UIDs, but the same provider ID.

I've added some specs to `spec/models/authentication_spec.rb` that (hopefully) illustrates it.

I'm afraid I'll need a hint on removing the "sign in with google button"  I found https://github.com/openstax/accounts/blob/master/app/views/identities/new.html.erb but don't see anywhere that buttons are hidden.  Am I looking at the wrong place or is there some JS/CSS combo that hides it?

Dantemss:
Nathan is right JP, that validation is unrelated to this issue.
Not sure about the button hiding, but I did see it being hidden in the Accounts demo during the meeting.

jpslav:
At first I agreed with y'all that I was wrong, but I was mistaken ;-)

@nathanstitt correctly points out that the authentication UID uniqueness validation in the model is ok.  I agree with that.  But that means we should leave the uniqueness constraint in the migration (what's good for the model is even better for the DB).  

So I'm pretty sure that means these aren't the changes we were looking for.  Maybe since there aren't existing validations that require a provider to be unique for each User, that means there aren't business logic changes to be made.  

But there are at least view changes to be made (the disappearing Google login, etc).  I'll try to comment on places where we need changes.  

Does this all make sense?

nathanstitt:
It makes 1/2 sense ;-)  I think the confusion is stemming from the problem that the constraint in the DB didn't match the AR validation.

The DB constraint that I removed was enforcing uniqueness between the `user_id` and `provider` fields, while the AR validation is enforcing uniqueness between the `uid` and `provider` fields. 

The `uid` field is populated by the oauth provider and represents a unique id for the user in their system.  The `user_id` is the ID of our user object.

If order to allow a user to have two different Google logins, they'll have the same provider listed multiple times, but each should have a unique `uid`.

Your point about the DB matching the business logic validations is good.  We should add a unique constraint in the DB to match the AR validation (on `uid` <-> `provider`).

I just found where the buttons are hidden in the view layer: [views/sessions/_login.html.erb](https://github.com/openstax/accounts/blob/master/app/views/sessions/_login.html.erb#L1-L18).  I'll get that corrected in a bit

nathanstitt:
I've just committed the first pass at reworking the UI to support multiple logins with same provider.  I'm not at all happy with what I've came up with, but am hopeful that this is enough for us to understand the problem and brainstorm solutions around.

This is what I've came up with:
screen shot 2015-03-12 at 4 27 00 pm

I'm overlaying the plus onto the icon and changing the mouse-over (title) to reflect that the service has been used before and clicking it will add an additional account. [css](https://github.com/openstax/accounts/blob/multiple-auth-from-provider/app/assets/stylesheets/sessions.css.scss#L23-L35)  I'm unhappy with this because I really doubt that anyone's going to know what the "+" sign indicates and will probably not read the tooltip.

I think a better implementation would be to remove the icon if they've already used it and replace with either a different icon or just text reading **Add additional Twitter Login** 

However there's also space at the left that we could use to provide a grid layout that would allow us to space to write a longer description of what's going on.   People usually don't read longer descriptions though.

Any other thoughts?

edwoodward:
Closing this due to age. It can be revisited if necessary.

PR #147, opened 2015-03-10
Fix oauth application management rake tasks.

lakshmivyas:
- Add more comments to explain the purpose of the file and tasks
- Refactor to make sure applications can be created from blank slate
- Fix application names to match the branding guidelines

\cc @jpslav

### DIFF
--- a/lib/tasks/add_ost_apps.rake
+++ b/lib/tasks/add_ost_apps.rake
@@ -1,13 +1,23 @@
+# The rake tasks in this module are used to manage trusted OpenStax
+# oauth applications.  Currently, Exercises, Tutor, Exchange and
+# Biglearn oauth applications are managed by these tasks.
 require 'doorkeeper/models/active_record/application'
 
 namespace :accounts do
   namespace :ost do
     # These are the apps that we want to create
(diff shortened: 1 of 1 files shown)


lib/tasks/add_ost_apps.rake
```diff
@@ -17,23 +27,30 @@ namespace :accounts do
...
           auth = Authentication.find_or_create_by_uid(user.identity.id.to_s) do |new_auth|
             new_auth.provider = 'identity'
             new_auth.user_id = user.id
             new_auth.save!
           end
           # For each app above, create and assign defaults if it doesn't exist
+          cb_path = "accounts/auth/openstax/callback"
           app_data.each do |app|
-            url = "https://#{app[:prefix]}#{suffix}.openstax.org/accounts/auth/openstax/callback"
+            # If the suffix parameter starts with `http` assume that
```
Dantemss: Why is this necessary?

lakshmivyas: On vagrant installations, we are serving from a non-standard port (port forwarding on host).  In those cases, we need to provide the full-url with the non-standard port.

lakshmivyas:
Closing and reopening the PR to trigger travis. (failed because of rubygems server issues).

PR #150, opened 2015-03-13
testing .travis.yml

cnuber:
testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
+sudo: false
 language: ruby
+cache: bundler
+addons:
+  postgresql: "9.3"
 rvm:
   - "1.9.3-p547"
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,8 +10,9 @@ before_install:
   - tar xf phantomjs-1.9.7-linux-x86_64.tar.bz2
   - export PATH=$PATH:phantomjs-1.9.7-linux-x86_64/bin/
 install:
-  - bundle install --without production --retry=6
+  - bundle install
```
jpslav: Does this explicit call to `bundle install` conflict with the `bundler_args` below?

jpslav: @cnuber any thoughts on this?

cnuber: this could be represented either way.  the bundler args are just passed to bundler.  it's not a separate run.

PR #151, opened 2015-03-13
Error pages

jpslav:
@karenc; the `current_human_user` error you saw was due to the fact that JSON requests tend to fall into our api machinery (which is where we want them), but the route took it to the admin controller which was missing some pieces of that machinery.  So I changed the JSON tests to point to `/api` routes.  

Also on the 404 request (`/asdf` giving a 500), I think this must be some capybara thing.  Maybe it assumes that  the route passed to `visit` is a good one.  When I stepped into capybara in the debugger I saw a `status: :success` that somehow magically converted to a 500.  So I sidestepped the issue and just had the 404 specs hit the endpoint that raises a routing error.

And somehow when I tried to push my changes there was a merge conflict (don't understand how that's possible since no one had changed the branch after me).  Anyway, there's a lovely merge commit at the end that is my special gift to you ;-)

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -4,6 +4,8 @@
 module Admin
   class BaseController < ApplicationController
     
+    include FakeExceptionHelper
+
     if Rails.env.production?
       before_filter :authenticate_admin!
     else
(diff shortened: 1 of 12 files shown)


karenc:
Just need a few more minutes to fix the commits.

karenc:
Added another commit which fixes the 404 page.  I tested it locally, it was using public/404.html.  It seems in Rails 3 it's not possible to catch the routing error easily so added a route and some code in the application controller to get this to work.

karenc:
This PR is ready for review.

PR #154, opened 2015-03-17
Create users with identity and authentication in dev console

karenc:
Dev console created only users, you can't login with just user.  Change
it to also create identity and authentication.  The password for all the
users created in the dev console is "password".  Also the users created
in the dev console are set to be not temp (not new users) and all the
contracts are automatically accepted.

@pumazi is this what you wanted?


pumazi:
@karenc Yep. That sounds very useful and definitely what I wanted.

pumazi:
Looks like this works! Thank you. :+1:

PR #157, opened 2015-03-19
action_interceptor update

Dantemss:
All done.


Dantemss:
This PR is not needed to fix the cookie overflow bug, but it contains the bugfix that was applied to Exercises, so it's recommended.

PR #166, opened 2015-03-24
Upgrade doorkeeper

karenc:
Update gems: openstax_api, openstax_utilities, roar-rails

Update the gems using "bundle update --source &lt;gem-name&gt;".

Add Api::V1::ApiController which has some code in the old version of
Openstax::Api::V1::ApiController that the other api controllers depend
on.

Update code and tests to work with the new version of
OpenStax::Api::Roar standard_read, standard_create, standard_update,
standard_destroy.

================
Upgrade doorkeeper to 1.4.2 to support multiple callback urls

Add migration for doorkeeper tables by comparing the doorkeeper
migration script we had and the migration script generated by the new
version.

Add all the additional comments in config/initializers/doorkeeper.rb
from the new version.

Update config/locales/doorkeeper.en.yml to the new version.

Change most of the doorkeeper templates to the new version while keeping
all of our customizations.


pumazi:
Works for me. I tested it locally with authoring and publishing using the same app keys. :+1:

Closes #134

Dantemss:
Nice! I'm upgrading openstax_api to support doorkeeper 2+, but that can come later.

Dantemss:
Looks good to me, but I think JP should take a look too.

app/controllers/api/v1/api_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V1::ApiController < OpenStax::Api::V1::ApiController
+  skip_before_filter :authenticate_user!
```
jpslav: how was this working before when not skipped? or was it not working?

config/initializers/doorkeeper.rb
```diff
@@ -49,15 +54,27 @@
...
 
-  # Change the test redirect uri for client apps
+  # Change the native redirect uri for client apps
   # When clients register with the following redirect uri, they won't be redirected to any server and the authorization code will be displayed within the provider
   # The value can be any string. Use nil to disable this feature. When disabled, clients must provide a valid URL
   # (Similar behaviour: https://developers.google.com/accounts/docs/OAuth2InstalledApp#choosingredirecturi)
   #
-  # test_redirect_uri 'urn:ietf:wg:oauth:2.0:oob'
+  # native_redirect_uri 'urn:ietf:wg:oauth:2.0:oob'
+
```
jpslav: Where did blocks like the following go?  (want to make sure we are still authenticating and restricting access appropriately):

```ruby
  # This block will be called to check whether the resource owner is authenticated or not.
  resource_owner_authenticator do
    authenticate_user!
    current_user
  end

  # If you want to restrict access to the web interface for adding oauth authorized applications, you need to declare the block below.

  # We allow all users of Accounts to manage applications
  # We subclassed Doorkeeper::ApplicationsController to provide better
  # control over access to the Doorkeeper::Application pages
  admin_authenticator do
    authenticate_user!
    current_user
  end
```

jpslav:
Other than my 2 comments above (about auth), this looks good to me.

app/controllers/api/v1/api_controller.rb
karenc: It used to be in openstax_api, it got removed here: https://github.com/openstax/openstax_api/commit/643f4b5f2f0fa013f84f6c0808008b330393af50

Similarly for redirect_to.

config/initializers/doorkeeper.rb
karenc: If you expand the diff for this file, these blocks are still in this file.
https://github.com/openstax/accounts/pull/166/files#diff-e621685ac1c2689eaedaedf334c6123bL7

jpslav: Ah thought I did them all but missed the one at the top!

PR #171, opened 2015-03-27
Added FinePrint on home page

Dantemss:
So admins get asked to sign terms before they reach the admin console and it errors out.

### DIFF
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -3,7 +3,7 @@ class StaticPagesController < ApplicationController
                      only: [:api, :copyright, :home, :status]
 
   fine_print_skip :general_terms_of_use, :privacy_policy,
-                  only: [:api, :copyright, :home, :status]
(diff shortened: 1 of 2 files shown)


app/controllers/static_pages_controller.rb
```diff
@@ -3,7 +3,7 @@ class StaticPagesController < ApplicationController
                      only: [:api, :copyright, :home, :status]
 
   fine_print_skip :general_terms_of_use, :privacy_policy,
-                  only: [:api, :copyright, :home, :status]
```
jpslav: What happens if a user isn't logged in?  Does this change indirectly mean they'll be asked to login? (so they can eventually sign the agreements)

Dantemss: Sorry I never saw this... fine print should not ask for a contract if you are not logged in, I think.

Dantemss:
OBE. See https://github.com/openstax/accounts/pull/250.

PR #176, opened 2015-05-04
Change db config to use postgresql for development and tests

karenc:


.travis.yml
```diff
@@ -9,6 +9,8 @@ before_install:
   - wget 'https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.7-linux-x86_64.tar.bz2'
   - tar xf phantomjs-1.9.7-linux-x86_64.tar.bz2
   - export PATH=$PATH:phantomjs-1.9.7-linux-x86_64/bin/
+  - psql -c "CREATE USER accounts WITH SUPERUSER PASSWORD 'accounts'" -U postgres
```
jpslav: How does tutor-server not need to set this? https://github.com/openstax/tutor-server/blob/master/.travis.yml

Anything we can do to make the approach in Account mirror the existing approach in tutor-server is a good thing.

karenc: tutor-server uses postgres as the user.
https://github.com/openstax/tutor-server/blob/master/.travis.yml#L31

karenc: I guess I can add environment variables and use it in config/database.yml.

.travis.yml
```diff
@@ -9,6 +9,9 @@ before_install:
   - wget 'https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.7-linux-x86_64.tar.bz2'
   - tar xf phantomjs-1.9.7-linux-x86_64.tar.bz2
   - export PATH=$PATH:phantomjs-1.9.7-linux-x86_64/bin/
+  - psql -c 'CREATE DATABASE "accounts-testing"' -U postgres
```
karenc: I'm using the default test database name here instead of `travis_ci_test` like [tutor-server](https://github.com/openstax/tutor-server/blob/master/.travis.yml#L30).  Not completely sure why it was set to `travis_ci_test` in tutor-server.

.travis.yml
lakshmivyas: Its probably a good idea to create a separate user that has privileges similar to production environment when running the tests.

https://github.com/openstax/tutor-deployment/blob/master/tasks/init_app_db.yml

karenc: @lakshmivyas is that `psql -c "CREATE USER accounts WITH SUPERUSER PASSWORD 'accounts'" -U postgres` or do I need to do something else?

lakshmivyas: @karenc - The user should be regular user (not super) and there should be an additional line that grants the user, privileges for everything in the test database.

karenc: @lakshmivyas I changed it to:

```bash
psql -c "CREATE USER accounts WITH CREATEDB PASSWORD 'accounts'" -U postgres
psql -c 'CREATE DATABASE "accounts-testing" WITH OWNER accounts' -U postgres
psql -c 'GRANT ALL PRIVILEGES ON DATABASE "accounts-testing" TO accounts' -U postgres
```

I have to give the user permission to create db otherwise I get something like `PG::InsufficientPrivilege: ERROR:  permission denied to create database` when I run the tests.

Does this look ok?

karenc:
Ready for review.

.travis.yml
lakshmivyas: @karenc - Looks good. 

It'd be nice if we can remove the create db requirement at a later point.   Instead the tests can do drop all on the tables and recreate them.

jpslav:
@nathanstitt can you review since you did this before for Tutor?

will target merge for just after this sprint ends.

spec/routines/search_users_spec.rb
```diff
@@ -124,4 +124,4 @@
 
   end
 
-end
\ No newline at end of file
+end
```
nathanstitt: You probably already figured out what was wrong with the .first, but it's because Postgresql sorts differently.

You could fix it by specifying a sort in the query, i.e.  `user_1.contact_infos.email_addresses.order(:value).first`  

Marking them all verified is probably the right thing do though, so that's moot.

nathanstitt:
Looks good other than you might consider the travis technique I outlined in the exchange PR.

spec/routines/search_users_spec.rb
karenc: I think your solution is better.  I've changed it to `email_addresses.order(:value).first` :)

PR #178, opened 2015-05-28
Accept terms and unexpire passwords

karenc:
- Add accounts:accept_all_terms rake task
  
  Accept all terms and licenses for all users

- Add accounts:unexpire_all_passwords rake task

  Unexpire all passwords for all identities (local users).  This means
that users won't be forced to change their passwords.  (For demo)

### DIFF
--- a/lib/accept_all_terms.rb
+++ b/lib/accept_all_terms.rb
@@ -1,0 +1,12 @@
+class AcceptAllTerms
+  BATCH_SIZE = 1000
+
+  def run
+    contracts = FinePrint::Contract.all
+    User.find_each(batch_size: BATCH_SIZE) do |user|
+      contracts.each do |contract|
+        FinePrint.sign_contract(user, contract)
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/accept_terms.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(users)
+    @users = users
+  end
+
+  def run
+    i = 0
+    contracts = FinePrint::Contract.all
+    while true do
+      users = @users.limit(OBJECTS_PER_TRANSACTION)
```
jpslav: Instead of manually doing batch queries, just use `find_each` -- does the same thing.

http://api.rubyonrails.org/classes/ActiveRecord/Batches.html

karenc: Ok I'll do that

jpslav:
My bad @karenc.  This looks great but I asked you for the wrong thing.  Instead of setting users' passwords to 'password', really I should have asked you to remove the password expiration date for all passwords.  Tutor already sets the passwords to 'password' but the expiration date is set to Time.now to force users to change their password on login.  We just want to remove this requirement to reset their password, and we can do that by setting `identity.password_expires_at = nil`.

While you're in there, please rename the `AcceptTerms` class to `AcceptAllTerms` (just to be overly clear).  And of course `SetPassword` should change but then also include "All", so maybe `UnexpireAllPasswords`.

PR #180, opened 2015-07-22
Add seeding tasks

pumazi:
This adds two tasks. The first is a task for creating an admin user. The second is a task for invoking custom seed files. For example, this pull request contains a custom seed called 'hanna_barbera'. This can be invoked using ``rake db:seed:hanna_barbera``.

### DIFF
--- a/db/seeds/hanna_barbera.rb
+++ b/db/seeds/hanna_barbera.rb
@@ -1,0 +1,44 @@
+# Creates a number of users themed from Hanna-Barbera cartoon characters.
+
+PASSWORD = 'password'
+filepath = File.join(Rails.root, 'db', 'seeds', 'hanna_barbera.yml')
+
+
+yaml_root = YAML.load_file(filepath)
+master_group = Group.find_or_create_by_name('hanna_barbera_group')
+
+yaml_root['users'].each do |username, values|
+  user_info = values.slice('first_name', 'last_name', 'full_name',
+                           'title', 'suffix')
+  # Create users...
+  user = User.find_or_create_by_username(username)
+  user.update_attributes(user_info)
+  identity = Identity.find_or_create_by_user_id(user.id) do |identity|
+    identity.password = PASSWORD
+    identity.password_confirmation = PASSWORD
+    identity.save!
+  end
+  user.is_administrator = values['is_administrator']
+  user.state = 'activated'
+  master_group.add_member(user)
+  identity_uid = user.identity.id.to_s
+  auth = Authentication.find_or_create_by_uid(identity_uid) do |auth|
+    auth.provider = 'identity'
+    auth.user_id = user.id
+    auth.save!
+  end
+  user.save
+end
+
+yaml_root['groups'].each do |groupname, values|
+  # Create groups
+  group = Group.find_or_create_by_name(groupname)
+  values['owners'].each do |username|
+    user = User.find_by_username(username)
+    group.add_owner(user)
+  end
+  values['members'].each do |username|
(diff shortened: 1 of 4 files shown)


db/seeds/hanna_barbera.rb
```diff
@@ -0,0 +1,44 @@
...
+
+
+yaml_root = YAML.load_file(filepath)
+master_group = Group.find_or_create_by_name('hanna_barbera_group')
+
+yaml_root['users'].each do |username, values|
+  user_info = values.slice('first_name', 'last_name', 'full_name',
+                           'title', 'suffix')
+  # Create users...
+  user = User.find_or_create_by_username(username)
```
karenc: Do we really need to check if the user exists?  This is test/development data, maybe it's better to just fail?
Also, shall we use `FactoryGirl.create` instead of `User.create`?

lib/tasks/add_ost_apps.rake
```diff
@@ -11,6 +11,36 @@ namespace :accounts do
...
+    # Task specifically to create an administrative user and/or set the
+    # password for said user. This task accepts to arguments, the username
+    # and optionally a password. If the password is not supplied, the
+    # username will be used for the password.
+    task :create_admin, [:username, :password] => :environment do |t, args|
+      ActiveRecord::Base.transaction do
+        begin
+          username = args[:username]
+          password = args[:password] || args[:username]
+          admin_group = Group.find_or_create_by_name('ost_app_admin_group')
```
karenc: Is it necessary to add the user to a group?  Is `user.is_administrator = true` not enough?

karenc:
I don't know much about groups... What are they used for?

karenc:
@jpslav @Dantemss probably know more about this :)

lib/tasks/add_ost_apps.rake
pumazi: I'm adding them to the group, because that's what the other task within this file does.

db/seeds/hanna_barbera.rb
pumazi: ``find_or_create`` is used incase you want to change the data. This way you don't need to wipe the db to update the data. Or is this something ``FactoryGirl.create`` also does?

lib/tasks/add_ost_apps.rake
karenc: This task is about adding apps, don't know if creating admins belongs here.  (I don't know the details)

db/seeds/hanna_barbera.rb
karenc: I suggested `FactoryGirl.create` because in case the model changes and requires more fields, this script won't fail.
I guess we won't accidentally run this script on non-development environments given the script name, so I guess it's ok?
Usually I just wipe the db and populate it again in development.

jpslav:
I agree with @karenc.  The seeds should be used for initial data needed by deployments, not for test data.  For test/development data I recommend making specific rake tasks in /lib/tasks, e.g. [the rake task we use in Tutor to set up demo data](https://github.com/openstax/tutor-server/blob/master/lib/tasks/demo.rake).

Also, would be good to not double-book the `add_ost_apps.rake` rake file with a task to create an admin.   If an admin is needed for development purposes, you can make one in the development console (`/dev`).  If an admin user is needed for production-like deployments, there could be a separate rake file for creating that admin user.  No need to add to groups, just need to do similar things to what goes on in https://github.com/openstax/accounts/blob/master/app/handlers/dev/users_create.rb.

PR #181, opened 2015-07-23
Added skipping of terms

### LINKED ISSUE #179: Manually type in /logout --> 500 error
An exception occurred: bad argument (expected URI object or URI string)
.../lib/ruby/1.9.1/uri/common.rb:996:in `URI'
...app/controllers/sessions_controller.rb:40:in `destroy'

jpslav:
* Lets terms of use be skipped for users from certain applications (to avoid double signing when that application's terms subsume Accounts terms)
* Set Tutor to be one of these skipped terms apps (in rake setup file)
* Cleaned up session state on sign out (getting rid of `return_to`, `client_id`, etc).
* Fixed #179

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -375,3 +375,6 @@ DEPENDENCIES
   thin
   uglifier (>= 1.0.3)
   unicorn
+
+BUNDLED WITH
+   1.10.1
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -14,6 +14,7 @@ class SessionsController < ApplicationController
     store_fallback
     referer = request.referer
     session[:from_cnx] = (referer =~ /cnx\.org/) unless referer.blank?
+    session[:client_id] = params[:client_id]
     @application = Doorkeeper::Application.where(uid: params[:client_id]).first
   end
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -22,7 +22,7 @@ class SessionsController < ApplicationController
     handle_with(SessionsCallback, user_state: self,
       complete: lambda {
         case @handler_result.outputs[:status]
-        when :returning_user    then redirect_back
+        when :returning_user    then redirect_to action: :returning_user
         when :new_user          then render :ask_new_or_returning
         when :multiple_accounts then render :ask_which_account
         else                    raise IllegalState
@@ -31,6 +31,10 @@ class SessionsController < ApplicationController
   end
 
   def destroy
+    session[ActionInterceptor.config.default_key] = nil
+    session[:registration_return_to] = nil
+    session[:client_id] = nil
+
     sign_out!
 
     # Hack to find root of referer
@@ -38,8 +42,12 @@ class SessionsController < ApplicationController
     # that are not at the root of their host after logout
     # TODO: Replace with signed or registered return urls
     #       Need to provide web views to sign or register those urls
-    uri = URI(request.referer)
-    url = "#{uri.scheme}://#{uri.host}:#{uri.port}/"
+    url = begin
+      uri = URI(request.referer)
+      "#{uri.scheme}://#{uri.host}:#{uri.port}/"
+    rescue # in case the referer is bad (see #179)
+      root_url
+    end
 
     redirect_to url, notice: "Signed out!"
   end
@@ -48,6 +56,12 @@ class SessionsController < ApplicationController
   end
 
   def i_am_returning
+  end
+
+  # This is an official action instead of just doing `redirect_back` in callback
+  # handler so that fine_print can check to see if terms need to be signed.
+  def returning_user
+    redirect_back
   end
 
   # Omniauth failure endpoint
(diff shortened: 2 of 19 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -14,6 +14,7 @@ def new
     store_fallback
     referer = request.referer
     session[:from_cnx] = (referer =~ /cnx\.org/) unless referer.blank?
+    session[:client_id] = params[:client_id]
```
jpslav: Storing `client_id` in session so that all phases of sign up / in can know that this user came from this application (so they can know not to require terms if applicable).

app/controllers/sessions_controller.rb
```diff
@@ -30,6 +31,10 @@ def callback
   end
 
   def destroy
+    session[ActionInterceptor.config.default_key] = nil
+    session[:registration_return_to] = nil
+    session[:client_id] = nil
```
jpslav: @Dantemss can you bless this clearing of session vars?  (there were cases where old session data was messing up some links during registration)

app/controllers/sessions_controller.rb
```diff
@@ -37,8 +42,12 @@ def destroy
...
     # TODO: Replace with signed or registered return urls
     #       Need to provide web views to sign or register those urls
-    uri = URI(request.referer)
-    url = "#{uri.scheme}://#{uri.host}:#{uri.port}/"
+    url = begin
+      uri = URI(request.referer)
+      "#{uri.scheme}://#{uri.host}:#{uri.port}/"
+    rescue # in case the referer is bad (see #179)
+      root_url
+    end
```
jpslav: This is the fix for #179

spec/controllers/api/v1/application_groups_controller_spec.rb
```diff
@@ -179,8 +179,7 @@
     end
 
     it "should not let a user call it through an app" do
-      api_get :updates, user_2_token
-      expect(response.status).to eq(403)
+      expect{api_get :updates, user_2_token}.to raise_error(SecurityTransgression)
```
jpslav: I have no idea why these old specs started failing all of the sudden.

jpslav:
@karenc @Dantemss can you review?

karenc:
Looks good to me.

karenc:
I tried to single out the change that broke the tests, it's in [config/initializers/controller.rb](https://github.com/openstax/accounts/pull/181/files#diff-602a8d9fdd33f6afda6d4a95a3e3a073).  Something like `return false` would work, but even `@contracts_not_required ||= (request.format == :json)` breaks the test...  Any ideas?

Gemfile.lock
```diff
@@ -375,3 +375,6 @@ DEPENDENCIES
   thin
   uglifier (>= 1.0.3)
   unicorn
+
+BUNDLED WITH
```
Dantemss: ???

jpslav:
Karen figured out that my changes revealed an existing bug of sorts re the 403 spec failures (what she is referring to above), and so the spec changes are needed (yay)

Gemfile.lock
jpslav: new version of bundler does this maybe?

app/controllers/sessions_controller.rb
Dantemss: You can't guarantee that the session vars will be cleared, because you can't guarantee that the user will ever come here.

karenc:
Turns out @jpslav actually fixed the api tests! :+1: and the way the tests are working now is as they should've.

What was happening is that we hit `fine_print_require`, then `fine_print_redirect`, but in our [API controller](https://github.com/openstax/accounts/blob/master/app/controllers/api/v1/api_controller.rb#L6) returns `head :forbidden` for redirects.  So the `403` we saw in the tests are actually from `def redirect_to`, not from the access policy code.

app/controllers/sessions_controller.rb
```diff
@@ -49,6 +58,12 @@ def ask_new_or_returning
   def i_am_returning
   end
 
+  # This is an official action instead of just doing `redirect_back` in callback
+  # handler so that fine_print can check to see if terms need to be signed.
```
Dantemss: I think this is unnecessary? FinePrint should check that on the doorkeeper authorize route.

app/controllers/sessions_controller.rb
Dantemss: Could use the client_id in the session to replace this hack.

Dantemss: But of course that means you need to calculate this before you clear the client_id

Gemfile.lock
jpslav: Hopefully this will [only ever result in a warning message](https://meta.discourse.org/t/bundler-always-adds-bundled-with-to-gemfile-lock/29435/5), and looks like something here to stay.  Seem ok to leave?

app/controllers/sessions_controller.rb
Dantemss: This probably needs some manual testing...

app/controllers/sessions_controller.rb
jpslav: This is needed when someone is just signing in directly to Accounts (without getting there from an app like Tutor) -- i.e. when there's no doorkeeper authorize route in the mix.  Discovered this when writing the last scenario in the feature spec below.

PR #186, opened 2015-07-30
Filter passwords in JSON parameters from logs

karenc:


config/initializers/filter_parameters.rb
```diff
@@ -0,0 +1,46 @@
...
+string_regexp = Regexp.new(string_filters.collect { |p| p.to_s }.join('|'))
+
+# Get all the non string filter parameters
+non_string_filters = filter_parameters - string_filters
+
+# Implement a filter proc that filters out JSON parameters and include existing
+# string filters
+json_key_regexp = /password/
+
+Rails.application.config.filter_parameters = [lambda do |k, v|
```
jpslav: can you put this lambda in its own lib file and write a spec for it?

config/initializers/filter_parameters.rb
```diff
@@ -0,0 +1,46 @@
...
+#
+# The way to get around this is to remove the string filters and include them
+# in the filter proc below.
+
+# First get all the filter parameters
+filter_parameters = Rails.application.config.filter_parameters
+
+# Get all the string filter parameters (these are used to match against the key
+# of the parameters and filter the value
+string_filters = filter_parameters.select { |p| p.is_a?(Symbol) || p.is_a?(String) }
```
jpslav: Is it possible that since `string_filters` is removed from the overall filters, is it possible we're removing other non-password filters here?

config/initializers/filter_parameters.rb
```diff
@@ -0,0 +1,46 @@
...
+string_regexp = Regexp.new(string_filters.collect { |p| p.to_s }.join('|'))
```
jpslav: or are you basically replacing all filters with the one below, which uses `string_regexp`?

karenc: Yes, I'm adding all the string filters into the one filter, using pretty much the same code that rails is using for string filters.

config/initializers/filter_parameters.rb
karenc: Ok, I'll move this code and add a spec.

PR #187, opened 2015-07-31
Add names to find or create user API

karenc:
Set first_name, last_name and full_name in find or create user API


karenc:
@jpslav is this ok to merge?

PR #188, opened 2015-07-31
Assorted cleanup

### LINKED ISSUE #98: Handle user landing on Accounts main page
Although we try to always redirect the user back when they are done using Accounts, it is possible that they could end up on the Accounts main page either through Google or if we somehow lose track of where they came from.

Right now the Accounts root page is just a blank page.

Should we display a list of OpenStax sites on the Accounts main page and let the user select where they want to go if that happens?

jpslav:
* Avoids blank home page -- signed in users go to the profile page; signed out users go to the sign in page
* Correct out-of-date class names: specifically those around the users/find-or-create capability still had old names that only made sense w.r.t a prior implementation.
* Corrected API docs
* Added tabs to the profile page
* Link to profile page from username in header

Fixes #98

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -59,11 +59,11 @@ class Api::V1::UsersController < Api::V1::ApiController
 
     Examples:
 
-    `username:ric` &ndash; returns ApplicationUsers for 'richb' and 'ricardo' Users.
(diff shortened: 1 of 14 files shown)


jpslav:
@karenc please review

karenc:
Looks good to me.  Should I merge?

PR #190, opened 2015-08-10
Add "back" link on profile form

karenc:
Was removed when tabbifying the profile form.

The condition `if controller.request.env["HTTP_REFERER"].present?` is
added because we don't want rails to use "history.back()" in case the
referer is empty.

accounts-profile-back-link

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -31,3 +31,7 @@
                email_addresses: current_user.email_addresses } %>
   </div>
 </div>
+
+<% if controller.request.env["HTTP_REFERER"].present? %>
+  <%= link_to 'Back', :back %>
+<% end %>


karenc:
May be it should be a button?

jpslav:
Should we just add a `return_to` URL parameter to this profile path?  Then when present (when set by Tutor or CNX), we can show a big button / header link that @fredasaurus and @jason-holmes are happy with?

karenc:
I added a check `<% if controller.request.env["HTTP_REFERER"].present? %>` and the back link goes back to the dashboard (since I clicked on "My Account" from the dashboard) even though the profile page shows up in a new window.  The user can still click on the back link and it would do something that makes sense.  The behavior seems ok to me.

app/views/users/edit.html.erb
```diff
@@ -31,3 +31,7 @@
                email_addresses: current_user.email_addresses } %>
   </div>
 </div>
+
+<% if controller.request.env["HTTP_REFERER"].present? %>
+  <%= link_to 'Back', :back %>
```
jpslav: Can it be an explicit link to that referrer?

@Dantemss you are the king of all things having to do with passing return URLs around -- what do you think?

karenc: actually we probably need to store this in the session?  from what i remember, the back link still worked after the profile form is saved...

Dantemss: Since the profile is a single page, probably not necessary...

jpslav:
@karenc I'm going to close this PR, as in many cases the profile page is either in an iframe or a new tab, or changes like my latest PR offer other ways to get out of the profile page.  If you have other thoughts, please feel free to reopen.

PR #201, opened 2015-10-06
Integrate RescueFrom

joemsak:


config/initializers/rescue_from.rb
```diff
@@ -0,0 +1,22 @@
+require 'openstax_rescue_from'
+
+OpenStax::RescueFrom.configure do |config|
+  config.raise_exceptions = Rails.application.config.consider_all_requests_local
+
+  config.app_name = 'Accounts'
+  config.app_env = ENV['APP_ENV']
```
jpslav: SECRET_SETTINGS[:environment_name]

config/initializers/rescue_from.rb
```diff
@@ -0,0 +1,22 @@
...
+  config.contact_name = DEPLOY_SETTINGS[:exceptions]['contact_name']
```
jpslav: secret settings, but just realized accounts roll-your-own settings probably don't do nesting like secrets.yml in rails 4

joemsak: Well the example file has nesting in it and also I debugged and found that the first level is symbol keys while the second layer is string keys :)

PR #202, opened 2015-10-13
WIP: JS API for remote access to account resources

nathanstitt:
Implements a javascript api that can be remotely loaded and display a modal dialog with a `iframe`


app/assets/javascripts/application.js
```diff
@@ -13,8 +13,12 @@
 //= require jquery
 //= require jquery_ujs
 //= require jquery.ui.all
-//= require_tree .
+//= require ./do
+//= require ./jquery_extensions
+//= require ./ui
+//= require ./ox-account
```
jpslav: Does `ox-account` need to be required in this main JS manifest file?  My feeling (admittedly without a thorough review of everything else) is that since it is only loaded by another site (e.g. Tutor) it doesn't need to be included on normal Accounts page views.

app/controllers/remote_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    headers['Access-Control-Allow-Headers'] = '*'
+    headers['Access-Control-Max-Age'] = '1728000'
+    render :text => '', :content_type => 'text/plain'
+  end
+
+  private
+
+  def set_wideopen_cors
+    # TODO: Replace asterisk with list of domains (cnx.org, etc)
+    headers['Access-Control-Allow-Origin'] = '*'
```
jpslav: When you list domains explicitly, please make configurable and set from tutor-deployment.

app/assets/javascripts/application.js
nathanstitt: A tiny bit of it does.  It needs to:
 * observe that it's loaded inside an iframe and then apply the `iframe` css class to `body`
 * Fire a page load notification to the parent of the iframe so it can possibly close the iframe.
 * installs a window.resize listener so the iframe can avoid scroll bars.

I'm planning to de-couple that so it can only load what's needed.  Right now it's tucked in the `initWebsite` method of `host.coffee`

app/controllers/remote_controller.rb
jpslav: And I like that this is called `set_wideopen_cors` now (to draw attention to the fact that that is what it is doing, but when you limit the domains please change the method name.

jpslav:
Love it @nathanstitt!  I gave it a brief read -- can someone from @openstax/tutor-fe-dev do a full review when the time comes?

app/assets/stylesheets/iframe.css.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  }
+
+  //signup page
+  #signup-alternatives {
+    float: inherit;
+    width: 80%;
+    margin: 0 auto 20px auto;
+  }
+}
```
nathanstitt: Styling could use further tweaks.  For instance, we might want to hide the footer content with the terms and about links.

nathanstitt:
Closing in favor of https://github.com/openstax/accounts/pull/206

PR #203, opened 2015-10-13
Fix global before_filters

jpslav:
1. skip all filters for special 'routing_error' action
2. correct global before_filters to not use non-existent `current_human_user` method

Addresses this exception

```
A NameError occurred in application#routing_error:

 undefined local variable or method `current_human_user' for #<ApplicationController:0x00000004eb6438>
 config/initializers/controllers.rb:41:in `registration'
```

### DIFF
--- a/config/initializers/controllers.rb
+++ b/config/initializers/controllers.rb
@@ -38,31 +38,24 @@ ActionController::Base.class_exec do
   end
 
   def registration
-    user = (request.format == :json) ? current_human_user : current_user
-    return unless user.try(:is_temp?)
-    store_url key: :registration_return_to
+    return true if request.format != :html
 
-    respond_to do |format|
-      format.html { redirect_to register_path }
-      format.json { head(:forbidden) }
-    end
+    return unless current_user.is_temp?
+    store_url key: :registration_return_to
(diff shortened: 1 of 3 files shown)


config/initializers/controllers.rb
```diff
@@ -38,31 +38,24 @@ def contracts_not_required
   end
 
   def registration
-    user = (request.format == :json) ? current_human_user : current_user
```
jpslav: `current_human_user` is not a method available in this class (only available in descendents of the api controller from `openstax_api`)

jpslav:
@karenc please review

Dantemss:
I don't know if I like this... why not just NOT have the routing error route?

Dantemss:
Seems weird that we have a controller action to handle routing error when Rack does it just fine...

jpslav:
I think the thing is that it doesn't.  I will send you and Karen an email separately with links, don't want to cause a comment in a rails/rails issue :-)

karenc:
@Dantemss This is the original commit https://github.com/openstax/accounts/commit/d8030ab069d5366987c036778c32f3c53ed6ea9c.  I can't remember the details...

PR #212, opened 2015-11-22
removed application name from sign in page

jpslav:
Application name didn't work for Tutor vs Concept Coach; decision was to simplify message (remove app name).

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -2,9 +2,7 @@
   #top-nav-account { display: none;}
 </style>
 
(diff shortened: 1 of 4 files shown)


jpslav:
@nathanstitt heads up - I think you probably implemented some JS dynamic changing of "OpenStax Tutor" text on login page for CC.  You can yank that out after this merged.

@Dantemss / @nathanstitt, since this is trivial I'm going to go ahead and merge, but feel free to review.

nathanstitt:
Thanks @jpslav. I've updated CC accordingly.

PR #217, opened 2015-12-01
Allow user to continue with registration when email address is confirmed

karenc:
- store where the site (url) the user came from
- if the came from url is present, add a "continue" button on the email
  confirmation page
- user can continue to the registration form and go back to the site
  they came from


jason-holmes:
Is this the primary page where the user goes from the email, or is a fallback in case the user comes back to this page they might already have opened?  Also please add some white space below the button for the continue message and change the text to, "Already verified your email address?  [link]Complete your profile.[/link]

karenc:
There's a possibility that "Continue" may not go to the profile page, so I didn't want to the link to say "Complete your profile".  Any suggestions?

----

The page that the user goes from the email is:

login12

----

The fallback case: (user will have this page open when they sign up or if the user tries to login without confirming their email address)

login10

jason-holmes:
Got it.  Ok, how about, ""Already verified your email address? [link]Continue."[/link]

karenc:
Does this look ok?

login14

jason-holmes:
Yep.  Looks great, Karen.

-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

jpslav:
Looks great, Karen.  A few comments:

1. Can you do a survey of the pages / flash notices and make sure that "confirm" is always "verify" (make it consistent).  There are a few flash notices in the ContactInfosController.  Also see some in `app/views/confirmation_mailer/instructions.text.erb` and `app/views/contact_infos/confirm.html.erb` but there may be more.
2. Let's check with @Dantemss that passing the return URL around in the confirmation link is ok (from a security perspective).  Probably is, but he has the most experience dealing with return URLs and the like.  Worst case we merge and address it later.  An alternative would be to store it in the DB, which would also probably reduce our need to pass it around a bunch, but there's also a lot to be said for how you have implemented it, so we'll just discuss.
3. In `ContactInfosController#confirm`, can we redirect to the stored_url immediately with a "success" flash if the confirm is successful? (instead of asking them to click a button to continue).  This would mean that `confirm.html.erb` would pretty much just be the error condition page.  In that error message's text, we suggest resending the confirmation -- adding the button to do so here (like you just did elsewhere) would be nice.
4. Dante discovered a wrinkle this afternoon, namely that the social logins (in particular, Twitter) are not guaranteed to return an email address.  This means we can arrive at the email verification page but not have an email on record; this happened to Dante and he got the 404 shown below).  So I think this means we need to add a screen with a form that solicits an email address if there isn't one.  This would either be a pre-before_filter on the existing `registration` before filter, or the `redirect_to(current_user.is_email_pending? ? verification_sent_path : register_path)` line in the `registration` method would have a new case (if no email, send to the get-an-email page, then to the verification sent, and finally to the complete-registration screen).  Give this page your best shot, and I'm sure that we're going to need to run through this whole flow again with @jason-holmes because it feels like there are some simplifications we should be able to make.  Incidentally, this standalone page to solicit an email seems like it is what we could use when we decide we want to get emails from existing users who have never had them or from those who have deleted them.

If you could give these a shot tonight (your day), I can merge in the morning and then we can clean up as needed as CC testing gets underway.

image

karenc:
- Change email confirmation to email verification
- Fix twitter login (or users registered without any email addresses)
- Store the url that the user should return to after registration in the database
- Change "Continue" button on the email verified page to just go to the next page with a message

email_verification

app/controllers/contact_infos_controller.rb
```diff
@@ -41,7 +41,7 @@ def resend_confirmation
                   path = current_user.is_email_pending? ? verification_sent_path :
                                                           profile_path(active_tab: :email)
```
jpslav: After `ContactInfoResendConfirmation` runs, when would `current_user.is_email_pending?` be false?

karenc: Oh... it's probably always true.  I think maybe I'm looking for `current_user.is_temp?`.

app/handlers/contact_infos_confirm.rb
```diff
@@ -15,6 +15,10 @@ def handle
     fatal_error(code: :no_contact_info_for_code) if params[:code].nil?
     contact_info = ContactInfo.where(confirmation_code: params[:code]).first
     fatal_error(code: :no_contact_info_for_code) if contact_info.nil?
+    user = contact_info.user
+    # the came_from url is only used when user is signing up
+    outputs[:stored_url] = user.came_from if user.is_temp?
```
jpslav: I would prefer for this `user.is_temp?` logic to not be here as it isn't the primary responsibility of this routine (which should pretty much just be confirming emails).  Since you've added `contact_info` as an output, seems like the controller can do this work.

And if it is true that `came_from` is only set during sign up, can we rename it to reflect that?  E.g. `registration_redirect_url`.  Then if that is not nil, we know we are in the middle of the registration process and can redirect accordingly.

```ruby
class User 
  def registration_redirect_url
    is_temp? ? nil : read_attribute(:registration_redirect_url)
  end
  ...
```

```ruby
handle_with(ContactInfosConfirm,
            complete: lambda {
              # If in the middle of registration, redirect to continue process
              redirect_url = @handler_result.outputs.contact_info.user.registration_redirect_url
              if redirect_url.present?
                redirect_to redirect_url,
                            notice: 'Thanks for adding your email address.'
              else
                render :confirm, status: @handler_result.errors.any? ? 400 : 200
              end
            })
```

jpslav: Actually, this alternate may be clearer (and doesn't require overriding the `registration_redirect_url` like I did above.

```ruby
handle_with(ContactInfosConfirm,
            complete: lambda {
              # If in the middle of registration, redirect to continue process
              user = @handler_result.outputs.contact_info.user
              if user.is_temp?
                redirect_to user.registration_redirect_url,
                            notice: 'Thanks for adding your email address.'
              else
                render :confirm, status: @handler_result.errors.any? ? 400 : 200
              end
            })
```

app/models/user.rb
```diff
@@ -93,7 +93,7 @@ def is_unclaimed?
   end
 
   def is_email_pending?
-    !email_addresses.any?(&:verified)
+    !email_addresses.any?(&:verified) && !email_addresses.empty?
   end
```
jpslav: Can we rename this method to make it more clear what it does?  "pending" is a little vague to me.  In fact I'm having a hard time figuring out exactly what it is saying.  Seems like it is returning true if and only if there is at least one email address and none of them are verified.  Maybe just saying `has_emails_pending_verification?` is enough to clear it up?  Hm.. that doesn't account for the added logic.  `has_emails_and_all_unverified?` ?  Better ideas?

karenc: We can change it to `has_verified_email?` and I can reverse the logic?

app/views/contact_infos/confirm.html.erb
```diff
@@ -1,6 +1,13 @@
...
 <% if @handler_result.errors.any? %>
-  <p>Sorry, we couldn't verify an email using the confirmation code you provided.  If you typed it in manually, please check for errors.  If you're still having trouble, you might try to send the confirmation email again.</p>
+  <p>Sorry, we couldn't verify an email using the verification code you provided.  If you typed it in manually, please check for errors.  If you're still having trouble, you might try to send the verification email again.</p>
+
+  <% if signed_in? %>
+    <p><%= button_to 'Resend Verification',
+                     resend_confirmation_contact_info_path(current_user.email_addresses.first),
+                     method: :put,
+                     class: 'standard' %></p>
+  <% end %>
```
jpslav: I know I asked you to do this, but it doesn't actually make sense b/c

1. there might not be a `current_user` (login not required to confirm an email address, which is correct)
2. The email they are trying to confirm might not be their first email.  

Can you instead make the message read:

> Sorry, we couldn't verify an email using the verification code you provided.  If you typed it in manually, please check for errors.  If you're still having trouble, please resend the verification email from your user profile.

... and make "your user profile" be a link to their profile page on the email tab.  If they are logged in, the link will take them right there.  If they aren't logged in, they'll be asked to log in.

config/initializers/controllers.rb
```diff
@@ -42,6 +42,7 @@ def registration
 
     return unless current_user.is_temp?
     store_url key: :registration_return_to
+    flash.keep
     redirect_to(current_user.is_email_pending? ? verification_sent_path : register_path)
```
jpslav: If the meaning of `is_email_pending?` is really `has_emails_and_all_unverified?`, this would mean that if a user has one verified email address and one unverified email address we won't let registration continue (but I think we should -- the minimum requirement is one email address verified).

app/models/user.rb
karenc: Wait.. or is `has_unverified_email?` ok?

karenc: Sorry that doesn't work at all...

jpslav: sounds good to me

jpslav:
Which part of the PR addresses "Fix twitter login (or users registered without any email addresses)"?

karenc:
Fix twitter login is this commit "User#is_email_pending? to return false if there are no email addresses".

jpslav:
Oh, but in the case where they don't have email addresses we don't want to send them on to the registration page -- we instead need to collect an email address from them (with a new "give us an email address" page).  Does that make sense? (if social doesn't provide an email, we still have a requirement of having an email so need to get it from the user)

app/models/user.rb
jpslav: obviously my "sounds good to me" was meant to be about your last two comments :-)

karenc:
What are we going to do about old users who registered without an email address?

jason-holmes:
I don't think there's a need or desire for grandfathering older users to have email addresses.

app/controllers/contact_infos_controller.rb
karenc: Instead of trying to guess the path I think `redirect_to :back` actually does the right thing so I'll do that.

jpslav:
@karenc -- what's the status of this PR? ready for review or more things to be done?

karenc:
The commit for fixing twitter login is now "Ask for an email address for users signing up using twitter".

karenc:
- After logging in using twitter, the user sees the "merge logins" page:

twitter1

- After the "merge logins" page, since the user has no email addresses, they see the "give us an email address" page:

twitter2

- Then the "verification sent" page:

twitter3

- After clicking on the verification link, they see the registration form:

twitter4

----

After this the user is redirected to tutor but they are **not** logged in.  They are logged in to accounts but not logged in to tutor.  I'm not sure why yet...  @jpslav Any ideas?

jpslav:
Which Url are they redirected back to on tutor?

karenc:
http://vivid.local:3001/accounts/auth/openstax/callback?code=95b04872e816a4bb6b25b5b3f6fe9cf4d027f7062dc03fd445cba26ac574c7f3&state=6f09d68bcca9d3e38673d51b0fffa6b43bf4d984c9056825

(http://vivid.local:3001 is where my tutor-server is)

karenc:
And the user just ends up on `/` instead of `/dashboard`.

karenc:
This might provide some clues (tutor-server development log):

```
Started GET "/accounts/auth/openstax/callback?code=[FILTERED]&state=6f09d68bcca9d3e38673d51b0fffa6b43bf4d984c9056825" for 192.168.122.1 at 2015-12-04 14:14:32 +0000
(openstax) Setup endpoint detected, running now.
(openstax) Callback phase initiated.
Processing by OpenStax::Accounts::SessionsController#callback as HTML
  Parameters: {"code"=>"[FILTERED]", "state"=>"6f09d68bcca9d3e38673d51b0fffa6b43bf4d984c9056825", "provider"=>"openstax"}
   (0.1ms)  BEGIN
   (0.1ms)  SET TRANSACTION ISOLATION LEVEL REPEATABLE READ
  OpenStax::Accounts::Account Load (0.3ms)  SELECT  "openstax_accounts_accounts".* FROM "openstax_accounts_accounts" WHERE "openstax_accounts_accounts"."openstax_uid" = $1 LIMIT 1  [["openstax_uid", 4]]
  OpenStax::Accounts::Account Exists (0.3ms)  SELECT  1 AS one FROM "openstax_accounts_accounts" WHERE "openstax_accounts_accounts"."openstax_uid" = 4 LIMIT 1
  OpenStax::Accounts::Account Exists (0.3ms)  SELECT  1 AS one FROM "openstax_accounts_accounts" WHERE "openstax_accounts_accounts"."username" = 'karenc4' LIMIT 1
   (0.2ms)  ROLLBACK
Redirected to http://vivid.local:3001/
Completed 302 Found in 13ms (ActiveRecord: 1.3ms)
```

nathanstitt:
@karenc I also received a similar error a few days ago with Accounts redirecting to Tutor after login but Tutor didn't pickup that the user was logged in, even though I had log messages similar to what you show in Tutor.  It was puzzling because the logs suggest that Tutor does know what account it is, but `current_user` remains unchanged and anonymous.

In my case, it stopped after I loaded the oauth applications in the accounts admin console and deleted Tutor, re-created it and then updated Tutor's secrets.yml with the new secrets.

I didn't really like doing that (or suggesting it to you) because it smacks of cargo cult debugging and the real error is probably elsewhere, but it at least worked and allowed me to focus on the "real" bug I was attempting to fix :(

karenc:
It seems to work for local users and facebook users though...

jpslav:
you might try resetting your DBs in case they got into a weird state mid-PR

nathanstitt:
That's probably a different bug then.  Mine affected all users, local or social. I'd also dropped my tutor db and re-ran several times while I was receiving it.   I chalked my bug up to that at the time but then saw your message and wondered if it was deeper.

nathanstitt:
Sorry - wrong button :(  Meant to hit cancel since JP suggested what I was going to say.

jpslav:
If you want you can put a debugger jn lev where the rollback happens (in
routine.rb) to see what is causing it

karenc:
I looked at the code in `OpenStax::Accounts::SessionsController#callback` and I think the problem was that I cleared the accounts database without clearing the tutor database.  I have used my twitter login before and so it probably couldn't create the account in tutor.  So this is just a set up error, the code should be ok.

jpslav:
Thanks @karenc!

PR #219, opened 2015-12-03
More concept coach login style tweaks

nathanstitt:
The "ask-new-or-returning page" was pretty janked at certain sizes, and the signup's boxes had quite a bit of space between them

This uses flexbox to align items since it'll adjust to arbitrary width / heights the best.  Note that it's IE 10+, but that's OK for concept coach audience.   As soon as accounts itself can drop legacy browser support we should move these styles over to the main site.

screen shot 2015-12-03 at 1 35 15 pm
screen shot 2015-12-03 at 1 25 39 pm

### DIFF
--- a/app/assets/stylesheets/iframe.css.scss
+++ b/app/assets/stylesheets/iframe.css.scss
@@ -28,13 +28,36 @@ body.iframe {
     margin: 0 auto 20px auto;
   }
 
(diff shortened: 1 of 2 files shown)


nathanstitt:
@jpslav this is ready for review/merge

PR #225, opened 2015-12-09
Remove duplicate username taken error message at sign up

karenc:
User sees "Username has already been taken" twice if the username chosen
at sign up is taken.  The reason is we have uniqueness check in the user
model twice because we changed from case sensitive unique usernames to
case insensitive unique usernames without affecting the validation of
existing users.

I've removed the case sensitive uniqueness check because I think it's
redundant.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,7 +27,6 @@ class User < ActiveRecord::Base
   has_many :oauth_applications, through: :member_groups
 
   validates :username, presence: true,
-                       uniqueness: { case_sensitive: true },
                        length: { minimum: 3, maximum: USERNAME_MAX_LENGTH },
                        format: { with: /\A[A-Za-z\d_]+\z/,
                                  message: "Usernames can only contain letters, numbers, and underscores." }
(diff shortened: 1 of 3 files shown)


karenc:
@Dantemss could you have a look at this?

jpslav:
+1 on @Dantemss taking a look -- I'm wondering if we still need that case sensitive check for when we import CNX users to real production Accounts?

Dantemss:
I think we need both... I would make one trigger only `if: :username_changed?` and the other trigger only `unless: :username_changed?`

Dantemss:
But of course the presence check should always happen.

jpslav:
I like Dante's idea.

karenc:
I'm not sure when we're doing the case sensitive check.  What test case should I add to https://github.com/openstax/accounts/blob/master/spec/models/user_spec.rb#L27?

karenc:
We have the following model tests:

- username must be unique (case-insensitive) on creation
- username cannot be updated to match (case-insensitive) an existing username
- username does not interfere with updates if duplicated but not changed

And I just added a test to import users from csv to make sure we can create users from cnx legacy site with same username but different cases.  

So I think we've covered everything.

Dantemss:
OK I think this is fine.

PR #232, opened 2015-12-15
Replace iframe based login with a popup based one

nathanstitt:


app/views/sessions/new.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <%= page_heading("Sign in to your one OpenStax account!".html_safe) %>
 <div class="new-login">
-  <%= render 'sessions/login', hide_signup: (session[:from_cnx] && !session[:from_iframe]) %>
+  <%= render 'sessions/login', hide_signup: session[:from_cnx] %>
```
nathanstitt: I'm a bit worried about this line.  We'll have to make sure we check that the signup option is displayed for CC.  If it isn't we'll have to figure out a different check to re-enable it.

jpslav:
Looks good.

PR #240, opened 2015-12-17
Allow CNX users to sign up

nathanstitt:
I left in the `:hide_signup` option on `sessions/_login` since it might be used elsewhere.

PR #241, opened 2015-12-18
Comment out line to match staging env manual edit

cnuber:
Staging env had a manual edit of this file.  updating to match.


karenc:
I think this is already covered by #240.

Dantemss:
Yep. #240 already does this. Closing.

PR #242, opened 2015-12-23
Login flow refinements and general refactoring

jpslav:
## General Refactoring

Reorganizes the registration actions into their own controller (removing them from the Users and StaticPages controllers).  Simplifies some of the redirection logic (more direct redirects, fewer action_interceptor-style redirects).

## Login Flow Refinements

Removes the sign up link from the header of the sign up page:

image

Makes sure that the "We haven't seen that you clicked the verification link" message really only appears when the user clicks the button but hasn't clicked the link.  Previously it could show up in some unintended situations.

image

Users can now click their username to access their profile *before* they've completed registration.  This will let them add other email addresses or delete the one they entered on the sign up screen (in case they mis-typed it)

When they click the username in this situation, they'll go to the profile page and directly to the email address tab.  There is a "Finish Sign Up!" button in the upper corner to return them to the registration process.

image

If they click the other tabs they see a message saying they need to complete sign up before they can see the forms:

image

Cleans up the "Manage Email Addresses" tab of the profile page:

image

### DIFF
--- a/app/access_policies/user_access_policy.rb
+++ b/app/access_policies/user_access_policy.rb
@@ -9,7 +9,7 @@ class UserAccessPolicy
     when :search
       requestor.is_application? || requestor.is_activated?
     when :read, :update
-      requestor.is_human? && requestor.is_activated? && \
+      requestor.is_human? && \
       (requestor == user || requestor.is_administrator?) # Self or admin
     when :register
       requestor.is_human? && !requestor.is_activated? && \
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,8 +1,10 @@
 class UsersController < ApplicationController
 
-  skip_before_filter :registration, only: [:register, :ask_for_email]
+  skip_before_filter :registration,
+                     only: [:register, :ask_for_email, :edit, :update, :i_clicked_verification_link]
 
-  fine_print_skip :general_terms_of_use, :privacy_policy, only: [:register, :ask_for_email]
+  fine_print_skip :general_terms_of_use, :privacy_policy,
+                  only: [:register, :ask_for_email, :edit, :update, :i_clicked_verification_link]
 
   def edit
     OSU::AccessPolicy.require_action_allowed!(:update, current_user, current_user)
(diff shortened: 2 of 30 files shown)


jpslav:
@jason-holmes please take a look at these screenshots and let me know if you'd like changes.

app/access_policies/user_access_policy.rb
```diff
@@ -9,7 +9,7 @@ def self.action_allowed?(action, requestor, user)
     when :search
       requestor.is_application? || requestor.is_activated?
     when :read, :update
-      requestor.is_human? && requestor.is_activated? && \
+      requestor.is_human? && \
```
jpslav: Makes it so users who haven't completed registration can update their profile (change their email addresses)

jpslav:
@karenc can you please review?

spec/access_policies/user_access_policy_spec.rb
```diff
@@ -24,20 +24,20 @@
   context 'read, update' do
     it 'cannot be accessed by anonymous, temp users or apps' do
```
karenc: Can you update the test name?

app/controllers/users_controller.rb
```diff
@@ -21,29 +23,15 @@ def update
...
-                    render :register, status: errors ? 400 : 200
-                  })
-    else
-      store_fallback key: :registration_return_to
-    end
-  end
-
   def ask_for_email
+    # FWIW, when this code is this resurrected, the registration_redirect_url
+    # may not be what we need.
```
karenc: What do you mean?

karenc:
Looks good to me

app/controllers/users_controller.rb
jpslav: When we ask for email in the future, I don't think it'll be part of the registration process -- it'll just be us seeing that someone who logged in does not have a registered email address and we'll want to solicit one.  This note here is just saying that `registration_redirect_url` might be misleading for future developers of that use case.

spec/access_policies/user_access_policy_spec.rb
jpslav: done!

jason-holmes:
These pages look OK to me.  Sorry about the delay...

PR #247, opened 2016-01-21
DON'T MERGE Added a spec to show a bad account merge situation in OpenStax Accounts

Dantemss:
The spec passes, showing that the teacher and student accounts are merged.


Dantemss:
Closing since we decided to get rid of account merging.

PR #250, opened 2016-01-28
Made the home page require the entire login process before it redirects you

Dantemss:
Spiritual successor to https://github.com/openstax/accounts/pull/171 (since now the home page always takes you to the login page if you are not already logged in or profile page if you are).

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,10 +1,8 @@
 class UsersController < ApplicationController
 
-  skip_before_filter :registration,
-                     only: [:ask_for_email, :edit, :update]
+  skip_before_filter :registration, only: [:edit, :update, :ask_for_email]
 
-  fine_print_skip :general_terms_of_use, :privacy_policy,
-                  only: [:ask_for_email, :edit, :update]
+  fine_print_skip :general_terms_of_use, :privacy_policy, only: [:edit, :update, :ask_for_email]
 
   def edit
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -1,10 +1,8 @@
 class UsersController < ApplicationController
 
-  skip_before_filter :registration,
-                     only: [:ask_for_email, :edit, :update]
+  skip_before_filter :registration, only: [:ask_for_email]
```
jpslav: Why did `:edit` and `:update` go away here?  I think they are needed because the edit screen is now open _during_ registration (whereas it didn't used to be).

Dantemss: I *think* that this controller action is only for the edit profile page. The registration uses a different one.

jpslav: I agree, but you can get to the edit profile page now while you're still in the middle of registration (when you are not fully registered), so I think that if these are removed and you try to go to the profile page during registration you'll be taken back to registration.... probably should have written some specs (doh)

Dantemss: OK so maybe this is OBE...

Dantemss: Or I can just change the filters back and see if it works hahaha

jpslav:
Will merge this after Wednesday so it doesn't make it into this production release and because we are doing a lot of Accounts dev next week so we can fold this in with that work and testing.

PR #257, opened 2016-02-23
Allow resetting password without verified emails

jpslav:
Per customer support, we want to let users who have not verified their password reset their password.  There's a danger we could spam some people here, but there isn't a security problem.


jpslav:
@karenc please review

PR #263, opened 2016-03-05
Fixed the 500 error on double register and added some specs

Dantemss:
You now get this message:

screen shot 2016-03-04 at 7 03 29 pm

PR #264, opened 2016-03-08
Alternate: Fixed the 500 error on double register

jpslav:
@Dantemss instead of writing some comments on https://github.com/openstax/accounts/pull/263 I decided to put it in a PR.  Instead of letting the error bubble up from a lower-level routine, I'd prefer to be more explicit in `IdentitiesRegister` that only one identity is allowed (and `fatal_error` out immediately if that's not the case).  This approach also changes the flow to continue to the "merge logins" page with an error flash.

### DIFF
--- a/app/handlers/identities_register.rb
+++ b/app/handlers/identities_register.rb
@@ -32,7 +32,7 @@ class IdentitiesRegister
     user = caller
 
     if user.identity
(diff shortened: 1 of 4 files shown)


jpslav:
@Dantemss please take a look at this and let me know if you approve of this alternate approach.

PR #270, opened 2016-03-16
Fixed more specs, moved can't login pieces, cleanup

jpslav:
@karen, I pulled in the commits from your PR because I needed to modify some of that code.  Here's a rundown of what I changed:

* Previously we had a mix of "log in" / "log out" and "sign in" / "sign out".  I have changed everything (?) to be "sign in" / "sign out" for consistency.
* Moved your `login_help` stuff into `sessions#help` with a `/signup/help` route.  Wasn't enough of a `User` thing to go in that controller, and for some reason the underscores in the route was bugging me (even tho I have used them elsewhere).  
* Removed "Sign up" from the header, because there are now two sign up paths (social or password)
* Fixed a number of specs.

I will also go through the diffs here and make some inline comments.

Feel free to review and "merge".  I'm going to tackle some of the more complex specs in the morning. 

## Update

I either fixed or disabled the remaining failing specs.  Some we can't fix until we reimplement things like "change your password", "update profile", etc.  There are some migrations in here, fyi.


spec/features/hide_sign_up_from_cnx_users_spec.rb
```diff
@@ -1,12 +1,15 @@
 require 'rails_helper'
 
+# TODO KC - I think you would have written this way back -- any recollection
+# of what this spec is trying to show?  I have temporarily disabled the scenario
```
jpslav: cc @karenc

karenc: This is the original test: https://github.com/openstax/accounts/commit/cb738c341bf0d7d633f575964eeb2de2a3c06b62

This commit removed most of it: https://github.com/openstax/accounts/commit/69ddd25c8a88cfce6bccdb5faeac6863199b672b

I think it's safe to say that we don't need this test anymore.

karenc:
Merged.

PR #273, opened 2016-03-19
Removed `User#full_name` DB attr, left method to build full name on demand.

jpslav:
Users can still be created with `full_name`s (e.g. via `CreateUser`) but the value is only used to guess what the first and last names are if they aren't given (`first_name` and `last_name` are the remaining database name fields).  There is also still a `full_name` method on `User`, but it now just builds the name on demand.

Also, fixed the bug where rake was returning `0` when specs failed (the coveralls gem needed to be updated -- the bug was there).


spec/lib/import_users_spec.rb
```diff
@@ -25,12 +25,12 @@
   end
 
   it 'creates users from a csv file' do
-    headers = [:row_number, :username, :password_digest, :title, :first_name, :last_name, :full_name, :email_address]
```
karenc: We might need to use the suffix field if we need to import users from Connexions again if we're getting rid of `full_name`.

@reedstrm Are we still going to import users from Connexions?

jpslav: We could put `full_name` back in the importer but would need a way to split out suffixes.

PR #275, opened 2016-03-22
Change import users from csv to generate username if it's missing

karenc:
This is for importing users from legacy osc site.

### DIFF
--- a/lib/import_users.rb
+++ b/lib/import_users.rb
@@ -10,16 +10,18 @@ class ImportUsers
   def initialize(csv_file, app_id)
     @csv_file = csv_file
     @app_id = app_id
+    @usernames = []
   end
 
   def read
     # Output results to import_users_results.{timestamp}.csv
-    output_headers = [:row_number, :old_username, :new_username, :errors]
+    output_headers = [:row_number, :email_address, :old_username, :new_username, :errors]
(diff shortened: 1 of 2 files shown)


lib/import_users.rb
```diff
@@ -56,7 +58,13 @@ def read
   end
 
   def create_user(username, password_digest, title, first_name, last_name, full_name, email_address)
-    person = Person.create!
+    # Check whether the user is already in the database
+    ea = EmailAddress.find_by_value(email_address)
```
jpslav: should we limit to just verified email addresses?  `EmailAddress.verified.where{value: email_address}`

lib/import_users.rb
```diff
@@ -86,4 +94,24 @@ def create_user(username, password_digest, title, first_name, last_name, full_na
     ea.verified = true
     ea.save!
   end
+
+  def generate_username(first_name, last_name, email)
+    # Use first name and last name as the basis for generating username
+    base = "#{first_name} #{last_name}".strip
```
jpslav: is there a reason for the space between first and last name?  gets `gsub`bed out later.

lib/import_users.rb
jpslav: if we don't use verified emails, there's an outside chance that we'd link different people together.

jpslav:
Review complete

karenc:
Updated.

PR #278, opened 2016-03-23
Add "last login" check sign on sign in page

karenc:
last-login

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -39,6 +39,7 @@ class SessionsController < ApplicationController
       complete: lambda {
         case @handler_result.outputs[:status]
         when :returning_user, :new_password_user, :transferred_authentication
+          set_last_signin_provider(@handler_result.outputs[:authentication].provider)
           redirect_to action: :returning_user
         when :new_social_user
           redirect_to signup_social_path
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -22,10 +22,6 @@ class SessionsController < ApplicationController
     # Attempt to redirect to the fallback url stored above
     redirect_back if signed_in?
 
-    # Hack to figure out if the user came from CNX to hide the login
-    # In the future, use the client_id and some boolean flag in the client app
-    referer = request.referer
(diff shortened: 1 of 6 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -22,10 +22,6 @@ def new
     # Attempt to redirect to the fallback url stored above
     redirect_back if signed_in?
 
-    # Hack to figure out if the user came from CNX to hide the login
-    # In the future, use the client_id and some boolean flag in the client app
-    referer = request.referer
-
```
jpslav: thanks for the cleanup!

app/controllers/sessions_controller.rb
```diff
@@ -41,6 +37,7 @@ def callback
 
     handle_with(SessionsCallback, user_state: self,
       complete: lambda {
+        cookies.signed[:authentication_provider] = @handler_result.outputs[:authentication_provider]
```
jpslav: I think we need to only set this in certain cases.  I think this is good to call for the `when :returning_user, :new_password_user, :transferred_authentication` case.  For `:new_social_user`, the user isn't all the way created yet (as you can see on line 45 they are redirected to the social signup page).  I'd rather only set the `:authentication_provider` right before [this line](https://github.com/openstax/accounts/blob/spring-16-flow/app/controllers/signup_controller.rb#L25).  In that `success` block, the user will always only have one authentication, so you could just query it with `current_user.authentications.first.provider`.

Since it will be called in two places from two controllers, can you add a helper method in the base controller (can do in `/config/initializers/controllers.rb`) that takes the provider string and sets the cookie value?

Also, can you change the cookie key to `:last_signin_provider`?

app/handlers/sessions_callback.rb
```diff
@@ -76,6 +76,7 @@ def handle
     end
 
     outputs[:status] = status
+    outputs[:authentication_provider] = authentication.provider
```
jpslav: could just store the authentication and call `.provider` later (in case it becomes useful for someone else to have it.  obviously not big deal tho

karenc:
Updated.

PR #279, opened 2016-03-23
Hook up all email edit controls and enable editing after creating new email

nathanstitt:
Mostly code organization into modules.   Enables the searchable and delete actions and handles creating an email then editing it immediatly.


app/views/users/edit.html.erb
```diff
@@ -106,3 +72,37 @@
...
+    title: "<%= current_user.title %>",
+    first_name: "<%= current_user.first_name %>",
+    last_name: "<%= current_user.last_name %>",
+    suffix: "<%= current_user.suffix %>"
+  },
+  success: function(response, newValue) {
+    $(this).html(response.full_name);
+  }
+});
+</script>
```
jpslav: Should these calls move to assets JS?

nathanstitt: Yes, and I started to do so, but then ran into the issue that they use `erb` interpolation that's not available on an asset, only a view.  Well I suppose an asset can use it, but it doesn't have access to the current user, since the `ERB` is only ran once at compilation :grin:

PR #283, opened 2016-03-23
Change password using inline editing

nathanstitt:
screen shot 2016-03-23 at 8 11 06 pm

screen shot 2016-03-23 at 8 11 16 pm

screen shot 2016-03-23 at 8 11 29 pm


app/controllers/identities_controller.rb
```diff
@@ -19,8 +19,10 @@ def new
   def update
     @active_tab = :password
```
jpslav: Can you ditch this line? (old form)

PR #287, opened 2016-03-24
Made emails all HTML and added autogenerated text email parts

jpslav:
* Made emails all HTML
* added autogenerated text email parts
* Made "can't sign in" email work for multiple users and multiple emails per user.


app/handlers/sessions_help.rb
```diff
@@ -16,15 +16,52 @@ def authorized?
...
+
+  def matching_users
+    if @matching_users.nil?
+      username_or_email = help_params.username_or_email.strip
+
+      @matching_users = matching_users_by_username(username_or_email) ||
+                        matching_users_by_email(username_or_email)
+
+      if @matching_users.blank?
+        fatal_error(code: :user_not_found,
```
nathanstitt: I think this code will aways be true and then raise a `fatal_error`. Since the if statement block is checking `.nil?`, then `.blank?` will also be true since it will match both `.nil?` and `.empty?`

app/handlers/sessions_help.rb
```diff
@@ -16,15 +16,52 @@ def authorized?
...
+    @matching_users
+  end
+
+  def matching_users_by_username(username)
+    user = User.find_by_username(username)
+
+    return nil if user.nil?
+
+    email_addresses = user.contact_infos.email_addresses.map(&:value)
+
```
nathanstitt: It's a slight optimization to use `pluck` because then that only queries the columns that were specified rather than slurping all columns into AR models and then calling `map`.

Could be:
`email_addresses = user.contact_infos.email_addresses.pluck(:value)`

app/handlers/sessions_help.rb
```diff
@@ -32,8 +69,23 @@ def handle
...
+    [{
+      user: user,
+      email_addresses: email_addresses
+    }]
+  end
+
+  def matching_users_by_email(email)
+    users = ContactInfo.where(value: email).all.collect(&:user)
+
```
nathanstitt: This is going to load all the matching contact infos, then fire a select for the user model of each one.  Would be more efficient (and arguably more clear) to: `User.joins(:contact_infos).where({contact_infos: {value: email}})`

app/handlers/sessions_help.rb
jpslav: only `true` if the calls on lines 45 and 46 don't return non-nil, right?

app/handlers/sessions_help.rb
nathanstitt: Actually.. neither of our implementations will work.  `contact_infos` is a `has_many`, and the there's not a `email_addresses` method/attribute.  So to get all the email addresses for a user it's probably best to:

`email_addresses = user.contact_infos.pluck('value')`

jpslav: thanks, made change.

app/handlers/sessions_help.rb
jpslav: thanks, made change.

app/handlers/sessions_help.rb
nathanstitt: Ah, right you are.  I missed that ` @matching_users` was re-assigned there and was only looking at the `if` on line 42. Sorry for the false alarm :)

PR #288, opened 2016-03-24
moved test-unit out of :test group to make production work

jpslav:


jpslav:
Merge after specs pass

PR #289, opened 2016-03-24
Enable / disable different authentication types

nathanstitt:
screen shot 2016-03-24 at 4 02 18 pm


app/handlers/identities_update.rb
```diff
@@ -21,6 +21,13 @@ def authorized?
   def handle
     identity_attributes = identity_params.as_hash(:password,
                                                   :password_confirmation)
+
+    # This may be better moved elsewhere, but we'll do so here since it doesn't make
+    # sense to set the password of a identity that won't work without an authentication
+    unless caller.authentications.where(provider: 'identity').any?
+      caller.authentications.create!(provider: 'identity', uid: caller.id)
+    end
+
```
nathanstitt: Not sure if t's a good idea to auto-create an authentication for an email if it's missing.  By doing so, the FE code can just submit a password and have it "just work".  If it didn't then the code would have to fire two requests, or hit a different end-point.

PR #295, opened 2016-03-25
styling for alerts 

nathanstitt:
Hide success alerts automatically after 15 secs
vertical align content
Hide existing alerts before editing password

Note: vertical alignment is now using "baseline".  Center may be better?
screen shot 2016-03-25 at 3 57 15 pm

### DIFF
--- a/spec/features/user_updates_password_spec.rb
+++ b/spec/features/user_updates_password_spec.rb
@@ -8,6 +8,10 @@ feature 'User updates password', js: true do
     expect(page).to have_content('Welcome, user')
   end
 
+  after(:each) do
+    wait_for_ajax
+  end
+
   context 'without local password' do
     before(:each) do
       user = User.find_by_username('user')


spec/rails_helper.rb
```diff
@@ -32,6 +45,8 @@
   # config.mock_with :flexmock
   # config.mock_with :rr
 
+  config.include WaitForAjax, type: :feature
+
```
nathanstitt: I've had good success triggering the error by running rspec with a set seed:
`bundle exec rspec ./spec/features/user_updates_password_spec.rb --seed 15515`

I also attempted to install a global `after(:each)` by adding:

```ruby
  config.after(:each, type: :feature, js: true) {
    puts 'AFTER EACH'
    wait_for_ajax
  }
```

The blog is executed, but the error still persists.  However when I add the `wait_for_ajax` call to the the spec directly it seems to suppress the error.

jpslav: Yay!

PR #296, opened 2016-03-25
Re-enable sessions callback specs

karenc:
- Only return authentication_taken if the user linked to the authentication is activated
- If authentication is linked to the current user and the user is
      already signed in, there's nothing to do
- If the user linked to the authentication is not activated, transfer
      the authentication to the signed in user and delete the other user

### DIFF
--- a/app/handlers/sessions_callback.rb
+++ b/app/handlers/sessions_callback.rb
@@ -18,6 +18,7 @@
 # :new_social_user              if the user is signing up and just authenticated socially
 # :transferred_authentication   if the user signed up and we can find an existing user to add the auth to
 # :authentication_added         if the user is adding an authentication from the profile page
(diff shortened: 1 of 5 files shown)


app/handlers/sessions_callback.rb
```diff
@@ -45,12 +45,24 @@ def handle
       Authentication.find_or_create_by_provider_and_uid(@data.provider, @data.uid.to_s)
     authentication_user = authentication.user
 
-    if signed_in?
+    if signed_in? && authentication_user == current_user
+      status = :no_action
```
jpslav: Can you add a comment at top about this status (in the list of statuses)?

jpslav:
Amazing spec @karenc

karenc:
Moved destroy user to transfer authentications routine.

PR #298, opened 2016-03-29
Add data (e.g. email) from social networks when adding auths on profile

karenc:


jpslav:
Thanks @karenc -- can you add a spec when you have a chance?

karenc:
^ spec added :D

PR #299, opened 2016-03-31
Add rake task to archive temp users

karenc:


karenc:
I'm not sure why the user is deleted but the authentications are not:

```
   1) ArchiveTempUsers saves user data to a file and deletes them
      Failure/Error: expect(Authentication.where(user_id: user_w_facebook.id)).to be_empty
        expected `[#<Authentication id: 218, user_id: 5029, provider: "facebook", uid: "7eb4cd", created_at: "2016-03-30 22:54:01", updated_at: "2016-03-30 22:54:01">].empty?` to return true, got false
      # ./spec/lib/archive_temp_users_spec.rb:120:in `block (2 levels) in <top (required)>'
```

PR #309, opened 2016-04-27
Email verification by PIN

jpslav:
Provides new APIs for getting contact info and confirming (verifying) contact infos.

* Contact info information provided in `GET /api/user` (see API docs)
* `PUT /api/contact_infos/:id/resend_confirmation` (see API docs)
* `PUT /api/contact_infos/:id/confirm` (see API docs)

## Misc
* Updates `lev` to `2.2.1` to get support for `after_transaction` blocks.
* Monkey patched in ActiveRecord enum support from Rails 4.2 :-)
* Added a `to have_routine_error_code` and `to have_routine_error_status` rspec helpers.
* Added should-matchers

### DIFF
--- a/app/routines/confirm_by_pin.rb
+++ b/app/routines/confirm_by_pin.rb
@@ -1,0 +1,41 @@
+class ConfirmByPin
+  lev_routine
+
+  uses_routine ConfirmContactInfo
+
+  MAX_PIN_FAILURES = 5
+
+  def self.sequential_failure_for(contact_info)
+    SequentialFailure.confirm_by_pin
+                     .find_or_initialize_by_reference(contact_info.value).tap do |sf|
+      sf.num_failures_allowed = MAX_PIN_FAILURES
+    end
+  end
+
+  protected
+
+  def exec(contact_info:, pin:)
+    return if contact_info.confirmed?
+
+    sequential_failure = self.class.sequential_failure_for(contact_info)
+
+    if !sequential_failure.attempts_remaining?
+      fatal_error(code: :no_pin_confirmation_attempts_remaining)
+    else
+      if contact_info.confirmation_pin == pin
+        after_transaction do
+          sequential_failure.reset!
+        end
+
+        run(ConfirmContactInfo, contact_info)
+      else
+        after_transaction do
+          sequential_failure.increment!
+        end
+
+        fatal_error(code: :pin_not_correct)
+      end
+    end
+  end
+
+end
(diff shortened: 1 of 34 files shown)


jpslav:
@nathanstitt please review when you have a chance

spec/access_policies/contact_info_access_policy_spec.rb
```diff
@@ -35,19 +35,4 @@
...
-                                               contact_info)).to eq true
-      expect(OSU::AccessPolicy.action_allowed?(:confirm, temp,
-                                               contact_info)).to eq true
-      expect(OSU::AccessPolicy.action_allowed?(:confirm, user,
-                                               contact_info)).to eq true
-      expect(OSU::AccessPolicy.action_allowed?(:confirm, admin,
-                                               contact_info)).to eq true
-    end
-  end
-
```
nathanstitt: Were these specs moved into the controller spec below? If not, should we have retained them?

lib/active_record_enum.rb
```diff
@@ -0,0 +1,235 @@
...
+          }
+        elsif !klass_method && method_defined_within?(method_name, _enum_methods_module, Module)
+          raise ArgumentError, ENUM_CONFLICT_MESSAGE % {
+            enum: enum_name,
+            klass: self.name,
+            type: 'instance',
+            method: method_name,
+            source: 'another enum'
+          }
+        end
```
nathanstitt: This is a neat use of `%` to fill in the `ENUM_CONFLICT_MESSAGE` messages.  It's really just `sprintf` but I like how it reads. I know, you just copied it in :)

app/routines/confirm_by_pin.rb
```diff
@@ -0,0 +1,41 @@
...
+        after_transaction do
+          sequential_failure.reset!
+        end
+
+        run(ConfirmContactInfo, contact_info)
+      else
+        after_transaction do
+          sequential_failure.increment!
+        end
+
```
nathanstitt: I'm probably missing something, but why are call `.reset!` and `.increment!` in an `after_transaction`   Doesn't that mean that those methods may fail silently without causing the lev routine to fail?  Maybe we want that behavior for some reason though.

nathanstitt:
@jpslav looks great!  I'm guessing the `after_transaction` blocks are needed but thought I'd point them out just in case.

spec/access_policies/contact_info_access_policy_spec.rb
jpslav: Confirmation by code is always allowed -- previously the access policy just returned true but it wasn't even called (the handler just said `def authorized? true; end`) so I took out the code and this spec that tested it.

app/routines/confirm_by_pin.rb
jpslav: The `sequential_failure.reset!` call should probably be inside the routine's transaction, but the `increment!` call needs to be afterwards.  The reason is that the `fatal_error` call will rollback the routine transaction which will undo the increment (which we don't want).

PR #312, opened 2016-05-05
Update openstax_api to 6.1.5

nathanstitt:


nathanstitt:
I'm merging this so it can be deployed (runback is failing to deploy from a sha).  Pretty sure everyone's ok with it since @Dantemss merged the real code in `openstax_api` for this purpose

PR #313, opened 2016-05-06
remove explicit doorkeeper require

nathanstitt:
The file no longer exists, and the require doesn't seem to be needed


jpslav:
Thanks NS

PR #328, opened 2016-07-19
Handle any invalid characters from omniauth request

BryanEliDimas:
This PR contains a valid change to how usernames should be handled.

Previously, invalid characters were stripped from the username and we attempted to use the filtered name. However, that code contained a bug when the filtered username turned out to be non-unique.

To correct for this, any time the given username is invalid in any way, a new, unique, and random username is created.

@jpslav

### DIFF
--- a/app/routines/create_user.rb
+++ b/app/routines/create_user.rb
@@ -30,36 +30,22 @@ class CreateUser
   end
 
   def generate_unique_valid_username(username)
-    return username if username_is_valid?(username) && username_is_unique?(username)
+    return username if username_is_valid?(username) && User.new(username: username).username_is_unique?
 
     username_max_attempts = 10
 
     username_max_attempts.times do
-      username = create_random_username(7)
-      return username if username_is_unique?(username)
+      username = User.new.create_random_username(7)
+      return username if User.new(username: username).username_is_unique?
     end
 
-    return create_random_username(12) # last-ditch effort
-
-    raise "could not create a unique username after 10 tries"
+    raise "could not create a unique username after #{username_max_attempts} tries"
   end
 
   def username_is_valid?(username)
-    return false if username.nil?
-
-    good_length = (username.length >= User::USERNAME_MIN_LENGTH) && (username.length <= User::USERNAME_MAX_LENGTH)
-    all_valid_chars = username.match(User::USERNAME_VALID_REGEX)
-
-    username_is_valid = good_length && all_valid_chars
-  end
-
-  def username_is_unique?(username)
-    return false if username.nil?
-    User.where('LOWER(username) = ?', username).none?
-  end
-
-  def create_random_username(num_digits_in_suffix)
-    "user#{rand(10**num_digits_in_suffix).to_s.rjust(num_digits_in_suffix,'0')}"
+    user = User.new(username: username)
+    user.valid?
+    user.errors[:username].none?
   end
 
   def guessed_first_name(full_name)
--- a/app/routines/create_user.rb
+++ b/app/routines/create_user.rb
@@ -35,9 +35,11 @@ class CreateUser
     username_max_attempts = 10
 
     username_max_attempts.times do
-      username = create_random_username
+      username = create_random_username(7)
       return username if username_is_unique?(username)
     end
+
+    return create_random_username(12) # last-ditch effort
 
     raise "could not create a unique username after 10 tries"
   end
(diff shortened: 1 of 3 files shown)


app/routines/create_user.rb
```diff
@@ -41,19 +29,33 @@ def exec(username:, title: nil, first_name: nil, last_name: nil,
...
+    return username if username_is_valid?(username) && username_is_unique?(username)
 
-    # Go ahead and sanitize the username knowing that User.create will
-    # expect it to adhere to certain rules on length and content.
-    base = base.gsub(User::USERNAME_DISCARDED_CHAR_REGEX, '')
-               .slice(0..User::USERNAME_MAX_LENGTH - 1).downcase
+    loop do
+      username = create_random_username
+      break if username_is_unique?(username)
+    end
```
jpslav: Do we need, for safety's sake, any mechanism to ward off malicious user creations that try to get us stuck in an infinite loop here?

jpslav:
Looks good.  One comment.

app/routines/create_user.rb
BryanEliDimas: The loop breaks out of it on our terms—only once we create a unique username using rand(1000000). The loop breaking out is not dependent on any user input. Correct me if I'm wrong.

app/routines/create_user.rb
```diff
@@ -41,19 +29,44 @@ def exec(username:, title: nil, first_name: nil, last_name: nil,
...
+
+    username
+  end
+
+  def max_attempts_exceeded?
+    @attempts_count ||= 0
+    @attempts_count += 1
+    did_exceed = @attempts_count > username_max_attempts
+    did_exceed
+  end
```
jpslav: Don't have to, but can reduce to one line 

`(@attempts_count ||= 0) += 1 > username_max_attempts`

If you prefer the existing, i can be ok with that too.

jpslav: On second thought, I think I'd prefer this logic to be directly in the generate method.  While I'm normally a fan of private methods to make things more clear and readable, since this one is actually mutating state (and doing more than just checking if max attempts exceeded), I'd vote for this to be just inline in the method.

kevinburleigh75: @BryanEliDimas I agree with @jpslav: `max_attempts_exceeded?` should probably be inlined back into `generate_unique_valid_username`.  Also, if the number of attempts exceeds the max and `ensure_no_errors` is set, it seems like we should raise an exception instead of just breaking the loop since that is the contract this routine implies.

kevinburleigh75: Actually, it looks like `generate_unique_valid_username` is only called when `ensure_no_errors` is set, so really that loop should always succeed or raise an exception.

app/routines/create_user.rb
```diff
@@ -41,19 +29,23 @@ def exec(username:, title: nil, first_name: nil, last_name: nil,
...
+
+    raise "could not create a unique username after #{username_max_attempts} tries"
   end
 
-  def randomify_username(base)
-    base = 'user' if base.blank?
-    "#{base}#{rand(1000000)}"
+  def username_is_valid?(username)
+    user = User.new(username: username)
+    user.valid?
```
jpslav: This line does the presence check, so we don't need `username_is_unique?`

app/routines/create_user.rb
```diff
@@ -41,19 +29,23 @@ def exec(username:, title: nil, first_name: nil, last_name: nil,
...
+    user.errors[:username].none?
   end
```
jpslav: Move this method into `User` as a static method `def self.username_is_valid?(username)`

app/routines/create_user.rb
```diff
@@ -41,19 +29,23 @@ def exec(username:, title: nil, first_name: nil, last_name: nil,
...
 
-    # Go ahead and sanitize the username knowing that User.create will
-    # expect it to adhere to certain rules on length and content.
-    base = base.gsub(User::USERNAME_DISCARDED_CHAR_REGEX, '')
-               .slice(0..User::USERNAME_MAX_LENGTH - 1).downcase
+    username_max_attempts = 10
+
+    username_max_attempts.times do
+      username = User.new.create_random_username(7)
+      return username if User.new(username: username).username_is_unique?
```
jpslav: replace with `return username if User.username_is_valid?(username)`

app/routines/create_user.rb
jpslav: Or, said better, it delegates all validity checks to `User` which can decide if usernames are required.

jpslav:
Thanks @BryanEliDimas!

BryanEliDimas:
I ran out of positive emojis—but thank you for all the help and for merging! @jpslav

PR #329, opened 2016-07-26
Rate limit logins by username and IP address

Dantemss:
Currently set to 100 username login failures and 10,000 IP login failures


app/controllers/sessions_controller.rb
```diff
@@ -129,19 +129,21 @@ def help
 
   # Omniauth failure endpoint
   def failure
-    security_log :sign_in_failed, reason: params[:message]
```
jpslav: Why did this line go away?

spec/features/rate_limit_logins_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(page).to have_content('You have made too many login attempts recently.')
+
+      visit '/'
+      expect_sign_in_page
+
+      fill_in 'Username', with: 'user'
+      fill_in 'Password', with: 'password'
+      click_button 'Sign in'
+      expect(page).to have_content('You have made too many login attempts recently.')
+    end
```
jpslav: Can you timecop and show that the lockout is lifted after an hour?

app/controllers/sessions_controller.rb
Dantemss: Moved to the CustomIdentity strategy. Basically CustomIdentity redirects you here, but a bot could just not follow the redirect and prevent the log from being created. So I moved it to when the password is actually checked.

spec/features/rate_limit_logins_spec.rb
Dantemss: Oh sure forgot about that test :)

PR #330, opened 2016-07-26
Log out Admins after 30 minutes of inactivity

BryanEliDimas:
If a user with administrator privileges steps away from their computer, a malicious user could take advantage of admin access to every feature. 

This feature is supposed to log admins out after **30 minutes of non-admin activity** for increased security. We don't want to annoy them when they are just using non-admin functionality.

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -6,13 +6,25 @@ module Admin
 
     include FakeExceptionHelper
 
-    if Rails.env.production?
-      before_filter :authenticate_admin!
-    else
+    if Rails.env.development?
       skip_before_filter :authenticate_user!
       skip_before_filter :finish_sign_up
 
       fine_print_skip :general_terms_of_use, :privacy_policy
+    else
+      before_filter :authenticate_admin!
+      before_filter :log_out_inactive_admins
+    end
+
+    def log_out_inactive_admins
+      if current_user.is_administrator?
+        if session[:last_admin_activity].to_time <= 30.minutes.ago
+          sign_out!
+          authenticate_admin!
+        else
+          session[:last_admin_activity] = DateTime.now.to_s
+        end
+      end
     end
 
     def cron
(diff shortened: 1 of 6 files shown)


jpslav:
Looks good.  Rebase and then we'll merge.

jpslav:
Oh, also do remember to make this for real production only, e.g. a `request.host` check on `accounts.openstax.org`.

app/controllers/admin/base_controller.rb
```diff
@@ -6,13 +6,25 @@ class BaseController < ApplicationController
...
+    else
+      before_filter :authenticate_admin!
+      before_filter :log_out_inactive_admins
+    end
+
+    def log_out_inactive_admins
+      if current_user.is_administrator?
+        if session[:last_admin_activity].to_time <= 30.minutes.ago
+          sign_out!
+          authenticate_admin!
```
Dantemss: Doesn't matter too much, but I think you could remove this line if you put this check before authenticate_admin! in the filter list.

Gemfile
```diff
@@ -85,3 +85,4 @@ group :production do
 end
 
 gem "codeclimate-test-reporter", group: :test, require: nil
+gem "timecop", group: :test
```
Dantemss: Remove this after rebase (I added timecop in the development, test group so it will show up after the rebase and we don't want gems twice in the Gemfile).

Dantemss:
Looks good. 2 comments above.

app/controllers/admin/base_controller.rb
BryanEliDimas: I agree, the first authenticate_admin! call is not necessary, but if I do remove the first one then I should rename log_out_inactive_admins to something more descriptive such like authenticate_active_admin.

jpslav: I'd say let's leave, b/c it is fairly clear as is.

BryanEliDimas: Ok.

PR #332, opened 2016-07-29
Some rate limit improvements

Dantemss:
- Reduce max failed login attempts per user to 10
- Reset per-user login attempts on successful login (e.g. password reset)
- Link to password reset page in the error message

### DIFF
--- a/lib/omniauth/strategies/custom_identity.rb
+++ b/lib/omniauth/strategies/custom_identity.rb
@@ -12,7 +12,7 @@ module OmniAuth
     class CustomIdentity
 
       LOGIN_ATTEMPTS_PERIOD = 1.hour
-      MAX_LOGIN_ATTEMPTS_PER_USER = 100
+      MAX_LOGIN_ATTEMPTS_PER_USER = 10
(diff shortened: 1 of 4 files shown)


lib/omniauth/strategies/custom_identity.rb
```diff
@@ -49,17 +49,26 @@ def cookies
...
         return true if ip_attempts >= MAX_LOGIN_ATTEMPTS_PER_IP
 
         user = locate_conditions[:user]
-        user_attempts = user.nil? ? 0 : security_log_relation.where(user: user).count
+        if user.nil?
+          user_attempts = 0
+        else
+          last_login_time = SecurityLog.sign_in_successful.maximum(:created_at)
+          user_attempts_time = last_login_time.nil? ? ip_attempts_time :
+                                                      [ip_attempts_time, last_login_time].max
```
jpslav: shorten to `[ip_attempts_time, last_login_time || 0].max` ?

Dantemss: Can't, actually... It complains about being unable to compare DateTime with Fixnum.

PR #335, opened 2016-08-02
Bug fix in CreateUser routine and better test coverage

BryanEliDimas:
This PR is an attempt make better specs and to make sure that routine CreateUser does what it's supposed to do.

Previously:
- CreateUser routine would still be able to return errors even when `ensure_no_errors` was `true`. 
- There was no test coverage to make sure that non-admins cannot access admin features.

Now:
- When `ensure_no_errors` is `true` in routine CreateUser, an exception is returned instead when validations fail and we cannot generate or determine a valid input (as we currently do for `username` and `first_name` and `last_name`.
- There is better test coverage in `spec/features/log_out_inactive_admins.rb`
- To do the fixes above, I was forced to change the settings for `rescue_from` which are in an initializer. Previous behavior was that it would raise exception when `Rails.application.config.consider_all_requests_local` was true. Now, it raises exception only in development environment `Rails.env.development?`

### DIFF
--- a/app/controllers/signup_controller.rb
+++ b/app/controllers/signup_controller.rb
@@ -1,7 +1,7 @@
 
 class SignupController < ApplicationController
 
-  skip_before_filter :authenticate_user!
+  skip_before_filter :authenticate_user!, only: [:index, :password]
   skip_before_filter :finish_sign_up
 
   fine_print_skip :general_terms_of_use, :privacy_policy
--- a/spec/controllers/signup_controller_spec.rb
+++ b/spec/controllers/signup_controller_spec.rb
@@ -1,0 +1,35 @@
+require "rails_helper"
+
+RSpec.describe SignupController, type: :controller do
+  let(:params) do
+    {
+      signup: {
+        i_agree: true,
+        username: 'sheep',
+        title: 'Miss',
+        first_name: 'Little',
+        last_name: 'Sheep',
+        email_address: 'sheep@example.org',
+        contract_1_id: FinePrint::Contract.first.id,
+        contract_2_id: FinePrint::Contract.last.id }
+    }
+  end
+
+  let(:user) { FactoryGirl.create :temp_user }
+  let!(:authentication) { FactoryGirl.create :authentication, user: user }
+
+  context "POST social" do
+    it "creates a new user with a social network and redirects on success" do
+      controller.sign_in! user
+      post :social, params
+
+      expect(response).to redirect_to root_path
+    end
+
+    it "redirects to login page if user is not logged in" do
+      post :social, params
+
+      expect(response).to redirect_to signin_path
+    end
+  end
+end


spec/features/log_out_inactive_admins.rb
```diff
@@ -90,6 +90,21 @@
...
+      visit admin_feature_url
+
+      expect(page).not_to have_current_path(admin_feature_url)
+    end
+    scenario "can access visitor pages" do
+      visit visitor_page_url
+
+      expect(page).to have_current_path(visitor_page_url)
+    end
+    scenario "can access non-admin (user required login) features when logged in" do
```
kevinburleigh75: Can the wording of this be improved?  It isn't very clear to me...

spec/routines/create_user_spec.rb
```diff
@@ -2,37 +2,71 @@
...
-    it "succeeds when a username is already taken" do
-      FactoryGirl.create(:user, username: "bubba")
-      outcome = nil
-      expect {
-        outcome = CreateUser.call(username: "bubba", ensure_no_errors: true, state: 'activated')
-      }.to change{User.count}.by 1
-      expect(outcome.errors).to be_empty
+  context "when validations fail" do
+    context "when ensure_no_errors is false" do
+      it "returns errors on invalid input" do
```
kevinburleigh75: I think this spec could/should be split into two - one for the exception check, and one for the errors check.

spec/routines/create_user_spec.rb
```diff
@@ -2,37 +2,71 @@
...
 
-    it "succeeds when the sanitized downcased username is already taken" do
-      FactoryGirl.create(:user, username: "Userone")
-      outcome = nil
-      expect {
-        outcome = CreateUser.call(username: "User One", ensure_no_errors: true, state: 'activated')
-      }.to change{User.count}.by 1
-      expect(outcome.errors).to be_empty
+    context "when ensure_no_errors is true" do
+      let(:invalid_input) {
```
kevinburleigh75: This `let` block is used only once, and extracting it doesn't seem to add any clarity - perhaps re-inline it?

spec/features/add_application_spec.rb
```diff
@@ -26,15 +28,10 @@
...
-    expect(page.text).to match(/Application Id: [a-z0-9]+/)
-    expect(page.text).to match(/Secret: [a-z0-9]+/)
-    expect(page).to have_content('Trusted? No')
+    visit '/oauth/applications/new'
+
+    expect(page.status_code).to eq 500
   end
 end
+
+ENV["ACTION_DISPATCH_SHOW_EXCEPTIONS"] = "false"
```
kevinburleigh75: This construct seems _very_ error-prone - is there really no better way to manage this setting?

BryanEliDimas: It's one way, I was hoping that at least it worked, but it didn't. So I have to do something else now anyway.

spec/features/log_out_inactive_admins.rb
BryanEliDimas: Improved. Now it says, simply, "can access user features".

spec/routines/create_user_spec.rb
BryanEliDimas: Now it's split into two.

spec/routines/create_user_spec.rb
BryanEliDimas: I re-inlined it. For some reason I had extracted but then it didn't make sense to have done that and had forgotten to re-inline it.

spec/features/add_application_spec.rb
BryanEliDimas: So Travis failed because it was never really working, only for my specs locally. The alternative forced me to do a bug fix in all the specs (in OpenStax's `rescue_from` gem settings which are stored in an initializer).

spec/support/feature_helpers.rb
```diff
@@ -141,6 +141,15 @@ def with_error_pages
   end
 end
 
+def with_show_exceptions_true
```
Dantemss: Unused code?

spec/features/unknown_route_spec.rb
```diff
@@ -3,15 +3,13 @@
...
+    visit '/lkajsdlkjdklfsjldkfjsl.json'
+    expect(page).to have_http_status 404
   end
 
   scenario 'when it is an HTML request' do
-    expect{
-      visit '/lkajsdlkjdklfsjldkfjsl'
-    }.to raise_error(ActionController::RoutingError)
+    visit '/lkajsdlkjdklfsjldkfjsl'
+    expect(page).to have_http_status 404
```
Dantemss: :not_found (and line 7 too)

spec/features/add_application_spec.rb
```diff
@@ -26,15 +26,8 @@
...
-    create_new_application
-    expect(page).to have_content('Application created.')
-    expect(page).to have_content('Application: example')
-    expect(page).to have_content('Callback urls: https://localhost/')
-    expect(page.text).to match(/Application Id: [a-z0-9]+/)
-    expect(page.text).to match(/Secret: [a-z0-9]+/)
-    expect(page).to have_content('Trusted? No')
+
+    visit '/oauth/applications/new'
+    expect(page).to have_http_status 403
```
Dantemss: :forbidden

spec/controllers/api/v1/contact_infos_controller_spec.rb
```diff
@@ -63,9 +62,8 @@
     end
 
     it "403s if the wrong user makes the request" do
-      expect{
-        api_put :confirm_by_pin, wrong_user_token, parameters: {id: contact_info.id}
-      }.to raise_error(SecurityTransgression)
+      api_put :confirm_by_pin, wrong_user_token, parameters: {id: contact_info.id}
+      expect(response).to have_http_status 403
```
Dantemss: :forbidden (and line 25 too)

spec/support/feature_helpers.rb
BryanEliDimas: Oh, yes. I'm deleting it.

PR #342, opened 2016-08-11
Allow rendering site in requested language

aiwenar:
This PR enables the OpenStax Accounts website to be translated and displayed in user's preferred language.

The preferred languages are determined based on value of the `Accept-Language` header, which is assumed to be correctly formed according to [Section 5.3.5 of RFC 7231][1], using the Basic Filtering matching scheme as defined by [Section 3.3.1 of RFC 4647][2], except ignoring all language variants (so e.g. `en-US` and `en-GB` would both be matched as `en`). Other matching scheme may be implemented if needed. In case of none of specified languages being available or malformed or missing header the default language will be assumed (value of `I18n.default_language`).

[1]: https://tools.ietf.org/html/rfc7231#section-5.3.5
[2]: https://tools.ietf.org/html/rfc4647#section-3.3.1

The content in Ruby scripts and ERB templates was internationalized using [Rails Internationalisation API][3], which addition of a system used to properly format enumerations with respect to language's grammatical rules. For details about this system see `config/initializers/locale.rb`, for implemented enumerators see `config/locales/enumerators`.

A small part of the site content is located within CoffeeScript scripts. Those cannot be directly internationalized using Rails I18n. We have implemented two ways of achieving it: one indirectly via Rails I18n (see branch [`I18n-html-ror`][4]), the other using L20n - a library for internationalizing
HTML/JavaScript (see brach [`I18n-html-l20n`][5]). It is our view, however, that an ideal solution would be to rewrite this part so that no strings are ever present in JavaScript/CoffeeScript.

[3]: http://guides.rubyonrails.org/i18n.html
[4]: https://github.com/katalysteducation/accounts/tree/I18n-html-ror
[5]: https://github.com/katalysteducation/accounts/tree/I18n-html-l20n

In addition to basic work on internationalization we have also modified RSpec tests to check for missing translations and not use hardcoded strings. We have also provided Polish translation.

Several more issues need to be addressed in order to achieve complete internationalization. Those were not included in this PR due to reasons stated and, should this PR be accepted, will be subject to further work and PRs.

1.  User language preference.

    User should be able to choose a language (or a list of languages) in which they prefer to use the website (and other related sites such as CNX Webview) and receive communications.

    This would require altering the database to enable storing this preference and including a selector in the user preferences.

2.  Ability to change language from the website without changing browser language for user not logged in.

    This functionality allows an user who understand one of non-default languages but have it not listed in `Accept-Language` to discover that the site is available in it.

    This would require including a language selector somewhere on the website and a way to respect this selection (for example using cookies or URL query parameters).

3.  Mailers.

    All communications should be sent to the user in their preferred language. This requires that we know what is this language and therefore can only be implemented after 1.

4.  Model validation error messages.

    Due to the way those messages are assembled their internationalization if done with minimal impact on Accounts would force a certain structure on those messages, which might feel unnatural in some languages, and even make it impossible to translate into some.

    For that reason we have not internationalized this part and we're looking for best solution.

All of these (except 4) require making a design decision, which we at Katalyst Education believe is best left to OpenStax.


jpslav:
Thanks @aiwenar, will review.

app/controllers/application_controller.rb
```diff
@@ -8,7 +8,29 @@ class ApplicationController < ActionController::Base
...
+          [lang.to_sym, q]
+        end.keep_if do |x|
+          I18n.available_locales.include? x[0]
+        end.sort {|a, b| b[1] <=> a[1] }
+           .map {|x| x[0] }
+           .first || I18n.default_locale
+      else
+       I18n.default_locale
+      end
+  end
```
jpslav: Can you a spec to test this code for various possible values of the accept language?  And optionally put this code in a `lib` module to limit the code here and maybe make it easier to test.

app/controllers/application_controller.rb
```diff
@@ -8,7 +8,29 @@ class ApplicationController < ActionController::Base
...
+          q = 1 == v.size ? 1 : v[1].to_f
+          [lang.to_sym, q]
+        end.keep_if do |x|
+          I18n.available_locales.include? x[0]
+        end.sort {|a, b| b[1] <=> a[1] }
+           .map {|x| x[0] }
+           .first || I18n.default_locale
+      else
+       I18n.default_locale
+      end
```
Dantemss: I'm generally opposed to `unless... else` because it's confusing. I'd much rather it be `if... else`.

spec/support/i18n.rb
```diff
@@ -0,0 +1,20 @@
+module I18nMacros
+  def t key, *args
+      I18n.t key, *args
+  end
+end
+
+# If we had rails 4+ we could make application raise exception on missing
+# translation. Instead we have to check if page has elements with class
+# translation_missing which rails 3 inserts on missing translation.
+RSpec::Matchers.define :have_no_missing_translations do ||
```
Dantemss: No need for `||` but I guess it doesn't matter...

app/controllers/application_controller.rb
Dantemss: Or since you have `|| I18n.default_locale` at the end, could just do `accept_language = request.env['HTTP_ACCEPT_LANGUAGE'] || ''` in the beginning and get rid of the `unless` completely.

config/initializers/locale.rb
```diff
@@ -0,0 +1,83 @@
...
+  def self.define_enumerator language, &block
+    enumerator = Module.new
+    enumerator.instance_eval &block
+    Enumerators.define language, enumerator
+  end
+end
+
+Dir.glob('config/locales/enumerators/*.rb').each do |file|
+  load file
+end
```
jpslav: Can you write a spec for this code?

config/locales/enumerators/pl.rb
```diff
@@ -0,0 +1,9 @@
+I18n.define_enumerator :pl do
+  def enumerate kind, list, options
+    conjunction = { :all => ' oraz ', :any => ' lub ', :one => ' albo ' }[kind]
+    if 2 >= list.length
+      return list.join conjunction
+    end
+    return list[0...-1].join(', ') + conjunction + list[-1]
+  end
+end
```
jpslav: Can this enumerate method and the english version just delegate to a common enumerate method?  If we end up with 10 languages I wouldn't want 10 copies of this code.  Also, if you can write a spec for the common method, that'd be great.

spec/support/i18n.rb
```diff
@@ -0,0 +1,20 @@
...
+  include RSpec::Matchers::Composable
+
+  match do |actual|
+    actual !~ /class="translation_missing"/
+  end
+
+  failure_message do |actual|
+    "expected that response would have no missing translations but #{/title="translation missing: (.+?)"/.match(actual)[1]} was missing"
+  end
+end
```
jpslav: Thanks for the note about rails 4.  Accounts is our only site still on 3, hopefully we'll change it soon and this note will help us then!

jpslav:
Great PR @aiwenar et al - @Dantemss have reviewed and commented.

jpslav:
@nathanstitt re @aiwenar's bullet (1) in the description, we should consider a `ui-settings` feature on Accounts like you just did in https://github.com/openstax/tutor-server/pull/1230.  All sites could then pull common language preferences from the Accounts API.

@aiwenar, re bullet (2), is there a standard "Change Language" text or icon that we could put in the footer of Accounts for users to change their language preference before being logged in?  If we end up having such an option on all of our sites, I'd like it to be something subtle enough to not stick out but easy enough to find for non English speaking users.

re bullet (3), yep that is a good point.  If you guys wanted to take a stab at an Accounts version of @nathan's PR above, you'd be welcome to.

(4) does sound tricky.  I don't have any great ideas off the top of my head.

reedstrm:
Note that for bullet 2 - 'allow language selection for not-logged-in-user' , in general that's going to be a function of the frontend of the site that's using accounts for validation, rather than accounts itself. Because, by definition, that user is not authenticated, so doesn't have an accounts profile.

jpslav:
@reedstrm - re bullet 2, I think non-logged-in users will still need Accounts to be language-aware during the process of signing up for an account (through the Accounts UI)?

aiwenar:
@jpslav, @Dantemss: I have refactored locale selection into separate module under lib/ and removed `unless ... else` in favour of `request.env['HTTP_ACCEPT_LANGUAGE'] || ''`. After some benchmarking I have also replaced the `Accept-Language` parser with a regexp-based one.

@jpslav, @reedstrm: re bullet 2 - Yes, I meant that users who have not yet signed in/up should still be able to change site language without reconfiguring their browser/system.

> re bullet (2), is there a standard "Change Language" text or icon that we could put in the footer of Accounts for users to change their language preference before being logged in?

I believe it's a question for your design team.

config/locales/enumerators/pl.rb
aiwenar: I was thinking about organising it by language groups which use the same syntax, but was unable to find any information about it. Since I couldn't think of any other logical organisation I decided to just left it for now as separate implementations for each language.

If you have any idea where I could look for comparison of conjunctions in language groups, or can think of other ways to organise enumerator implementations, I would gladly look into it more.

jpslav: ok

jpslav:
Thanks @aiwenar - those changes look good.  All that's left is addressing https://github.com/openstax/accounts/pull/342/files#r74490405 and then rebasing (please rebase instead of merging in master if you don't mind -- that's our standard practice).

jpslav:
Hi @aiwenar -- can you rebase so we can merge this in?  We can skip the other spec until later.  Thanks!

aiwenar:
For some reason (I suspect failed merge replay) rebasing caused some of tests to fail while merging did not. Since I was unable to find what exactly was the cause for those failures I sent you the working merge. If that's not OK I'll try to find what exactly is wrong with the rebase.

PR #343, opened 2016-08-23
Some admin account management improvements

Dantemss:
- Admins can now see the password they are typing when editing a user
- Removed password confirmation field from admin edit screen
- Signing in as a user now counts as a successful sign in for the purposes of changing authentications

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -36,8 +36,10 @@ module Admin
     end
 
     def become
+      admin = current_user
       security_log :admin_became_user, user_id: params[:id], username: @user.username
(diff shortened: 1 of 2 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -36,8 +36,10 @@ def destroy
     end
 
     def become
+      admin = current_user
       security_log :admin_became_user, user_id: params[:id], username: @user.username
       sign_in!(@user)
+      security_log :sign_in_successful, admin_user_id: admin.id, admin_username: admin.username
```
jpslav: Why are we saying the admin logged in?  Is this so that something else works?

PR #344, opened 2016-09-06
Disable Click to verify button on click

BryanEliDimas:
This PR helps prevent bots from clicking the `Click to verify` button (or in other words, following the URL) which sends emails to verify the logged in user's email address. The problem it presents otherwise is that the database will give the following error:
```
An ActiveRecord::StatementInvalid occurred in contact_infos#resend_confirmation:

  PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block
: SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL READ COMMITTED
  activerecord (3.2.22) lib/active_record/connection_adapters/postgresql_adapter.rb:650:in async_exec
```

### DIFF
--- a/app/assets/javascripts/profile/email.coffee
+++ b/app/assets/javascripts/profile/email.coffee
@@ -24,11 +24,15 @@ class Email
 
   sendVerification: (ev) ->
     ev.preventDefault()
+    ev.target.disabled = true
     $.ajax({type: "PUT", url: @url('resend_confirmation')})
       .success( (resp) =>
         OX.Alert.display(message: resp.message, type: 'success', parentEl: @$el)
       )
-      .error(OX.Alert.displayInsideElement(@$el))
+      .error( (e) =>
+        OX.Alert.display(_.extend(e, parentEl: @$el))
+        ev.target.disabled = false
+      )
 
   saveSearchable: (ev) ->
     @toggleSpinner(true)
--- a/app/assets/stylesheets/profile.css.scss
+++ b/app/assets/stylesheets/profile.css.scss
@@ -92,7 +92,18 @@
         margin-left: 6px;
       }
     }
-    .verify { margin-left: 20px; }
+    .verify {
+      margin-left: 20px;
+      form {
+        display: inline;
+        div { display: inline; }
+      }
+      input {
+        color: $link-color;
+        padding: 0;
+        font-size: 13px;
+      }
+    }
     &.new {
       .verify, .properties, .mod-holder { display: none; }
     }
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/profile/email.coffee
```diff
@@ -26,6 +26,7 @@ class Email
     ev.preventDefault()
     $.ajax({type: "PUT", url: @url('resend_confirmation')})
       .success( (resp) =>
+        ev.target.disabled = true
         OX.Alert.display(message: resp.message, type: 'success', parentEl: @$el)
```
nathanstitt: I think it's a good idea to disable the button, but it might be better to:

* Disable button as soon as click occurs, before the ajax call.
* Re-enable the button only if the call fails, in the `.error` callback

This would prevent double clicks on the button, which I suspect is the reason for the DB error.  Robots shouldn't have access to the page, since it requires a login.

BryanEliDimas: Thanks for your comment, Nathan.
Reason why Dante and I suspect that it's bots following the URL is because we were not able to reproduce the error by double clicking ourselves manually or with RSpec (`double_click`).  To abuse that button, we suspect it requires an async request; we tried creating multiple `Threads` but that didn't work with RSpec. So we agreed that I should use `button_to` instead of `link_to` since it "is the safest method to ensure links that cause changes to your data are not triggered by search bots or accelerators"—and also disable it on click.

I'll make the fixes that you recommended.

BryanEliDimas:
@nathanstitt: I just pushed the fixes, would please review? Thanks!

app/assets/stylesheets/profile.css.scss
```diff
@@ -96,6 +96,13 @@
     &.new {
       .verify, .properties, .mod-holder { display: none; }
     }
+    .verify form { display: inline; }
+    .verify form div { display: inline; }
+    .verify input {
+      color: #428bca;
+      padding: 0;
+      font-size: 13px;
+    }
```
nathanstitt: **nitpick:** Since this is scss, it could be written using nesting as:
```sass
.verify {
  form { 
    display: inline; 
    div { display: inline; }
  }
  input {
    color: #428bca;
    padding: 0;
    font-size: 13px;
  }
}
```

nathanstitt:
Looks good @BryanEliDimas - my jQuery's a bit rusty, but I think adding the `.disabled` should prevent the double-click, but sounds like that's possibly not the issue anyway.

app/assets/stylesheets/profile.css.scss
BryanEliDimas: Woops! You're right. Fixed.

BryanEliDimas:
@nathanstitt I made the fixes. Thanks!

jpslav:
Looks good @BryanEliDimas -- are you ready for merging?

BryanEliDimas:
@jpslav Yeap! Ready to merge. Thanks.

PR #345, opened 2016-09-07
Temporary fix for issue 308

BryanEliDimas:
Referenced in [Trello](https://trello.com/c/pxA70dpY/79-accounts-explodes-if-you-try-to-add-the-same-oauth-provider-twice)

_Accounts seems to explode with a 500 when the same OAuth provider is added to a user's account twice. Error comes from uniqueness constraint in the database_

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -84,6 +84,11 @@ class SessionsController < ApplicationController
           redirect_to profile_path, alert: "That sign in option is already used by someone " \
                                            "else. If that someone is you, remove it from " \
                                            "your other account and try again."
+        when :same_provider
+          security_log :authentication_transfer_failed, authentication_id: authentication.id
+          redirect_to profile_path, alert: "You are logged in as #{current_user.name}. A different #{authentication.provider} account " \
+                                            "is already linked to your OpenStax account. Only one #{authentication.provider} account " \
+                                            "can be linked to your OpenStax account at a time."
         else
           Rails.logger.fatal "IllegalState: OAuth data: #{request.env['omniauth.auth']}"
           raise IllegalState, "SessionsCreate errors: #{@handler_result.errors.inspect
--- a/app/handlers/sessions_create.rb
+++ b/app/handlers/sessions_create.rb
@@ -46,6 +46,7 @@ class SessionsCreate
     authentication =
       Authentication.find_or_create_by_provider_and_uid(@data.provider, @data.uid.to_s)
     authentication_user = authentication.user
+    outputs[:authentication] = authentication
 
     if signed_in? && authentication_user == current_user
       status = :no_action
@@ -57,6 +58,8 @@ class SessionsCreate
         status = :authentication_taken
       else
         return outputs[:status] = :new_signin_required if user_signin_is_too_old?
+        return outputs[:status] = :same_provider \
+                                  if current_user.authentications.any?{|user_auth| user_auth.provider == authentication.provider}
         run(TransferAuthentications, authentication, current_user)
         run(TransferOmniauthData, @data, current_user) if authentication.provider != 'identity'
         status = :authentication_added
@@ -88,7 +91,6 @@ class SessionsCreate
     end
 
     outputs[:status] = status
-    outputs[:authentication] = authentication
   end
 
   protected
--- a/app/routines/transfer_authentications.rb
+++ b/app/routines/transfer_authentications.rb
@@ -10,7 +10,9 @@ class TransferAuthentications
     authentications = [authentications] if !(authentications.is_a? Array)
     authentications.each do |authentication|
       authentication_user = authentication.user
-      authentication.update_attribute(:user_id, target_user.id)
+      authentication.update_attributes(user_id: target_user.id)
+      transfer_errors_from(authentication, {type: :verbatim}, true)
+
       run(DestroyUser, authentication_user) \
         if authentication_user && !authentication_user.is_activated? && \
            authentication_user.reload.authentications.empty?
--- a/spec/handlers/sessions_create_spec.rb
+++ b/spec/handlers/sessions_create_spec.rb
@@ -356,6 +356,59 @@ describe SessionsCreate, type: :handler do
             end
           end
         end
+
+        context "linked to no one" do
+          context "but user already has already authenticated with the same provider before" do
+            context "new_social user" do
+              let!(:signed_in_user) { FactoryGirl.create(:new_social_user) }
+              let!(:authentication) { FactoryGirl.create(:authentication, user: signed_in_user) }
+
+              it "returns a status output instead of an exception or errors" do
+                handle_with_uid(111)
+
+                same_provider_different_uid = nil
+                expect{same_provider_different_uid = handle_with_uid(222)}.not_to raise_exception
+                expect(same_provider_different_uid.outputs[:status]).to eq :same_provider
+                expect(same_provider_different_uid.errors).to be_empty
+              end
+            end
+
+            context "temp user" do
+              let!(:signed_in_user) { FactoryGirl.create(:temp_user) }
+              let!(:authentication) { FactoryGirl.create(:authentication, user: signed_in_user) }
+
+              it "returns a status output instead of an exception or errors" do
+                handle_with_uid(111)
+
+                same_provider_different_uid = nil
+                expect{same_provider_different_uid = handle_with_uid(222)}.not_to raise_exception
+                expect(same_provider_different_uid.outputs[:status]).to eq :same_provider
+                expect(same_provider_different_uid.errors).to be_empty
+              end
+            end
+
+            context "non-temp user" do
+              let!(:signed_in_user) { FactoryGirl.create(:user) }
+              let!(:authentication) { FactoryGirl.create(:authentication, user: signed_in_user) }
+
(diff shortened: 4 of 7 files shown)


BryanEliDimas:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pxA70dpY&#x2F;79-accounts-explodes-if-you-try-to-add-the-same-oauth-provider-twice"> Accounts explodes if you try to add the same OAuth provider twice</a>

BryanEliDimas:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pxA70dpY&#x2F;79-accounts-explodes-if-you-try-to-add-the-same-oauth-provider-twice"> Accounts explodes if you try to add the same OAuth provider twice</a>

app/handlers/sessions_create.rb
```diff
@@ -57,7 +57,14 @@ def handle
         status = :authentication_taken
       else
         return outputs[:status] = :new_signin_required if user_signin_is_too_old?
-        run(TransferAuthentications, authentication, current_user)
+        auth_transfer = run(TransferAuthentications, authentication, current_user)
+
+        if auth_transfer.outputs[:same_provider]
```
jpslav: To me, setting `same_provider` inside `TransferAuthentications` whenever the `update_attributes` call fails is a little too indirect.  What if that `update_attributes` call failed for another validation error?  

I also got down this line of thinking because I think `TransferAuthentications` needs to `fatal_error` if it can't transfer authentications as directed.  You're right that for now there is only one caller, but we can't assume that will always be true, and using a `fatal_error` are the only way to protect a future caller from what is actually an error.

I'd rather you do a pre-check that is more explicit..

```
return outputs[:status] = :same_provider if current_user.authentications.any?{|user_auth| user_auth.provider == authentication.provider}
```

and then in `TransferAuthentications` bubble up any model errors in the normal way...

```
authentication.update_attribute(:user_id, target_user.id)
transfer_errors_from(authentication, {type: :verbatim}, true) # true for fatal
```

Leave the DB uniqueness constraint in of course.

app/handlers/sessions_create.rb
```diff
@@ -57,7 +57,14 @@ def handle
...
+          outputs[:status] = :same_provider
+          outputs[:authentication] = authentication
```
jpslav: If need it, why not just set it up at the top so all callers can access it?

```
outputs[:authentication] = authentication = 
  Authentication.find_or_create_...
```

This way you can get it back even if you follow my suggestion above to `return outputs[:status] = ...`

spec/handlers/sessions_create_spec.rb
```diff
@@ -356,6 +356,70 @@
...
         end
+
+        context "linked to no one" do
+          context "but user already has already authenticated with the same provider before" do
+            context "new_social user" do
+              let!(:user) { FactoryGirl.create(:new_social_user) }
+              let!(:authentication) { FactoryGirl.create(:authentication, user: user) }
+
+              it "returns a status output instead of raising an exception" do
+                result = lambda { |uid|
```
jpslav: Minor, but... The above `result`s are true handler results.  I think this needs a different name, because it isn't a "result" but a callable thing that produces a result.  `handler_wrapper` or something

jpslav: actually seeing that you use this a couple of times, just put it in a method with a descriptive name... 

```
def handle_for_uid(uid)
  described_class.handle(
    user_state: user_state,
    request: MockOmniauthRequest.new(authentication.provider, uid, {})
  )
end
```

and then

```
handle_for_uid(111)
blah = handle_for_uid(222)
```

This test code shows you can define methods inside or outside of contexts and they will have access to `let` variables.

```
require 'rspec'

describe 'something' do

  context 'blah' do
    let(:foo) { "bar" }

    it 'works' do
      expect(foo).to eq "bar"
      expect(method_inside_context).to eq "bar"
      expect(method_outside_context).to eq "bar"
    end

    def method_inside_context
      foo
    end
  end

  def method_outside_context
    foo
  end

end
```

spec/handlers/sessions_create_spec.rb
```diff
@@ -356,6 +356,70 @@
...
+
+                result.call(111)
+
+                same_provider_different_uid = result.call(222)
+                expect{same_provider_different_uid}.not_to raise_exception
+                expect(same_provider_different_uid.outputs[:status]).to eq :same_provider
+              end
+            end
+          end
+        end
```
jpslav: may not need it, but in case you are testing for lev errors, note that there is a `expect(blah).to have_routine_error(:some_error_code)`.  See `rails_helper.rb`

spec/handlers/sessions_create_spec.rb
```diff
@@ -356,6 +356,70 @@
...
+                            described_class.handle(
+                              user_state: user_state,
+                              request: MockOmniauthRequest.new(authentication.provider, uid, {})
+                            )
+                          }
+
+                result.call(111)
+
+                same_provider_different_uid = result.call(222)
+                expect{same_provider_different_uid}.not_to raise_exception
```
jpslav: Did you ever get this line to fail?  I don't think it would, since `same_provider_different_uid` is not itself something that is being called (it is just an object).  Would have expected to see something like...

```
same_provider_different_uid = nil
expect{same_provider_different_uid = result.call(222)}.not_to raise_exception
```

By the way you can also put these three repeated lines in their own method call that just takes the `222` parameter.

jpslav:
Review complete @BryanEliDimas

app/handlers/sessions_create.rb
BryanEliDimas: 1. Very true! Will fix with 3.

2. The problem is that if I use a `fatal_error`, then the bug "Accounts explodes..." will remain in existence. So we really don't want to error out just because a user is trying to use a different account with the same provider they've already authenticated with. That's why I must output a status. Plus, if I do return a status, then we'll be able to handle it gracefully in `sessions#create`—with a nicer [alert](https://github.com/openstax/accounts/pull/345/files#diff-d5241d488259f32ecbe2f636133e5ddaR89)—which is all we want for now.

3., 4., 5. Will do! Ok.

spec/handlers/sessions_create_spec.rb
BryanEliDimas: Great! How did I not think of that before? :/

BryanEliDimas:
@jpslav I addressed your comments. Please review. Thank you!

app/handlers/sessions_create.rb
jpslav: I would like to discuss whether (2) can still be switched to a `fatal_error`.  I think it can because with (3) you'll be adding a check before you call `TransferAuthentications`, so the `fatal_error` will never be reached in your situation.

app/handlers/sessions_create.rb
```diff
@@ -57,7 +58,8 @@ def handle
         status = :authentication_taken
       else
         return outputs[:status] = :new_signin_required if user_signin_is_too_old?
-        run(TransferAuthentications, authentication, current_user)
+        auth_transfer = run(TransferAuthentications, authentication, current_user)
+        return outputs[:status] = :same_provider if auth_transfer.outputs[:status] == :same_provider
```
jpslav: What I was getting at previously was to change these lines 61 and 62 to

```ruby
return outputs[:status] = :same_provider if current_user.authentications.any?{|user_auth| user_auth.provider == authentication.provider}
run(TransferAuthentications, authentication, current_user) # not going to be reached in your bug case
```

app/routines/transfer_authentications.rb
```diff
@@ -10,7 +10,12 @@ def exec(authentications, target_user)
     authentications = [authentications] if !(authentications.is_a? Array)
     authentications.each do |authentication|
       authentication_user = authentication.user
-      authentication.update_attribute(:user_id, target_user.id)
+      return outputs[:status] = :same_provider \
+            if Authentication.exists?(user_id: target_user.id, provider: authentication.provider)
```
jpslav: you probably added this because of my point (3), but I intended this pre-check to be right before `TransferAuthentications` was even called.

app/controllers/sessions_controller.rb
```diff
@@ -84,6 +84,11 @@ def create
           redirect_to profile_path, alert: "That sign in option is already used by someone " \
                                            "else. If that someone is you, remove it from " \
                                            "your other account and try again."
+        when :same_provider
+          security_log :authentication_transfer_failed, authentication_id: authentication.id
+          redirect_to profile_path, alert: "A different #{authentication.provider} account " \
+                                           "is already linked to your account. Please login with that " \
+                                           "#{authentication.provider} account."
```
Dantemss: OK I thought about this some more and I think this message would be clearer:

```
"You are already logged in. A different #{authentication.provider} account " \
"is already linked to your OpenStax account. Only one #{authentication.provider} account " \
"can be linked to your OpenStax account at a time."
```

Dantemss: @jpslav thoughts?

BryanEliDimas:
@jpslav I made the changes you suggested https://github.com/openstax/accounts/pull/345#r78687126. Also your suggested changes @Dantemss https://github.com/openstax/accounts/pull/345#r78819259.
This PR is ready to be merged.

PR #347, opened 2016-09-13
Create rake task to find users with multiple accounts

BryanEliDimas:
Pushing for review only.

### DIFF
--- a/spec/factories/group.rb
+++ b/spec/factories/group.rb
@@ -5,8 +5,8 @@ FactoryGirl.define do
     is_public false
 
     transient do
-      members_count 1
-      owners_count 1
+      members_count 0
+      owners_count 0
     end
 
     after(:build) do |group, evaluator|
--- a/lib/tasks/find_duplicate_accounts.rake
+++ b/lib/tasks/find_duplicate_accounts.rake
@@ -1,0 +1,89 @@
+desc "Find possible duplicate acccounts"
+task :find_duplicate_accounts => [:environment] do
+  CSV.open("duplicate_users_by_name.csv", 'w+') do |csv|
+    csv << ['User First Name',
+            'User Last Name',
+            'Email Address(es)',
+            'User ID',
+            'Applications',
+            'Social Signup Successful?',
+            'Reset Password Help Requested?',
+            'Help Request Failed?',
+            'Authentication Transfer Failed?'
+          ]
+
+    where_names_match = User.joins{ User.unscoped.as(same)
+                    .on{
+                      (same.id != ~id) & (lower(same.first_name) == lower(~first_name)) & (lower(same.last_name) == lower(~last_name))
+                    }
+                  }.uniq.order(:first_name, :last_name).includes(:security_logs)
+
+    where_names_match.find_each do |user|
+      csv << [user.first_name,
+              user.last_name,
+              (user.contact_infos.any? ? user.contact_infos.map{ |ci| "#{ci.value} #{ci.verified ? '(verified)' : '(NOT verified)'}" }.join(", ") : ""),
+              user.id,
+              (user.applications.any? ? ( user.applications.map(&:name).join(", ") ) : ""),
+              sign_up_successful(user),
+              help_requested(user),
+              help_request_failed(user),
+              authentication_transfer_failed(user)
+             ]
+    end
+  end
+
+  CSV.open("duplicate_users_by_email.csv", 'w+') do |csv|
+    csv << ['Email Address',
+            'ContactInfo ID',
+            'User First Name',
+            'User Last Name',
+            'User ID',
+            'Applications',
+            'Social Signup Successful?',
+            'Reset Password Help Requested?',
+            'Help Request Failed?',
+            'Authentication Transfer Failed?'
+           ]
+
+    where_email_addresses_match = ContactInfo.joins{ ContactInfo.unscoped.as(same)
+                                               .on{ (same.id != ~id) & (lower(same.value) == lower(value)) }
+                                    }.joins{ user.outer }.preload(:user).uniq
+
+    where_email_addresses_match.find_each do |contact_info|
+      csv << [
+              "#{contact_info.value} #{contact_info.verified ? '(verified)' : '(NOT verified)'}",
+              contact_info.id,
+              contact_info.user.try(:first_name),
+              contact_info.user.try(:last_name),
+              contact_info.user.try(:id),
+              (contact_info.user && contact_info.user.applications.any?) ? contact_info.user.applications.map(&:name).join(", ") : "",
+              sign_up_successful(contact_info.user),
+              help_requested(contact_info.user),
+              help_request_failed(contact_info.user),
+              authentication_transfer_failed(contact_info.user)
+             ]
+    end
+  end
+end
+
+private
+
+def sign_up_successful(user)
+  return if user.nil?
+  user.security_logs.where(event_type: SecurityLog.event_types[:sign_up_successful]).map{|sec| "On #{sec.created_at}"}.join(" and ")
+end
+
+def help_requested(user)
+  return if user.nil?
+  user.security_logs.where(event_type: SecurityLog.event_types[:help_requested]).map{|sec| "On #{sec.created_at}"}.join(" and ")
+end
+
+def help_request_failed(user)
+  return if user.nil?
+  user.security_logs.where(event_type: SecurityLog.event_types[:help_request_failed]).map{|sec| "On #{sec.created_at}"}.join(" and ")
+end
+
+def authentication_transfer_failed(user)
+  return if user.nil?
+  user.security_logs.where(event_type: SecurityLog.event_types[:authentication_transfer_failed]).map{|sec| "On #{sec.created_at}"}.join(" and ")
+end
--- a/spec/lib/tasks/find_duplicate_accounts_spec.rb
+++ b/spec/lib/tasks/find_duplicate_accounts_spec.rb
@@ -1,0 +1,144 @@
+require 'rails_helper'
+
+RSpec.describe "find_duplicate_accounts" do
+  include_context "rake"
+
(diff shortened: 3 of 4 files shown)


lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,64 @@
+namespace :accounts do
+  namespace :duplicate_accounts do
+    desc "Find possible duplicate acccounts"
+    task :find => [:environment] do
+      CSV.open("duplicate_users_by_name.csv", 'w+') do |csv|
+        csv << ['User First Name', 'User Last Name', 'Email Address(es)', 'User ID', 'Social Signup Successful', 'Reset Password Help Requested', 'Help Request Failed']
```
kevinburleigh75: For boolean-like fields, perhaps adding a `?` will make the title more clear (e.g., "Social Signup Successful?")

lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,64 @@
...
+
+        where_names_match = User.joins{ User.unscoped.as(same)
+                        .on{
+                          (same.id != ~id) & ((lower(same.first_name) == lower(~first_name)) & (lower(same.last_name) == lower(~last_name)))
```
kevinburleigh75: There seems to be an extra set of parens around the last two checks.

lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,64 @@
...
+                        }
+                      }.uniq.order(id: :asc)
+
+        where_names_match.find_each do |user|
+          csv << ["#{user.first_name}",
+                  "#{user.last_name}",
+                  "#{user.contact_infos.map(&:value).join(", ") if user.contact_infos.any?}",
```
kevinburleigh75: If there are multiple `contact_infos`, will the `,` joining them mess up the csv file?

lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,64 @@
...
+                   user.id,
+                   sign_up_successful(user),
+                   help_requested(user),
+                   help_request_failed(user)
+                 ]
+        end
+      end
+
+      CSV.open("duplicate_users_by_email.csv", 'w+') do |csv|
```
kevinburleigh75: Perhaps this should be placed in a temp file (and the filename output somewhere) to avoid cluttering up the live code directories?

lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,64 @@
...
+
+        where_email_addresses_match = ContactInfo.joins{ ContactInfo.unscoped.as(same)
+                                                   .on{ (same.id != ~id) & (lower(same.value) == lower(value)) }
+                                        }.joins{ user.outer }.preload(:user).uniq
+
+        where_email_addresses_match.find_each do |contact_info|
+          csv << [contact_info.value,
+                  contact_info.id,
+                  "#{contact_info.user.try(:first_name)}",
+                  "#{contact_info.user.try(:last_name)}",
```
kevinburleigh75: It seems like there is no need for the extra quoting work; if the strings exist, they'll be strings, if not, `nil` will convert to an empty string.

lib/tasks/find_possible_duplicate_accounts.rake
BryanEliDimas: Nope. Columns with commas in CSV files are just put into quotes.
`Robert,Martin,"c2428a@48b89c.e9f831, f28661@a4a961.b062a5",3,true on: 2016-09-13 17:34:43 UTC,true on: 2016-09-13 17:34:43 UTC. true on: 2016-09-13 17:34:43 UTC,true on: 2016-09-13 17:34:43 UTC`

lib/tasks/find_possible_duplicate_accounts.rake
BryanEliDimas: I think @Dantemss would disagree? https://github.com/openstax/tutor-server/pull/1273/files#diff-f549f4d06b404586b923fb093317dee4R10

lib/tasks/find_possible_duplicate_accounts.rake
BryanEliDimas: True.

BryanEliDimas:
I made @kevinburleigh75's suggested changes and added a couple more data points (`user.applications`, and is `contact_info.verified` or not). 

It should be ready to roll now. @jpslav

lib/tasks/find_possible_duplicate_accounts.rake
Dantemss: Kevin is probably right though.

lib/tasks/find_possible_duplicate_accounts.rake
```diff
@@ -0,0 +1,93 @@
...
+                'Reset Password Help Requested?',
+                'Help Request Failed?',
+                'Authentication Transfer Failed?'
+              ]
+
+        where_names_match = User.joins{ User.unscoped.as(same)
+                        .on{
+                          (same.id != ~id) & (lower(same.first_name) == lower(~first_name)) & (lower(same.last_name) == lower(~last_name))
+                        }
+                      }.uniq.order(id: :asc)
```
jpslav: want to order based on first and last name so potential dupes are near each other?

jpslav: also maybe not a big deal if there aren't many users, but do you want to `.includes(:security_logs)` so you don't get N+1 queries below?

jpslav: either way is ok for me.

spec/factories/group.rb
```diff
@@ -5,8 +5,8 @@
     is_public false
 
     transient do
-      members_count 1
-      owners_count 1
+      members_count 0
+      owners_count 0
```
jpslav: why the change?

spec/lib/tasks/find_possible_duplicate_accounts_spec.rb
```diff
@@ -0,0 +1,159 @@
+require 'rails_helper'
+require 'rake'
+
+RSpec.describe "duplicate_accounts:find", type: :rake do
```
jpslav: See https://github.com/openstax/tutor-server/pull/1229/files for a spec helper I wrote for rake files -- you could pull it in here and reduce some of the code (and probably the top two `before` calls)

spec/lib/tasks/find_possible_duplicate_accounts_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+      expect(User.count).to eq 5
+      expect(result.count).to eq 2
+      expect(SecurityLog.count).to eq 6
+
+      expect(result[0]["User First Name"]).to eq user_1.first_name
+      expect(result[0]["User Last Name"]).to eq user_1.last_name
+      expect(result[0]["Email Address(es)"]).to eq user_1.contact_infos.map{ |ci| "#{ci.value} #{ci.verified ? '(verified)' : '(NOT verified)'}" }.join(", ")
+      expect(result[0]["User ID"]).to eq user_1.id.to_s
+      expect(result[0]["Social Signup Successful?"]).to eq user_1.security_logs.where(event_type: sign_up_successful_enum).map{|sec| "On #{sec.created_at}"}.join(" and ")
```
jpslav: Can replace 

`user_1.security_logs.where(event_type: sign_up_successful_enum)`

with

`SecurityLog.sign_up_successful.where(user: user_1)`

and ditch the `let(:sign_up_successful_enum) ...` line above.  Ditto other related lines.

jpslav: Also, I'd be interested in hearing what @Dantemss and @kevinburleigh75 think, but by putting the same code in this line that you have in the rake file, you're not really testing that you're getting the output that you want, only that the same code in two places returns the same results.  Ditto for the email address lines.  

E.g. you could have stored output of line 39 in `@sus` and then here check that the value equals `"On #{@sus.created_at}`.

jpslav: This is a bit unrelated, but why does `sign_up_successful` imply anything about **social** signup?

jpslav:
Looking good @BryanEliDimas -- I did my review

spec/factories/group.rb
BryanEliDimas: Because otherwise it has unintended consequences for the layman: it creates new users automatically by default so if the `User.count` matters to your spec, it will fail.

Eg.: In my spec, I'd have to do
```
let!(:application) { FactoryGirl.create :doorkeeper_application, owner: user_1 }
FactoryGirl.create :application_user, user: user_1, application: application
FactoryGirl.create :application_user, user: user_1, application: application
```
to get the `User.count` I expect, just because some default in the `group` factory—which `doorkeeper_application` needs—which `application_user` needs.

spec/lib/tasks/find_possible_duplicate_accounts_spec.rb
BryanEliDimas: re: This is a bit unrelated, but why does sign_up_successful imply anything about social signup?
Oh, you're right. It does not. For some reason I thought `sign_up_successful` security log was only used in `signup#social` controller but I was mistaken.

lib/tasks/find_possible_duplicate_accounts.rake
BryanEliDimas: That works great. Both comments.

lib/tasks/find_possible_duplicate_accounts.rake
BryanEliDimas: Ah, dang it! 😫

PR #349, opened 2016-09-19
Pull faculty verification from Salesforce

jpslav:
Links Accounts users to SF contacts.  Caches their faculty verification information and provides it to product sites via the user JSON.

Added a full page admin console...

image

Salesforce management screen:

image

Left the popup admin console in for now with a link to the full page one...

image

User detail page shows faculty status info, SF contact ID (with link)

## TODO

* [x] Include faculty verification status in oauth info returned to product sites
  * this will be handled in accounts-rails
* [x] Include faculty verification status in user GET API
* [ ] Update deployment repo to have new secrets (wait for Rails 4 merge; salesforce keys and salesforce mail recipient)
* [x] Handle `Email_alt`
* [x] Update AnonymousUser to match interface of User
* [x] Code to read recently updated SF Contacts in batches and copy SF ID and faculty verification status to Accounts User (when verified email matches) (store last modified date read for last Contact pulled from)
* [x] Cron scheduling of that code
* [x] Make sure admin views return to the admin dialog after login / destroy SF user
* [x] Provide a manual trigger for an admin to re-read all fac verification info
* [x] Spec to make sure only /api/user returns SF info.
* [X] SPECS!

### DIFF
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -16,5 +16,13 @@ module Admin
       redirect_to admin_salesforce_path
     end
 
+    def update_users
+      UpdateUserSalesforceInfo.call
+
+      flash[:notice] = "The update completed."
+
+      redirect_to admin_salesforce_path
+    end
+
   end
 end
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -6,8 +6,7 @@ module Admin
     end
 
     def callback
-      user = SalesforceUser.save_from_omniauth!(env["omniauth.auth"])
-      SalesforceUser.all.reject{|uu| uu.id == user.id}.each(&:destroy)
+      SalesforceUser.save_from_omniauth!(env["omniauth.auth"])
       redirect_to admin_salesforce_path
     end
 
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -13,14 +13,13 @@ module Admin
 
     def destroy_user
       SalesforceUser.destroy_all
+      ActiveForce.clear_sfdc_client! # since user is now gone, any client invalid
       redirect_to admin_salesforce_path
     end
 
     def update_users
       UpdateUserSalesforceInfo.call
-
       flash[:notice] = "The update completed."
-
       redirect_to admin_salesforce_path
     end
 
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -1,0 +1,21 @@
+module Admin
+  class SalesforceController < BaseController
+    def callback
+      user = SalesforceUser.save_from_omniauth!(env["omniauth.auth"])
+      SalesforceUser.all.reject{|uu| uu.id == user.id}.each(&:destroy)
+      redirect_to admin_salesforce_path
+    end
+
+    def destroy_user
+      SalesforceUser.destroy_all
+      redirect_to admin_salesforce_path
+    end
+
+    protected
+
+    def salesforce_path
+      admin_salesforce_path
+    end
+
+  end
+end
--- a/app/controllers/admin/salesforce_controller.rb
+++ b/app/controllers/admin/salesforce_controller.rb
@@ -1,5 +1,10 @@
 module Admin
   class SalesforceController < BaseController
+    layout 'admin'
+
+    def show
+    end
+
     def callback
       user = SalesforceUser.save_from_omniauth!(env["omniauth.auth"])
       SalesforceUser.all.reject{|uu| uu.id == user.id}.each(&:destroy)
@@ -9,12 +14,6 @@ module Admin
     def destroy_user
       SalesforceUser.destroy_all
       redirect_to admin_salesforce_path
-    end
-
-    protected
-
-    def salesforce_path
-      admin_salesforce_path
     end
 
   end
--- a/app/mailers/dev_mailer.rb
+++ b/app/mailers/dev_mailer.rb
@@ -1,6 +1,6 @@
 class DevMailer < SiteMailer
 
-  def pp(object:, to:, subject:)
+  def inspect_object(object:, to:, subject:)
     @object = object
     mail to: to,
          subject: subject
--- a/app/mailers/dev_mailer.rb
+++ b/app/mailers/dev_mailer.rb
@@ -1,0 +1,9 @@
+class DevMailer < SiteMailer
+
+  def pp(object:, to:, subject:)
+    @object = object
+    mail to: to,
+         subject: subject
+  end
+
+end
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -1,5 +1,7 @@
 require 'singleton'
 
+class AnonymousUserIsImmutableError < StandardError; end
+
 class AnonymousUser
   include Singleton
 
@@ -86,6 +88,7 @@ class AnonymousUser
     end
 
     define_method "#{status}!" do
+      raise AnonymousUserIsImmutableError, "Cannot set faculty status on the AnonymousUser."
     end
   end
 
@@ -94,6 +97,7 @@ class AnonymousUser
   end
 
   def faculty_status=(status)
+    raise AnonymousUserIsImmutableError, "Cannot set faculty status on the AnonymousUser."
   end
 
 end
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -80,4 +80,20 @@ class AnonymousUser
     full_name
   end
 
+  User.faculty_statuses.each do |status, value|
+    define_method "#{status}?" do
+      User::DEFAULT_FACULTY_STATUS.to_s == status
+    end
+
+    define_method "#{status}!" do
+    end
+  end
+
+  def faculty_status
+    User::DEFAULT_FACULTY_STATUS.to_s
+  end
+
+  def faculty_status=(status)
+  end
+
 end
--- a/app/representers/api/v1/user_representer.rb
+++ b/app/representers/api/v1/user_representer.rb
@@ -40,10 +40,18 @@ module Api::V1
              readable: true,
              writeable: false
 
-    property :faculty_status,
+    property :salesforce_contact_id,
              type: String,
              readable: true,
              writeable: false
+
+    property :faculty_status,
+             type: String,
+             readable: true,
+             writeable: false,
+             schema_info: {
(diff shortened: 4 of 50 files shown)


jpslav:
@Dantemss please review when you have a chance - will still rebase on top of your Rails 4 work.

app/controllers/admin/salesforce_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Admin
+  class SalesforceController < BaseController
+    layout 'admin'
+
+    def show
+    end
+
+    def callback
+      user = SalesforceUser.save_from_omniauth!(env["omniauth.auth"])
+      SalesforceUser.all.reject{|uu| uu.id == user.id}.each(&:destroy)
```
Dantemss: Whoa... This makes me wonder if SalesforceUser should even be an ActiveRecord thing (couldn't it just be something handled by rails-settings-cached?)

app/mailers/dev_mailer.rb
```diff
@@ -0,0 +1,9 @@
+class DevMailer < SiteMailer
+
+  def pp(object:, to:, subject:)
```
Dantemss: `pp`?

app/routines/update_user_salesforce_info.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    @errors.push(error)
+  end
+
+  def notify_errors
+    return if @errors.empty?
+    Rails.logger.warn("UpdateUserSalesforceInfo errors: " + @errors.inspect)
+    DevMailer.pp(object: @errors,
+                 subject: "UpdateUserSalesforceInfo errors",
+                 to: SECRET_SETTINGS[:mail_recipients][:salesforce]).deliver
```
Dantemss: This will need to change to either `deliver_now` or `deliver_later` in Rails 4. (now is done inline, later is sent to a background job)

app/mailers/dev_mailer.rb
Dantemss: Oh maybe this is intended to be used for testing only?

app/models/anonymous_user.rb
```diff
@@ -80,4 +80,20 @@ def casual_name
     full_name
   end
 
+  User.faculty_statuses.each do |status, value|
+    define_method "#{status}?" do
+      User::DEFAULT_FACULTY_STATUS.to_s == status
+    end
+
+    define_method "#{status}!" do
```
Dantemss: You might want to raise an error here to make sure we don't attempt making an AnonymousUser faculty.

app/models/anonymous_user.rb
```diff
@@ -80,4 +80,20 @@ def casual_name
...
+    end
+  end
+
+  def faculty_status
+    User::DEFAULT_FACULTY_STATUS.to_s
+  end
+
+  def faculty_status=(status)
```
Dantemss: Same here.

app/representers/api/v1/user_representer.rb
```diff
@@ -40,6 +40,21 @@ class UserRepresenter < Roar::Decorator
...
              writeable: false
 
+    property :salesforce_contact_id,
+             if: ->(args) { args[:render_salesforce_info] },
+             type: String,
+             readable: true,
+             writeable: false
+
+    property :faculty_status,
+             if: ->(args) { args[:render_salesforce_info] },
```
Dantemss: These will have to change to use the `user_options` stuff in Rails 4.

app/routines/update_user_salesforce_info.rb
```diff
@@ -0,0 +1,148 @@
...
+      else
+        contacts_by_email[contact.email_alt] = contact
+        contacts_by_id[contact.id] = contact
+      end
+    end
+
+    # Go through all users that have already have a Salesforce ID and make sure
+    # their SF information is stil the same.
+
+    User.where("salesforce_contact_id IS NOT NULL")
```
Dantemss: Could use squeel here: `where{salesforce_contact_id != nil}`

app/routines/update_user_salesforce_info.rb
```diff
@@ -0,0 +1,148 @@
+class UpdateUserSalesforceInfo
+  USER_BLOCK_SIZE = 1000
+
+  def initialize
+    @errors = []
+  end
+
+  def self.call
+    new.call
```
Dantemss: Are we giving up on Lev? Why not use a Lev routine with `transaction: :no_transaction`?

app/routines/update_user_salesforce_info.rb
```diff
@@ -0,0 +1,148 @@
...
+        error!(exception: ee, user: user)
+      end
+    end
+
+    # Go through all users that don't yet have a Salesforce ID and populate their
+    # salesforce info when they have verified emails that match SF data.
+
+    sf_emails = contacts_by_email.keys
+
+    User.eager_load(:contact_infos)
```
Dantemss: I suggest also adding a `joins(:contact_infos)` to force an inner join. Might not actually matter.

app/routines/update_user_salesforce_info.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    contacts.each do |contact|
+      next if contact.email_alt.nil?
+
+      if contacts_by_email[contact.email_alt].present?
+        error!(message: "#{contact.email_alt} is an alt email on contact #{contact.id} but a " \
+                        "primary email on contact #{contacts_by_email[contact.email_alt].id}")
+      else
+        contacts_by_email[contact.email_alt] = contact
+        contacts_by_id[contact.id] = contact
```
Dantemss: This line might be unnecessary unless a contact can have ONLY an alt email?

app/views/dev_mailer/pp.text.erb
```diff
@@ -0,0 +1 @@
+<%= @object.ai(plain:true, ruby19_syntax: true, indent: 2) %>
```
Dantemss: `ai`? where is that from? and why ruby1.9?

app/controllers/admin/salesforce_controller.rb
Dantemss: If you think it should remain as an ActiveRecord table, though, you could do something like:

```
user_id = user.id
SalesforceUser.where{id != user_id}.delete_all
```

config/initializers/active_force.rb
```diff
@@ -0,0 +1,27 @@
+module ActiveForce
+
+  class << self
+
+    # Use a lazy setting of the client so that migrations etc are in place
+    # to allow the RealClient to be successfully instantiated.
+    alias_method :original_sfdc_client, :sfdc_client
+    def sfdc_client
+      if !original_sfdc_client.is_a?(Salesforce::Client)
+        self.sfdc_client = Salesforce::Client.new
```
Dantemss: RIP thread safety :)

config/secret_settings.yml.example
```diff
@@ -46,9 +46,16 @@ exception:
...
 valid_iframe_origins:
   - http://localhost
   - https://cnx.org
   - https://openstax.org
 
 disable_corner_console: false
+
+salesforce:
+  consumer_key: ~
+  consumer_secret: ~
```
Dantemss: Any changes to this file will conflict with the Rails 4 PR because this got renamed to `secrets.yml.example`. Also don't forget to add these to the deployment script.

lib/salesforce/client.rb
```diff
@@ -0,0 +1,20 @@
+module Salesforce
+  class Client < Restforce::Data::Client
+
+    def initialize
+      user = SalesforceUser.first
+      if user.nil?
+        Rails.logger.error { "The Salesforce client was requested but no user is available." }
+        raise Salesforce::UserMissing
+      end
+      secrets = SECRET_SETTINGS[:salesforce]
```
Dantemss: This will also have to change, probably to `Rails.application.secrets[:salesforce]` or similar.

spec/controllers/api/v1/group_members_controller_spec.rb
```diff
@@ -357,4 +356,12 @@
     end
   end
 
+  def user_hash(user)
```
Dantemss: Nice

app/controllers/admin/salesforce_controller.rb
jpslav: I'm ok with it being an ActiveRecord, tho agree it is a little odd to have a table with max one record.  I guess I could also stick this in an `after_create` callback on the AR.  I don't like those callbacks generally, but this is a reasonable case and it only deals with other instances of the AR (and failures there would cause the whole thing to rollback).

app/mailers/dev_mailer.rb
jpslav: It originally called pretty print.  Wanted a way to pretty format an object for use by developers.  `pp` turned out to not format how I wanted, so I switched to `awesome_print`.  

Will choose a better name.

app/routines/update_user_salesforce_info.rb
jpslav: No not giving up, just in these cases where we don't want a big transaction, and we aren't nesting this in another routine or calling other routines from it, and we aren't using lev's error functionality, I don't see as much of a need.

app/routines/update_user_salesforce_info.rb
jpslav: Good point.  I checked with Denver and he said primary email is required, but that there are some old Contacts that don't have it, so there is a (very low) chance that a Contact could have an alt email but not a primary.  So I might leave it (but with a comment)

app/views/dev_mailer/pp.text.erb
jpslav: I don't know what it stands for, but maybe `ascii inspect`:

image

the ruby 1.9 thing is just the ruby 1.9 hash syntax, i.e. `x: 'blah'` instead of `x => 'blah'`.

config/initializers/active_force.rb
jpslav: ha! what should I do instead? (actually did we thread safety-ize Accounts or just Tutor?)

app/mailers/dev_mailer.rb
jpslav: Gonna call it `inspect`

app/routines/update_user_salesforce_info.rb
Dantemss: Just optimization... But maybe the query planner can optimize that for you so it doesn't matter.

app/views/dev_mailer/pp.text.erb
Dantemss: Ah ok

config/initializers/active_force.rb
Dantemss: I haven't looked at Accounts.

What we should do really depends on the goal. Maybe worry later.

PR #356, opened 2016-10-16
Upgrade to Rails 4

BryanEliDimas:
Rails 4.2.7.1 upgrade.

* [x] Make a tutor-deployment PR that changes `include_vars: "templates/vars/rails3_app_config.yml.j2"` to `include_vars: "templates/vars/rails4_app_config.yml.j2"`, possibly among other things.

Deployment PR: https://github.com/openstax/tutor-deployment/pull/257


app/views/errors/any.html.erb
```diff
@@ -1,11 +1,9 @@
-<% content_for :application_body do %>
```
jpslav: know why this went away?

config/environments/development.rb
```diff
@@ -1,39 +1,38 @@
...
 
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
 
-  # Log error messages when you accidentally call methods on nil.
-  config.whiny_nils = true
+  # Do not eager load code on boot.
+  config.eager_load = true
```
jpslav: `eager_load = true` means we DO eager load code boot, right? (comment conflicts with code?)

config/environments/production.rb
```diff
@@ -64,9 +74,14 @@
...
   config.active_support.deprecation = :notify
 
-  # Log the query plan for queries taking more than this (works
-  # with SQLite, MySQL, and PostgreSQL)
-  # config.active_record.auto_explain_threshold_in_seconds = 0.5
+  # Disable automatic flushing of the log to improve performance.
+  # config.autoflush_log = false
+
+  # Use default logging formatter so that PID and timestamp are not suppressed.
+  config.log_formatter = ::Logger::Formatter.new
```
jpslav: Does this conflict with lograge, which we use for one-line logs?

config/environments/production.rb
```diff
@@ -64,9 +74,14 @@
...
+
+  # Do not dump schema after migrations.
+  config.active_record.dump_schema_after_migration = false
```
jpslav: neat

config/initializers/controllers.rb
```diff
@@ -77,13 +80,3 @@ def last_signin_provider
...
-
-# Layout is not inheritable in Rails 3.2
-Rails.application.config.to_prepare do
-  # Put this code inside the class_exec above in Rails 4
-  # and remove it from ApplicationController
-  [Doorkeeper::ApplicationController,
-   FinePrint::ApplicationController].each do |klass|
-    klass.layout 'application'
-  end
-end
```
jpslav: did you check that doorkeeper and fineprint render ok?

config/routes.rb
```diff
@@ -155,7 +156,4 @@
     end
   end
 
-  # Any other routes are handled here
-  match '*path', to: 'application#routing_error'
```
jpslav: any idea why we used to do this and we don't now?

lib/active_record_enum.rb
```diff
@@ -1,235 +0,0 @@
...
-        end
-      end
-  end
-end
-
-module ActiveRecord
-  class Base
-    extend ActiveRecord::Enum
-  end
-end
```
jpslav: woohoo

lib/omniauth/strategies/custom_identity.rb
```diff
@@ -79,7 +79,7 @@ def fail_with_log!(reason)
           user: locate_conditions[:user],
           remote_ip: request.ip,
           event_type: :sign_in_failed,
-          event_data: { reason: reason.to_s }
+          event_data: { reason: reason.to_s }.to_s
```
jpslav: looking around, is this supposed to be `.to_json` instead of `.to_s`?

lib/tasks/accounts/create_admin.rake
```diff
@@ -1,23 +1,24 @@
 namespace :accounts do
   desc 'Create an administrative user and/or set the password for said user'
   # Task specifically to create an administrative user and/or set the
-  # password for said user. This task accepts to arguments, the username
+  # password for said user. This task accepts two arguments: the username,
```
jpslav: i appreciate your attention two detail :-)

app/controllers/api/v1/application_groups_controller.rb
```diff
@@ -28,7 +28,8 @@ class Api::V1::ApplicationGroupsController < Api::V1::ApiController
   def updates
     OSU::AccessPolicy.require_action_allowed!(:updates, current_api_user, ApplicationGroup)
     outputs = GetUpdatedApplicationGroups.call(current_application).outputs
-    respond_with outputs[:application_groups], represent_with: Api::V1::ApplicationGroupsRepresenter, location: nil
+    respond_with outputs[:application_groups], represent_with: Api::V1::ApplicationGroupsRepresenter, \
+                                                 user_options: { render_contact_infos: false }, location: nil
```
Dantemss: It seems that the default is false. I'll remove the explicit falses.

app/controllers/application_controller.rb
```diff
@@ -1,17 +1,7 @@
 class ApplicationController < ActionController::Base
+
   include Lev::HandleWith
 
   respond_to :html
 
-  layout 'application'
-
-  # skip all filters defined so far
-  skip_filter *_process_action_callbacks.map(&:filter), only: [:routing_error]
```
Dantemss: Moved to ExceptionsController.

Basically when we find an exception we don't care if the user is logged in, we just want to display the error page.

app/controllers/exceptions_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  skip_filter *_process_action_callbacks.map(&:filter)
+
+  before_filter :set_locale
+
+  def rescue_from
+    @exception = env["action_dispatch.exception"]
+
+    OpenStax::RescueFrom.perform_rescue @exception, self
+  end
```
Dantemss: :) For some reason it was really hard to get it to render the exception page on a bad request (see the unknown route feature spec). I think even tutor-server doesn't handle those properly: https://tutor.openstax.org/%EB just gives a blank page.

app/models/contact_info.rb
```diff
@@ -11,18 +9,18 @@ class ContactInfo < ActiveRecord::Base
...
-  scope :email_addresses, where(type: 'EmailAddress')
+  scope :email_addresses, -> { where(type: 'EmailAddress') }
   sifter :email_addresses do type.eq 'EmailAddress' end
 
-  scope :verified, where(verified: true)
+  scope :verified, -> { where(verified: true) }
   sifter :verified do verified.eq true end
 
-  scope :with_users, lambda { includes(:user) }
+  scope :with_users, lambda { joins(:user) }
```
Dantemss: You are right... I'll change it back.

Dantemss: Or actually I might just make it a `.joins(:user).eager_load(:user)` to force an inner join instead of an outer join...

app/routines/admin/search_security_log.rb
```diff
@@ -132,7 +132,7 @@ def exec(params = {}, options = {})
             type_strings = to_string_array(types)
             sanitized_event_types = Admin::SearchSecurityLog.sanitize_event_types(type_strings)
 
-            @items = @items.where(event_type: sanitized_event_types)
+            @items = @items.where(event_type: sanitized_event_types.flatten)
```
BryanEliDimas: Yep, Rails 4 change. It gave a deprecation warning, _I think_ that's the only reason.
```DEPRECATION WARNING: Passing a nested array to Active Record finder methods is deprecated and will be removed. Flatten your array before using it for 'IN' conditions. (called from lib/composite_primary_keys/relation/predicate_builder.rb:32)```

File: app/views/errors/any.html.erb (hunk 1 above)
Comment (Dantemss) [inline app/views/errors/any.html.erb]: If you notice in accounts the error page never rendered properly (or at least hasn't rendered properly in a long time): https://accounts.openstax.org/%EB

This change makes the error actually show.

File: config/environments/development.rb (hunk 2 above)
Comment (Dantemss) [inline config/environments/development.rb]: You are right... it should be false for dev.

File: config/environments/production.rb (hunk 3 above)
Comment (Dantemss) [inline config/environments/production.rb]: From reading the lograge readme it seems to completely disable Rails's logging so this options shouldn't matter... probably.

File: config/initializers/controllers.rb (hunk 5 above)
Comment (Dantemss) [inline config/initializers/controllers.rb]: Doorkeeper as usual uses its own completely separate layout... FinePrint seems to work.

File: config/routes.rb (hunk 6 above)
Comment (Dantemss) [inline config/routes.rb]: I found that this solution wasn't working in Rails 4 so I switched to using the `exceptions_app` configuration in `config/application.rb`

We used to do it so we could render a 404 page for routes like https://accounts.openstax.org/abc and https://accounts.openstax.org/%EB

But in Rails 4 it was working only for the first and not the second route above.

File: lib/omniauth/strategies/custom_identity.rb (hunk 8 above)
Comment (Dantemss) [inline lib/omniauth/strategies/custom_identity.rb]: I should revert this and add `serialize :event_data` to the record.

File: app/controllers/application_controller.rb (hunk 11 above)
Comment (jpslav) [inline app/controllers/application_controller.rb]: ok.

File: app/models/identity.rb [hunk 15]
```diff
@@ -15,7 +15,6 @@ class Identity < OmniAuth::Identity::Models::ActiveRecord
   # omniauth-identity does not provide them by default
   # These attributes are (obviously) not saved in the database
   validates :password, presence: true, length: { minimum: 8, maximum: 50 }
-  validates :password_confirmation, presence: true
```
jpslav: hmm... ok I'll try to take a look after we merge this.

app/views/errors/any.html.erb
jpslav: Ah you know what... then this is why the API docs also aren't formatted correctly (should be syntax highlighting of the schemas, at least).  That also uses `content_for :application_body` which does not go anywhere (see `views/layouts/application_body_api_docs.html.erb`).

config/environments/production.rb
jpslav: can we take it out?

lib/tasks/accounts/create_admin.rake
jpslav: hey, I don't even get a "haha"? :-)

jpslav:
Thanks @BryanEliDimas and @Dantemss!

PR #365, opened 2016-11-04
Make first and last name mandatory

nathanstitt:
screen shot 2016-11-04 at 1 34 38 pm

screen shot 2016-11-04 at 1 23 03 pm


nathanstitt:
Hm, looks like travis only runs on PR's against master?

kathi-fletcher:
Quick questions. If you put a space or other non-visible char, can you fool it?

And does the check happen also if you edit your profile?

jpslav:
@kathi-fletcher all names are "stripped" (leading and trailing whitespace removed) any time the underlying user record is updated (https://github.com/openstax/accounts/blob/master/app/models/user.rb#L40) and now we check to make sure first and last name are present and not blank when the record is saved (https://github.com/openstax/accounts/pull/365/files#diff-4676c008b11a5480d73d4a6de01e45b9R53)

app/models/user.rb
```diff
@@ -50,7 +50,7 @@ class User < ActiveRecord::Base
   validates :state, inclusion: { in: VALID_STATES,
                                  message: "must be one of #{VALID_STATES.join(',')}" }
 
-  validate :name_part_required_for_suffix_or_title
+  validates :first_name, :last_name, presence: true, allow_blank: false
```
jpslav: We're going to have a problem with existing users who don't have first or last names.  E.g. when the Salesforce routine updates their SF information (which is cached in the User record), it won't be able to save because of this validation.  

Could do something like make the validation smarter:

```ruby
validates :names_present

def names_present
  # freak out if (first_name_was.present? && first_name.blank?) -- ditto last_name
  # different conditions for new_record?
  # something like this anyway :-)
end

File: app/models/user.rb [hunk 2]
```diff
@@ -133,7 +133,7 @@ def guessed_last_name
   end
 
   def casual_name
-    first_name.present? ? first_name : username
+    first_name
```
jpslav: Might need to leave this as is -- legacy users can have blank first names but they must have non-blank usernames.  And we need some way to say "Hi <name>" to them.  Can discuss if you want.

lib/salesforce/lead.rb
```diff
@@ -0,0 +1,15 @@
...
+    field :salutation,          from: "Salutation"
+    field :subject,             from: "Subject__c"
+    field :school,              from: "Company"
+    field :status,              from: "Status"
+    field :email,               from: "Email"
+    field :source,              from: "LeadSource"
+
+    self.table_name = 'Lead'
+  end
+end
```
jpslav: hopefully these Lead files go away on rebase...

app/routines/create_user.rb
```diff
@@ -54,7 +54,12 @@ def guessed_first_name(full_name)
 
   def guessed_last_name(full_name)
     return nil if full_name.blank?
-    full_name.split("\s").drop(1).join(' ')
+    parts = full_name.split("\s")
+    if parts.many?
+      parts.drop(1).join(' ')
+    else
+      full_name
+    end
```
jpslav: If full_name is just one "Jimmy", this change means `guessed_last_name` will be `"Jimmy"` whereas before it was `""`.  What was the reason for the change?

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -235,11 +225,11 @@
       expect{
         api_post :find_or_create,
                  trusted_application_token,
-                 raw_post_data: {email: 'a-new-email@test.com'}
+                 raw_post_data: {email: 'a-new-email@test.com', first_name: 'Ezekiel', last_name: 'Jones'}
       }.to change{User.count}.by(1)
       expect(response.code).to eq('201')
       new_user_id = User.order(:id).last.id
-      expect(response.body).to eq({id: new_user_id}.to_json)
+      expect(response.body).to eq({'id'=> new_user_id}.to_json)
```
jpslav: curious... why this change?  Looks identical.

```ruby
2.2.3 :006 > {id: 2}.to_json
 => "{\"id\":2}" 
2.2.3 :007 > {'id' => 2}.to_json
 => "{\"id\":2}" 
```

nathanstitt: ah, yes it is. I had an earlier failure on that line where i was using the new `users_hash` method here.  Once I figured out that wasn't needed, it looks like I updated the line to use string keys.  I've set it back now.

app/models/user.rb
nathanstitt: Yes, that makes sense give we're going to allow legacy to keep the nil name.

app/models/user.rb
nathanstitt: Yes that makes sense.  I had to get a bit more complex on the logic to require names on new records, but think what I came up with is the intended behavior

app/routines/create_user.rb
nathanstitt: This is an attempt to work around identity services that may not provide a last name.  If full_name doesn't contain spaces, then I thought it'd be best to fallback to the first name and still create the record.  Maybe it'd be best to just reject the user though?

app/handlers/sessions_create.rb
```diff
@@ -80,8 +80,8 @@ def handle
         authentication_user = users_matching_oauth_data.first
         status = :transferred_authentication
       else
-        outcome = run(CreateUserFromOmniauthData, @data)
-        authentication_user = outcome.outputs[:user]
+        # This will be updated to use a routine that does more
```
jpslav: If this is a todo, please start it off with `# TODO` and we'll make sure to circle back to all of these.

jpslav: I think this entire `else` block is probably going to be pulled out in a subsequent, since we will no longer have a way to create an account by a first-time log in with social

app/routines/create_user.rb
```diff
@@ -47,14 +47,4 @@ def generate_unique_valid_username(username)
...
 
-  def guessed_first_name(full_name)
-    return nil if full_name.blank?
-    full_name.split("\s")[0]
-  end
-
-  def guessed_last_name(full_name)
-    return nil if full_name.blank?
-    full_name.split("\s").drop(1).join(' ')
-  end
```
jpslav: so long twitter "full name" :-)

app/handlers/sessions_create.rb
```diff
@@ -80,8 +80,8 @@ def handle
...
+        authentication_user = User.new
```
jpslav: Does this user ever get saved?

app/routines/transfer_omniauth_data.rb
```diff
@@ -10,7 +10,21 @@ def exec(data, user)
     # This routine is not called for identity, so error out
     raise Unexpected if data.provider == 'identity'
 
+    user.username   = data.nickname
+    user.first_name = data.first_name.present? ? data.first_name : guessed_first_name(data.name)
+    user.last_name  = data.last_name.present?  ? data.last_name  : guessed_last_name(data.name)
```
jpslav: I don't think this always be called with a user where these fields aren't set yet. (see other parts of SessionsCreate).  So I don't think these should overwrite values that are already there.  So maybe these should all be `... if not blank`

nathanstitt:
ok @jpslav fixed up in response to your comments

nathanstitt:
@jpslav fixed up the specs.  I had to make a change to how usernames are computed.  We were using the “nickname" but that could have spaces and dots, which the user model wouldn't allow.

PR #367, opened 2016-11-06
rough new login flow

jpslav:


nathanstitt:
Looks like a good start for us to continue work off of @jpslav - merge when you're ready

config/initializers/controllers.rb
```diff
@@ -67,6 +67,26 @@ def expired_password
...
+      providers: cookies.signed[:login_providers]
+    }
+  end
+
+  def clear_login_info
+    cookies.delete(:login_key)
+    cookies.delete(:login_names)
+    cookies.delete(:login_providers)
+  end
+
```
jpslav: that clears the whole enchilada?  I just want to clear a specific part.  Maybe clearing the whole thing is ok, but I could step on other's toes it seems.

PR #374, opened 2016-11-10
Made usernames optional

Dantemss:
This works and all but maybe we want to keep a hidden autogenerated username?


Dantemss:
Closing to make PR towards integration branch.

PR #376, opened 2016-11-11
Added option to auto verify emails for accounts-demo

Dantemss:
Requires https://github.com/openstax/tutor-deployment/pull/261


Dantemss:
OBE

PR #380, opened 2016-11-16
Signup styles n spec

nathanstitt:
Todo: implement TODO's and write more specs
screen shot 2016-11-16 at 5 26 30 pm

### DIFF
--- a/app/handlers/signup_start.rb
+++ b/app/handlers/signup_start.rb
@@ -23,7 +23,8 @@ class SignupStart
     end
 
     # If email in use, want users to login with that email, not create another account
-    fatal_error(code: :email_in_use) if email_in_use?
+    return outputs[:status] = :email_in_use if email_in_use?
+
 
     # Blow away the user's existing signup email, if it exists
     existing_signup_contact_info.try(:destroy)
--- a/app/handlers/signup_start.rb
+++ b/app/handlers/signup_start.rb
@@ -23,8 +23,7 @@ class SignupStart
     end
 
     # If email in use, want users to login with that email, not create another account
-    return outputs[:status] = :email_in_use if email_in_use?
-
+    fatal_error(code: :email_in_use) if email_in_use?
 
     # Blow away the user's existing signup email, if it exists
     existing_signup_contact_info.try(:destroy)
--- a/app/views/signup/start.html.erb
+++ b/app/views/signup/start.html.erb
@@ -22,7 +22,7 @@ signup_roles = {
                                 context: self,
                                 errors: @handler_result.try(:errors)) %>
     <%# TODO put select in form helper %>
-    <%# TODO put select options in translations %>
+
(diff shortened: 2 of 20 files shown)


app/handlers/signup_start.rb
```diff
@@ -23,7 +23,8 @@ def handle
     end
 
     # If email in use, want users to login with that email, not create another account
-    fatal_error(code: :email_in_use) if email_in_use?
+    return outputs[:status] = :email_in_use if email_in_use?
```
jpslav: ah, in this case you can just use the `fatal_error` -- that'll take you to the `failure` block in the `handle_with` call.  In there, you can do different things based on `@handler_result.errors`.  Output statuses are good, but typically I like to still split them into the `success` and `failure` camps.

app/views/signup/start.html.erb
```diff
@@ -10,11 +22,24 @@
                                 context: self,
                                 errors: @handler_result.try(:errors)) %>
     <%# TODO put select in form helper %>
-    <%# TODO put select options in translations %>
-    <%= f.select :role, options_for_select(%w(Instructor Student).map{|role| [role, role]},
-                        [saved_role].compact) %>
-    <%= fh.text_field name: :email, value: saved_email %>
-    <%= f.submit (t :".next"), class: 'btn btn-primary' %>
+
+    <%= f.select :role, options_for_select(signup_roles.invert.to_a, saved_role) %>
```
jpslav: I think `saved_role` might be capitalized... so the selected option might not work?  Or rather, might it be working now because the english translation has it capitalized...

spec/features/user_signs_up_spec.rb
```diff
@@ -63,15 +61,18 @@
...
     end
 
     scenario 'failure because suspected non-school email then success' do
       visit signup_path
-      complete_signup_email_screen("Instructor","bob@gmail.com")
-      skip # TODO Nathan should this be a JS check to see that email is not @blah.edu?
-      # checking in BE might be a waste
+      select 'Instructor', from: "signup_role"
+      fill_in (t :"signup.start.email"), with: "bob@gmail.com"
+      click_button(t :"signup.start.next")
```
jpslav: can still use the helper `complete_signup_email_screen`?  ditto above.

jpslav: oh... is the JS check happening before the next click?

nathanstitt: yeah, it's happening on the first click.  then the warning message is displayed if the email isn't edu.  I updated `complete_signup_email_screen` to do two clicks, but that means that it can't be used to actually test the message so this spec needs to do the steps itself.

app/views/signup/start.html.erb
nathanstitt: No, `saved_role` is set to the value of the select field.  I updated the select field to set the value as lowercase symbols and display the proper translated string

PR #388, opened 2016-11-22
single email per account / multi email login validation 

nathanstitt:
Hey, I made this one against the correct branch on the FIRST try

This enforces no shared emails between accounts.  Existing duplicates are ok, but a new email can't be added or it's `value` updated to have a duplicate email.

This implements the error message for when someone attempts to login with an email that's shared between multiple accounts.

screen shot 2016-11-22 at 2 04 59 pm

### DIFF
--- a/spec/factories/user.rb
+++ b/spec/factories/user.rb
@@ -2,8 +2,8 @@ FactoryGirl.define do
   factory :user do
     username { SecureRandom.hex(3) }
     state 'activated' # otherwise the default from DB will be to 'temp'
-    first_name Faker::Name.first_name
-    last_name  Faker::Name.last_name
+    first_name { Faker::Name.first_name }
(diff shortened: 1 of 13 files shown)


spec/factories/user.rb
```diff
@@ -2,8 +2,8 @@
   factory :user do
     username { SecureRandom.hex(3) }
     state 'activated' # otherwise the default from DB will be to 'temp'
-    first_name Faker::Name.first_name
-    last_name  Faker::Name.last_name
+    first_name { Faker::Name.first_name }
+    last_name  { Faker::Name.last_name  }
     trait :admin do
```
nathanstitt: this was a silly bug that was throwing off the specs.  User's were being created all with the same name, and the routine was calling `.uniq` on the names, making it look like there was only a single user with that email.

PR #420, opened 2016-12-08
support alternate signup URL

jpslav:


config/initializers/doorkeeper.rb
```diff
@@ -5,6 +5,8 @@
 
   # This block will be called to check whether the resource owner is authenticated or not.
   resource_owner_authenticator do
+    # JP doesn't think this is ever getting called because ActionController::Base has a
+    # before_filter :authenticate_user!
```
Dantemss: You are probably right. Well it will get called, but `authenticate_user!` will always just let it pass through. I think you can remove `authenticate_user!` but it might have to return the `current_user` still. (or maybe not , I'm not sure)

lib/user_session_management.rb
```diff
@@ -105,4 +108,43 @@ def signup_email
...
+
+  def set_alternate_signup_url(url)
+    if url.blank?
+      session[:alt_signup] = nil
+    elsif !url.is_a?(String)
+      raise IllegalArgument
+    elsif is_redirect_url?(application: get_client_app, url: url)
+      session[:alt_signup] = url
+    else
+      session[:alt_signup] = nil
```
Dantemss: Could move these ifs into `is_redirect_url?` but meh...

Would simplify the logic in this function though...

PR #442, opened 2016-12-16
Remove min-height on card body

nathanstitt:
This allows the footer to move up beside the inputs.

Note: The footer has 30px top padding + 30px top margin, which means it'll always have 60px of space between the last element and it.

I think that was by design though, so I've left it.  We can adjust if @Fredasaurus thinks best.

screen shot 2016-12-15 at 6 07 11 pm

screen shot 2016-12-15 at 6 07 04 pm

### DIFF
--- a/app/assets/stylesheets/controls/_card.scss
+++ b/app/assets/stylesheets/controls/_card.scss
@@ -30,9 +30,6 @@
     width: 50%;
     float: left;
   }
-  .card-body {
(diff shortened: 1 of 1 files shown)


Fredasaurus:
Yeah, I'd like, literally, 30px space between the field and the button.

Fredasaurus:
We should also have more space between the button and the text below it.

PR #445, opened 2016-12-16
fb and rescue from fixes (HOTFIX)

jpslav:
Includes the Facebook hotfix (https://github.com/openstax/accounts/pull/440) plus a hotfix to upgrade openstax_rescue_from so that the `OpenStax::RescueFrom.this{...}` call in the cron jobs will work.


Dantemss:
Already on master.

PR #454, opened 2017-01-04
only email SF user update errors on real production

jpslav:
Otherwise it is confusing to SF people because errors from -dev or -qa aren't really something they should deal with.

### DIFF
--- a/app/routines/update_user_salesforce_info.rb
+++ b/app/routines/update_user_salesforce_info.rb
@@ -1,12 +1,12 @@
 class UpdateUserSalesforceInfo
 
-  def initialize(enable_error_email:)
-    @enable_error_email = enable_error_email
+  def initialize(allow_error_email:)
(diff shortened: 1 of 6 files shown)


app/routines/update_user_salesforce_info.rb
```diff
@@ -140,13 +140,13 @@ def notify_errors
...
-      subject: "UpdateUserSalesforceInfo errors",
-      to: Rails.application.secrets[:salesforce]['mail_recipients']
-    ).deliver_later
+    if @allow_error_email && is_real_production?
+      DevMailer.inspect_object(
+        object: @errors,
+        subject: "UpdateUserSalesforceInfo errors",
+        to: Rails.application.secrets[:salesforce]['mail_recipients']
+      ).deliver_later
+    end
```
jpslav: The `is_real_production?` is the only substantive change

PR #459, opened 2017-01-10
Decode alt signup urls and disable alt signup error emails

Dantemss:


jpslav:
@Dantemss is going to add a spec

Dantemss:
@philschatz can you turn off comments in the team config? Seems like the `.codecov.yml` approach is still broken

Dantemss:
@philschatz actually leave it for now... maybe it will fix itself once PR's are based off of master (the issue might be that the PR's don't have .codecov.yml yet)

lib/user_session_management.rb
```diff
@@ -121,6 +121,12 @@ def get_client_app
                       Doorkeeper::Application.find_by(id: session[:client_app])
   end
 
+  def is_redirect_url?(application:, url:)
+    return false if application.nil? || url.nil?
+    # Let doorkeeper do the work of checking the URL against the app's redirect_uris
+    Doorkeeper::OAuth::Helpers::URIChecker.valid_for_authorization?(url, application.redirect_uri)
+  end
```
jpslav: Guessing you moved this method here from `ApplicationController` so it could be available in the spec, but this method doesn't belong in this module.  This module is about accessing the session for user stuff.  Can you move this method to some other library file or even monkey patch the doorkeeper application model with `is_redirect_url?(url)`?

spec/lib/user_session_management_spec.rb
```diff
@@ -0,0 +1,398 @@
...
+    end
+
+    it 'sign_in! can be used to sign in' do
+      controller.sign_in! user_1
+      expect(controller.current_user).to eq user_1
+    end
+
+    it 'sign_out! does nothing' do
+      controller.sign_out!
+      expect(controller.current_user).to eq AnonymousUser.instance
```
jpslav: also expect to receive `clear_signup_state`?

spec/lib/user_session_management_spec.rb
```diff
@@ -0,0 +1,398 @@
...
+  let(:user_2)     { FactoryGirl.create(:user) }
+
+  let(:controller) { ActionController::Base.new }
+  let(:main_app)   do
+    Class.new do
+      def login_path(params = {})
+        'https://localhost/login'
+      end
+    end.new
+  end
```
jpslav: Interesting how you set this up.

spec/lib/user_session_management_spec.rb
```diff
@@ -0,0 +1,398 @@
...
+      expect(controller.is_redirect_url?(application: app, url: url)).to eq false
+      expect(Doorkeeper::OAuth::Helpers::URIChecker).to(
+        receive(:valid_for_authorization?).with(app.redirect_uri, app.redirect_uri)
+                                          .and_call_original
+      )
+      expect(controller.is_redirect_url?(application: app, url: app.redirect_uri)).to eq true
+    end
+  end
+
+end
```
jpslav: nice spec

lib/user_session_management.rb
Dantemss: Sure

spec/lib/user_session_management_spec.rb
Dantemss: I could add that

spec/lib/user_session_management_spec.rb
Dantemss: Thanks :D

PR #460, opened 2017-01-11
(DO NOT MERGE) Allowing updating role on profile screen

nathanstitt:
screen shot 2017-01-11 at 11 04 27 am


jpslav:
UX going to advise

Fredasaurus:
I think we should. Form would need to be reloaded with appropriate fields.

nathanstitt:
Closing this until we gather more data to see if this is a problem we need to solve or not

PR #462, opened 2017-01-12
switch from coveralls to codecov

philschatz:
As part of https://github.com/openstax/napkin-notes/issues/68 this switches from coveralls to codecov.

I was not sure about a couple things but took some defaults:

- [x] When should coveralls make comments on the PR and what should be included
- [x] I used this recommendation from codecov.io: https://github.com/codecov/example-ruby

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+comment:
+  layout: "header, diff, changes, uncovered, tree"
+  behavior: default
+
+# To Turn off comments completely:
+# comment: false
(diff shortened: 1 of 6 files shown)


jpslav:
@philschatz I'd rather not have codecov comment on PRs.  I found them too loquacious with coveralls and not all that helpful.   

And I have no recollection of what simplecov does, @Dantemss do you know?

Dantemss:
@jpslav Simplecov reports code coverage, either to a service like coveralls or codecov or as an HTML page that is stored in the `coverage` folder (but this PR disables the HTML page part)

Dantemss:
You could probably disable the comment in codecov somewhere...

philschatz:
hehe, configuring comments is the very 1st file in the PR diff: `.codecov.yml` 😄

PR #464, opened 2017-01-13
Require at least one selection from titles on profile page

nathanstitt:
screen shot 2017-01-13 at 4 27 24 pm


kathi-fletcher:
Please change "titles" to "subjects" in "Select all OpenStax titles ..." to match the field on the website and the error message.

PR #465, opened 2017-01-14
Handle IllegalState situations by redirecting to the home page with a message

Dantemss:


app/controllers/sessions_controller.rb
```diff
@@ -107,11 +107,9 @@ def create
           redirect_to profile_path,
                       alert: "That way to log in cannot be added because it is associated to an email address that is already in use!" # TODO i18n
         else
-          Rails.logger.fatal "IllegalState: OAuth data: #{request.env['omniauth.auth']}; " \
-                             "status: #{@handler_result.outputs[:status]}"
-          raise IllegalState, "SessionsCreate errors: #{@handler_result.errors.inspect
-                              }; Last exception: #{$!.inspect}; Exception backtrace: #{$@.inspect
-                              }; status: #{@handler_result.outputs[:status]}"
+          Rails.logger.fatal "IllegalState: OAuth data: #{request.env['omniauth.auth']
+                             }; status: #{@handler_result.outputs[:status]}"
```
jpslav: Won't this result in weird spacing and a line break in the message?  

```
[1] pry(main)> Rails.logger.fatal "hi 
[1] pry(main)*         there"
hi 
        there
```

Maybe the original just needed some parens around the string?

jpslav: Don't we still want the exception so that we can know that a bad status is getting returned?

app/handlers/sessions_create.rb
```diff
@@ -36,7 +36,8 @@ class SessionsCreate
   protected
 
   def setup
-    @data = OmniauthData.new(request.env['omniauth.auth'])
+    @data = OmniauthData.new(request.env['omniauth.auth']) \
+      rescue fatal_error(code: :invalid_omniauth_data)
```
jpslav: Need to handle this error code in SessionsController `handle_with` call?

app/controllers/sessions_controller.rb
```diff
@@ -107,11 +107,9 @@ def create
...
+          redirect_to root_path, alert: 'Sorry, we lost you. Please log in again.'
```
jpslav: I think this should be more of the 500 error case where we show a 500 error and get an email, b/c we are not expecting users to get to this `else` block.

app/controllers/sessions_controller.rb
Dantemss: In that case we should add handlers for `unknown_callback_state` and `invalid_omniauth_data` so we don't get emails from these 2.

app/controllers/sessions_controller.rb
Dantemss: Sure we can leave the else block if we add handlers for `unknown_callback_state` and `invalid_omniauth_data`.

app/handlers/sessions_create.rb
Dantemss: I was planning to handle this in the else block but sure we can do this instead.

app/controllers/sessions_controller.rb
jpslav: I think we want emails for unknown callback state (because that means we're not handling something)?

And invalid omniauth data -- seems like we want emails for that too so we can detect if we're not handling something correctly?

app/controllers/sessions_controller.rb
jpslav: See above comment

PR #466, opened 2017-01-14
Prevent password reset from generating 500 errors when user not found

Dantemss:


app/controllers/identities_controller.rb
```diff
@@ -38,7 +38,10 @@ def send_password_email(kind:)
     # Can be reached before logged in (can't remember password) or when logged
     # in and asked to reauthenticate and can't remember password.
 
-    user = signed_in? ? current_user : User.find(get_login_state[:matching_user_ids].first)
+    user = signed_in? ? current_user :
+                        User.find_by(id: get_login_state[:matching_user_ids].try(:first))
+
+    redirect_to(root_path, alert: 'Sorry, we lost you. Please try again.') && return if user.nil?
```
jpslav: Can you put into `en.yml`?

spec/controllers/identities_controller_spec.rb
```diff
@@ -102,5 +102,31 @@ def reset_password(password, confirmation)
...
+    context 'send_add' do
+
+      it 'redirects to the home page with a message if the user is not found' do
+        post :send_add
+        expect(response).to redirect_to root_path
+        expect(flash.alert).to eq I18n.t(:'controllers.identities.lost_user')
+      end
+
+      xit 'sends a message to add a password to the account if the user is found' do
+      end
```
jpslav: These `xit` success cases are already in `spec/features/reset_password_spec.rb`.  Want to remove them from here?

PR #469, opened 2017-01-17
Fix user lookup bug and prevent direct callback crash

jpslav:
Direct calls to `/auth/identity/callback` were causing nilclass exceptions.  Root cause was an errant user lookup by nil username (now we have nil usernames).

### DIFF
--- a/lib/lookup_users.rb
+++ b/lib/lookup_users.rb
@@ -2,6 +2,9 @@ module LookupUsers
 
   def self.by_email_or_username(email_or_username)
     email_or_username = email_or_username.try(:strip)
+
+    return [] if email_or_username.blank?
+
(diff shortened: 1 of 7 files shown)


lib/lookup_users.rb
```diff
@@ -2,6 +2,9 @@ module LookupUsers
 
   def self.by_email_or_username(email_or_username)
     email_or_username = email_or_username.try(:strip)
+
+    return [] if email_or_username.blank?
```
jpslav: This was the fix for the underlying problem causing us to errantly reach the `source = request.params["login"]["source"]` line below which gave a nil class error.

jpslav:
Merging because travis passed.  Codecov issue... maybe codecov only uses `.codecov.yml` from master, not from the PR?

PR #471, opened 2017-01-21
allow case insensitive login if no case sensitive matches

jpslav:
Allow case insensitive login if no case sensitive match found.

Relatedly, handle the case where a user has multiple emails with different case, but no usernames (possible if created between Jan 2017 and Feb 2017 releases.  Normally people with multiple accounts with the same email address who try to login with an email address would get an email saying "you have to log in with your username, which is 'foo' or 'bar'".  But when there are no usernames, this message would be bad and so we instead say "contact support".

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -110,7 +110,7 @@ en:
         We did not find an account with the username or email you provided.
       found_account_but_it_has_no_emails: >-
         We found your account but can't send you an email because your account
-        doesn't have any email addresses listed. Please contact support for
+        doesn't have any email addresses listed. Please contact Support for
         assistance.
 
     signup_password:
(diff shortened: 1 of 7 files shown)


jpslav:
Failing for some weird travis / rubygems bug apparently.  Retry later.

config/locales/en.yml
```diff
@@ -110,7 +110,7 @@ en:
         We did not find an account with the username or email you provided.
       found_account_but_it_has_no_emails: >-
         We found your account but can't send you an email because your account
-        doesn't have any email addresses listed. Please contact support for
+        doesn't have any email addresses listed. Please contact Support for
```
jpslav: marcom has asked us to say "Support" instead of "support" or "customer service" or "customer support".

PR #475, opened 2017-01-25
Sort subjects list so "Not Listed" appears at end

nathanstitt:
screen shot 2017-01-25 at 1 46 22 pm

### DIFF
--- a/lib/settings/subjects.rb
+++ b/lib/settings/subjects.rb
@@ -7,7 +7,10 @@ module Settings
       def each(sorted: false)
         list = Settings::Db.store.subjects
         if sorted
-          list = list.sort_by{ |subject| subject.last['title']}
+          list = list.sort_by do |subject|
+            # force "not listed" to appear at end
+            subject.first == 'not_listed' ? 'zzzz' : subject.last['title']
+          end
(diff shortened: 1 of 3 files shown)


lib/settings/subjects.rb
```diff
@@ -7,7 +7,10 @@ class << self
       def each(sorted: false)
         list = Settings::Db.store.subjects
         if sorted
-          list = list.sort_by{ |subject| subject.last['title']}
+          list = list.sort_by do |subject|
+            # force "not listed" to appear at end
+            subject.first == 'not_listed' ? 'zzzz' : subject.last['title']
```
jpslav: Where does `'not_listed'` come from?

nathanstitt: It coms from either the seed data in `config/app.yml` or from whatever is set by the admin in `rails-settings-ui`  

As you can gather, this isn't super great since it relies on the key being set to `not_listed` and will break if someone sets it to `not-listed` or something else.

Without a real schema, I think this is about the best we can do though

jpslav:
Looks good.  Wait to merge until hotfix approved.

PR #479, opened 2017-01-26
don't retry emails after one AWS::SES::ResponseError

jpslav:


jpslav:
@Dantemss reviewed said good.

PR #480, opened 2017-01-27
full console user search with restored functionality

jpslav:
Leaves the current dialog-based user search as is, but adds a new user search on the "full console" screen.  

image

* Role is shown in the search results (can't currently be searched, but can be sorted)
* Sorting works now via the `Ordered by` field (see "Search help" page at bottom)
* Faculty status is shown in the search results
* Search doesn't use Javascript, so you can copy and paste a search URL and share it
* Extra details are expandable under each search result (admin status, account state, created at / updated at dates, self-reported school, securlty log link (fixed), full name, emails and confirmation status)
* Edit the user by clicking the ID hyperlink
* Log in as the user by clicking the "key" button at the far right of each result
* And sorry if you didn't know this, but you could always do complex searches like `first_name:bobby last_name:tables`.

also adds fine print to the full admin console and layout

### DIFF
--- a/app/assets/javascripts/signin/login.coffee
+++ b/app/assets/javascripts/signin/login.coffee
@@ -5,7 +5,6 @@ class OX.Signin.Login
     new OX.Signin.Login(card) if card.length
 
   constructor: (@el) ->
-    console.log @el.find('a.trouble')
     @el.find('a.trouble').click(@onHelpClick)
 
   onHelpClick: (ev) =>
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1,6 +1,7 @@
 @import "bootstrap";
 @import "common_colors";
 @import "admin_navbar";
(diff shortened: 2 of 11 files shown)


app/assets/javascripts/signin/login.coffee
```diff
@@ -5,7 +5,6 @@ class OX.Signin.Login
     new OX.Signin.Login(card) if card.length
 
   constructor: (@el) ->
-    console.log @el.find('a.trouble')
```
jpslav: gets rid of `[object Object]` in the rspec runs.

nathanstitt: Sigh, so it was me :(

I was searching for `puts` and `p` in the specs, didn't think to check JS.  Good catch @jpslav

jpslav: No worries @nathanstitt, thanks for all the fast work on Accounts

PR #482, opened 2017-01-27
Hide the role selection when :go param is "signup_student"

nathanstitt:
Used by tutor to redirect to signup and also force the role to be student


spec/features/user_signs_up_spec.rb
```diff
@@ -115,6 +115,13 @@
       expect(page).not_to have_content t('signup.start.teacher_school_email')
     end
 
+    scenario 'profile selection is set to student when coming from "student_signup"' do
```
jpslav: Can you change to "...is set to student and hidden when..." and then add a `screenshot!` call after the `visit` call?  I'd like to have this case included in the screenshots.

PR #484, opened 2017-01-29
better admin user details page

jpslav:
Improves the admin page for editing / viewing user details.

Big points:
* Cleaner form
* Lets admins change usernames (even lets them set usernames when one didn't exist before, helps for users who in January created two accounts with the same email with different capitalization)
* Shows the applications that the account has been used for (e.g. tutor, osweb)
* For Tutor, provides a link to an admin "info" page on Tutor where you can see which courses the user is a part of.  This is the easy alternative to seeing Tutor information displayed inline in Accounts.  See https://github.com/openstax/tutor-server/pull/1355  
* Shows the ways the user can log in
* Gives a link to the security log
* Link to log in as user (the key next to the name)

image

### DIFF
--- a/app/controllers/admin/contact_infos_controller.rb
+++ b/app/controllers/admin/contact_infos_controller.rb
@@ -5,10 +5,10 @@ module Admin
       contact_info = ContactInfo.find(params[:id])
       result = MarkContactInfoVerified.call(contact_info)
       if result.errors.any?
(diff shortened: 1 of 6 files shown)


nathanstitt:
looks good @jpslav - when we have some more time to dedicate to this, it'd be nice to make the columns on the search sortable when clicked.

PR #486, opened 2017-02-06
Treat unprefixed two word searches as first/last name

nathanstitt:
On production a search for common names is pulling up to many results to be useful.

This requires both last name and first name to match if the term is two words, which is the most common search.


Dantemss:
Not a huge fan of making n == 2 a special case, but I guess it works...

nathanstitt:
Yeah I know, but it is only for the un-prefixed case, where the searcher just types something like `john smith`.  If someone wants to be specific they can prefix the terms so we know what to do

PR #492, opened 2017-02-20
Report pending faculty status

jpslav:
Marks Accounts users as `pending_faculty` when they don't match to an applicable Salesforce `Contact` but they do match to a Salesforce `Lead`.

Uses a new `openstax_salesforce` gem that centralizes some of our Salesforce code.

Introduces automated tests connected to Salesforce sandboxes.

Moves Accounts to Ruby 2.3.3

When this gets merged and deployed, someone will need to go the admin settings page and check the box to send user info update error emails (should only be permanently checked on production, but can be checked on other environments as needed for testing)


app/routines/update_user_salesforce_info.rb
```diff
@@ -10,34 +10,53 @@ def self.call(allow_error_email: false)
   end
 
   def call
-    return if !SalesforceUser.any? && !is_real_production?
+    return if !OpenStax::Salesforce.ready_for_api_usage?
+    return if !Rails.application.is_real_production?
```
Dantemss: Can we just leave this out now that there are sandboxes?

Dantemss: Can we just leave this out now that there are sandboxes?

app/routines/update_user_salesforce_info.rb
```diff
@@ -143,7 +198,7 @@ def notify_errors
     return if @errors.empty?
     Rails.logger.warn("UpdateUserSalesforceInfo errors: " + @errors.inspect)
 
-    if @allow_error_email && is_real_production?
+    if @allow_error_email && Rails.application.is_real_production?
```
Dantemss: Maybe it would be a good idea to change this to something specific in the secrets file, like:

```
salesforce:
  enable_exception emails: true
```

or something

app/routines/update_user_salesforce_info.rb
jpslav: good point

app/routines/update_user_salesforce_info.rb
jpslav: might make it a editable setting so that QA can turn it on and off to test @Dantemss

app/routines/update_user_salesforce_info.rb
```diff
@@ -10,34 +10,52 @@ def self.call(allow_error_email: false)
   end
 
   def call
-    return if !SalesforceUser.any? && !is_real_production?
+    return if !OpenStax::Salesforce.ready_for_api_usage?
```
jpslav: yeah this is a weird thing with my brain.  for some reason I read `if !blah` just fine but every time I see `unless blah` I have to think about it for too long. :-)

nathanstitt: yeah, it's really 6 of one; 1/2 dozen type thing. no worries

app/routines/update_user_salesforce_info.rb
```diff
@@ -10,34 +10,52 @@ def self.call(allow_error_email: false)
...
       next if contact.email_alt.nil?
 
-      if contacts_by_email[contact.email_alt].present?
-        error!(message: "#{contact.email_alt} is an alt email on contact #{contact.id} but a " \
-                        "primary email on contact #{contacts_by_email[contact.email_alt].id}")
+      if (colliding_contact = contacts_by_email[contact.email_alt]).present?
+        colliding_emails.push(contact.email_alt)
+        error!(message: "#{contact.email_alt} is an ALT email on contact #{contact.id} and a " \
+                        "either a PRIMARY or ALT email on contact #{colliding_contact.id}")
       else
```
jpslav: good man!

nathanstitt:
Looks good @jpslav - sorry for the delay in re-examining it

PR #494, opened 2017-02-24
Fixed SF bug and added push lead spec

jpslav:
Not all old SF code has specs, and this bug fixes one that didn't and that didn't handle the recent SF changes well.

### DIFF
--- a/app/routines/push_salesforce_lead.rb
+++ b/app/routines/push_salesforce_lead.rb
@@ -39,7 +39,7 @@ class PushSalesforceLead
     if lead.errors.any?
       handle_errors(lead, user, role)
     else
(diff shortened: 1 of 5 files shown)


app/routines/push_salesforce_lead.rb
```diff
@@ -17,11 +17,11 @@ def exec(user:, email: nil, role:, phone_number:, school:, num_students:, using_
 
     source = (role || "").match(/instructor/i) ? "OSC Faculty" : "OSC User"
 
-    lead = Salesforce::Lead.new(
+    lead = OpenStax::Salesforce::Remote::Lead.new(
```
jpslav: This was the bug

PR #497, opened 2017-03-06
Re-enabling skipped specs, doing TODOs

jpslav:


lib/omniauth/strategies/custom_identity.rb
```diff
@@ -53,28 +50,8 @@ def protect_against_forgery?
...
-          user_attempts = SecurityLog.sign_in_failed.where{created_at > user_attempts_time}
-                                                    .where(user: user).count
-        end
-
-        return true if user_attempts >= MAX_LOGIN_ATTEMPTS_PER_USER
-
-        false
+        too_many_log_in_attempts_by_ip?(ip: request.ip) ||
+        too_many_log_in_attempts_by_user?(user: locate_conditions[:user])
       end
```
nathanstitt: If we wrote `RateLimiting` to have a single `too_many_login_attemps?(ip: , user: )` we could perform one call here and eventually a single SQL statement.

lib/rate_limiting.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    user_attempts = SecurityLog.sign_in_failed
+                               .where{created_at > attempts_considered_since}
+                               .where(user: user)
+                               .count
+
+    user_attempts >= max_attempts
+  end
+
+end
```
nathanstitt: As I mentioned above, we could have a single method.

```ruby
def too_many_login_attemps?(ip: , user: )
  query = ""
  query << "IP CLAUSE" if ip
  query << "USER CLAUSE" if user
  where(query)
end
```

Query building won't work quite like that of course and `or` statements are a mess in straight AR http://stackoverflow.com/questions/18860959/rails-activerecord-where-or-clause

Given that, maybe isn't worth it.  I'm not sure how performance critical this really is, since it's not called on every request, just on login attempts.

lib/omniauth/strategies/custom_identity.rb
jpslav: Kind of like it separate because we call `too_many_log_in_attempts_by_ip?` on its own in `SessionsController`, and it makes a bit more sense when reading it on its own vs passing in `nil` for `ip` or `user`.  Get what you're saying tho.

lib/rate_limiting.rb
jpslav: Yeah I'm not too worried about the multiple calls since it is just log in

PR #500, opened 2017-03-08
get SF user info 3 times an hour

jpslav:
instead of just once an hour; also have first run be closer to top of hour so that we are more certain to not have run out of API calls.


nathanstitt:
Looks good @jpslav

PR #501, opened 2017-03-08
Internationalised model validation messages

aiwenar:
This PR internationalises validation messages for ActiveRecord models and Lev handlers.

To account for non-SOV and non-SVO languages the default error message format has been changed from `"%{attribute}" "%{message}"` to just `"%{message}"`. While not necessary, this has been done also for English, as this locale will serve as a base for future translations. Default validation messages have also been changed to reflect this new format.

Error messages have been moved from main language files to `activerecord.{lang}.yml`. All error messages which may be returned to the user have been declared in translation files, including cases where the validations are using default messages, so that translation files explicitly list all user-visible messages.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 # Rails framework
 gem 'rails', '4.2.7.1'
+gem 'rails-i18n', '~> 4'
 
 # Bootstrap front-end framework
 gem 'bootstrap-sass', '~> 3.1.1'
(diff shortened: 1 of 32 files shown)


jpslav:
Hi @aiwenar - can you resolve the travis errors? (also you can make branches in the Accounts repo if you want)

aiwenar:
OK, took me way too long but I found it. Apologies for the delay.

jpslav:
Thanks @aiwenar!

PR #503, opened 2017-03-08
fix badly formatted event_data in some SecurityLogs

jpslav:
A larger number of `SecurityLog` records on production have badly-formatted `event_data`, e.g. it looks like:

```
"---\n:reason: cannot_find_user\n"
```

When these logs are loaded, the deserialization code raises a `JSON::ParserError` exception.  This prevents CS from looking at security logs for impacted users.

This PR adds a rake task that fixes the bad cases.

### DIFF
--- a/lib/tasks/fix_bad_security_log_event_data.rake
+++ b/lib/tasks/fix_bad_security_log_event_data.rake
@@ -1,0 +1,16 @@
+desc "Find possible duplicate acccounts"
+task fix_bad_security_log_event_data: [:environment] do
+
+  bad_to_fixed_event_data = {
+    "---\n:reason: cannot_find_user\n" =>        "{\"reason\":\"cannot_find_user\"}",
+    "---\n:reason: bad_password\n" =>            "{\"reason\":\"bad_password\"}",
+    "---\n:reason: multiple_users\n" =>          "{\"reason\":\"multiple_users\"}",
+    "---\n:reason: too_many_login_attempts\n" => "{\"reason\":\"too_many_login_attempts\"}"
+  }
+
+  bad_to_fixed_event_data.each do |bad, fix|
+    ActiveRecord::Base.connection.execute(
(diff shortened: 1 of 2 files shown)


lib/tasks/fix_bad_security_log_event_data.rake
```diff
@@ -0,0 +1,16 @@
...
+  bad_to_fixed_event_data = {
+    "---\n:reason: cannot_find_user\n" =>        "{\"reason\":\"cannot_find_user\"}",
+    "---\n:reason: bad_password\n" =>            "{\"reason\":\"bad_password\"}",
+    "---\n:reason: multiple_users\n" =>          "{\"reason\":\"multiple_users\"}",
+    "---\n:reason: too_many_login_attempts\n" => "{\"reason\":\"too_many_login_attempts\"}"
+  }
+
+  bad_to_fixed_event_data.each do |bad, fix|
+    ActiveRecord::Base.connection.execute(
+      "update security_logs set event_data='#{fix}' where event_data='#{bad}'"
```
Dantemss: No need to change but for future reference, you can write this as:

`SecurityLog.update_all("event_data='#{fix}' where event_data='#{bad}'")`

or even `SecurityLog.where(event_data: bad).update_all(event_data: fix)` but this option might do some conversions you don't want so it might not work (it might try to parse the bad YML in the JSON field)

Dantemss: And actually you want `SecurityLog.unscoped.where(etc)` or else the default_scope forces a subquery...

Dantemss: And actually either option seems to work fine for me.

jpslav: ah yes I had a problem with `update_all(event_data: fix)` escaping the `\n`s in the string; actually I guess this was more in the spec when I was trying to recreate the bad data.

PR #505, opened 2017-03-08
Hide signup when asked to log in from faculty application page

jpslav:
When a user visits the [faculty access application page](https://accounts.openstax.org/faculty_access/apply), they are expected to have an account and be logged in to it.  If for some reason they are not logged in, Accounts redirects them to the login page.  Before this PR, this log in page had a "Sign up" link, which doesn't really make sense because they should already have an account to get to the faculty application page.  Some users would click this link and get down a bad path (it would loop, etc).  

This is what it looked like before:

image

With this PR, we now hide the sign up link when users arrive from the faculty access application page:

image

### DIFF
--- a/app/controllers/faculty_access_controller.rb
+++ b/app/controllers/faculty_access_controller.rb
@@ -1,5 +1,7 @@
 class FacultyAccessController < ApplicationController
 
+  prepend_before_filter :disallow_signup, only: :apply
+
   helper_method :instructor_has_selected_subject
 
   def apply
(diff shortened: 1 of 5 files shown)


app/controllers/faculty_access_controller.rb
```diff
@@ -38,4 +40,9 @@ def instructor_has_selected_subject(key)
     params[:apply] && params[:apply][:subjects] && params[:apply][:subjects][key] == '1'
   end
 
+  def disallow_signup
+    # value doesn't really matter, if set at all, signup hidden
+    params[:no_signup] = 1
+  end
```
nathanstitt: I kinda feel like it's bad form to update params, since they should always correspond to what you see in the URL and it's a bit confusing when they don't match.  Would it be better to set a `@no_signup` flag, and check that in the view?

Maybe that's also set somewhere else though and you're just forcing it to be set here?

jpslav: Done because the `authenticate_user!` filter that runs after this does a redirect to the login page.  That before filter keeps this param around, and it will appear in the URL "/login?no_signup=1" so hopefully this isn't too bad, tho I agree it kind of appears on its own.

PR #506, opened 2017-03-20
Make security log search from user admin screen more specific

jpslav:
In the old way if we were looking at the admin user edit page for user with ID 9, the "Security Log" link would use a search query with `user:9` -- this would match all users who have a "9" in their username.  

This PR changes the link to just search by user ID.


jpslav:
@Dantemss reviewed and approved

PR #510, opened 2017-03-28
Fix exception when using a search string like `user:""`

Dantemss:
See rails/arel pull request 480.


jpslav:
@Dantemss possible to add a spec that this fixes?

Dantemss:
@jpslav added. For whatever reason only the search_security_logs one breaks if I remove the patch though.

PR #511, opened 2017-03-28
Handle existing emails in TransferOmniauthData

Dantemss:
Might fix some IllegalState exceptions... I hope

Well it at least adds some more specs

### DIFF
--- a/app/routines/transfer_omniauth_data.rb
+++ b/app/routines/transfer_omniauth_data.rb
@@ -25,9 +25,20 @@ class TransferOmniauthData
     end
 
     user.save
-    transfer_errors_from(user, {type: :verbatim}, true)
+    transfer_errors_from user, {type: :verbatim}, true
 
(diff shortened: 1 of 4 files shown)


app/routines/transfer_omniauth_data.rb
```diff
@@ -25,9 +25,20 @@ def exec(data, user)
...
+      unless existing_email.verified
+        existing_email.update_attribute :verified, true
+
+        # Ensure we get updated contact_infos if we try to use them
+        user.contact_infos.reset
+        user.email_addresses.reset
+      end
+    else
+      run AddEmailToUser, data.email, user, already_verified: true
+    end
```
jpslav: `AddEmailToUser` already does this verified logic stuff.  Is the difference the addition of the `reset` calls?

Dantemss:
@jpslav Removed the change and left only the `.resets` and the new specs

PR #514, opened 2017-04-12
Fixing up some random TODOs (WIP)

jpslav:
Some old WIP TODO fixing, making it a PR so I don't forget about it.


Dantemss:
When doing there changes, just keep in mind that I use the date on your most recent sign_in_successful and your most recent login date, for things that check that and require reauthentication. So you might have to add to either add the new security log types to what is checked or also create sign_in_successful logs.

PR #516, opened 2017-04-15
Let admins manually set SF contact IDs

jpslav:
Now admins can manually set SF contact IDs.  Before the new ID is saved on the User, the code checks to verify that the ID is that of a SF Contact.  If there is an existing ID that we want to remove, the admin can set the new ID as `remove` and this special string will cause the ID to be removed.  This code does not simultaneously update the faculty status -- we'll just have to wait < 20 mins for the next run of the SF update code to do that.  

image

A lot of the files are cassette reruns

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -22,7 +22,7 @@ module Admin
       was_administrator = @user.is_administrator
 
       respond_to do |format|
-        if change_user_password && add_email_to_user && update_user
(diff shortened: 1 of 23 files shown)


Dantemss:
Entering `remove` seems weird to me, otherwise looks good.
If you wanna keep `remove`, I suggest using `.downcase` to catch REMOVE and Remove too.

jpslav:
yeah @Dantemss I know it is weird - was just a quick thing instead of adding a "delete" button and separate action.

PR #517, opened 2017-04-17
Prohibit semicolons in email addresses

jpslav:
... if there's a semicolon it gets read as two different email addresses and then Amazon issues a "no domain" error for one of the two split addresses.


nathanstitt:
looks good, but spec needs updated

PR #521, opened 2017-05-23
Push email to SF Contact when marked faculty verified

jpslav:
Pushes an email address to the SF Contact when the user is marked faculty verified in Accounts so that SF can send they "yay you are verified" email.

Also adds a `guessed_preferred_email` method to `User` and to the user JSON -- is a heuristic way to figure out what email address out of many the user may prefer.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -84,31 +84,37 @@ module Admin
     def change_salesforce_contact
       new_id = params[:user][:salesforce_contact_id]
 
-      return true if new_id.blank? || new_id == @user.salesforce_contact_id
+      if new_id.blank? || new_id == @user.salesforce_contact_id
+        return true
+      end
 
-      new_id = nil if new_id.downcase == "remove"
+      if new_id.downcase == "remove"
+        flash[:notice] = "Removed the Salesforce Contact ID"
+        @user.salesforce_contact_id = nil
+        return @user.save
+      end
 
-      check_really_exists = new_id.present?
-
-      if check_really_exists && !OpenStax::Salesforce.ready_for_api_usage?
+      if !OpenStax::Salesforce.ready_for_api_usage?
         flash[:alert] = "Can't connect to Salesforce to verify changed contact ID"
         return false
       end
 
       begin
-        contact = OpenStax::Salesforce::Remote::Contact.find(new_id) if check_really_exists
+        contact = OpenStax::Salesforce::Remote::Contact.find(new_id)
 
-        if contact.present? || new_id.nil?
+        if contact.present?
+          # The contact really exists, so save its ID to the User
+          flash[:notice] = "Updated Salesforce Contact"
           @user.salesforce_contact_id = new_id
           return @user.save
         end
       rescue
         # exploded, probably due to badly formed SF ID
-      ensure
-        # either exploded or contact was `nil`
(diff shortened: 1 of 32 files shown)


config/initializers/openstax_salesforce.rb
```diff
@@ -24,8 +24,15 @@
...
     config.sandbox_instance_url = secrets['tutor_specs_instance_url']
+
+    # DO NOT set the refresh token, because if the oauth token has expired the
+    # specs will use the refresh token to get a new oauth token, but then recorded
+    # cassettes will contain that unfiltered token and future spec runs may end up
+    # having the "unused interactions" messages because they don't expect the refresh
+    # token interactions.
+    #
+    #config.sandbox_refresh_token = secrets['tutor_specs_refresh_token']
   end
```
nathanstitt: should we remove the commented out code?

app/models/user.rb
```diff
@@ -210,6 +210,23 @@ def self.non_student_known_roles
...
+                     .where{confirmation_sent_at != nil} # manually verified
+                     .order{created_at.asc}
+                     .first ||
+      email_addresses.verified
+                     .order{created_at.asc}
+                     .first
+    ).try(:value)
+  end
+
   protected
```
nathanstitt: I'm guessing that there's no real way to detect them, but it seems to me that the last manually entered address would be preferred.  I can see users maybe adding a new email address when they get one, and not removing the previous one.

app/controllers/admin/users_controller.rb
```diff
@@ -97,10 +97,15 @@ def change_salesforce_contact
...
+
+        if contact.present? || new_id.nil?
+          @user.salesforce_contact_id = new_id
+          return @user.save
+        end
       rescue
+        # exploded, probably due to badly formed SF ID
+      ensure
+        # either exploded or contact was `nil`
         flash[:alert] = "Can't find a Salesforce contact with ID #{new_id}"
```
nathanstitt: This block tricked me up. I first thought that since you're using `find` for the contact, it will always be present or will have raised a not found exception.  But then I realized I'd missed the `if check_really_exists` guard. 

Maybe it'd be clearer as:
```ruby
if check_really_exists && new_id.present?
  contact = OpenStax::Salesforce::Remote::Contact.find(id: new_id)
  @user.salesforce_contact_id = new_id
  return @user.save
end
rescue # catch NotFound or whatever
...
```

Also, the `ensure` block will always be executed, so it will populate the `flash` even if the `return @user.save` executed.  I think it'll need a `unless @user && @user.valid?` tacked onto the end.

config/initializers/openstax_salesforce.rb
jpslav: I'm leaving it in there so it isn't put back in (I want it to not be set by anyone)

app/controllers/admin/users_controller.rb
jpslav: I think I'm trying to do too many things in one block.  I do want to save a nil `new_id` value (that's how I'm letting users clear a contact ID).  The `find` is also a bit funky in that it is a SF finder, not an ActiveRecord one, so it gives a weird exception if the ID isn't of the proper format, it returns `nil` if the ID is of the proper format but actually not found.  I bet if I split it up I can make it more clear.

nathanstitt: Hah, I missed that it wasn't a active record finder - was expecting it to raise an exception if the find failed.

PR #533, opened 2017-07-19
Sped up Salesforce communication

Dantemss:
Most of the speed gains come from the new index.
The each_slice stuff seems to help a tiny bit, as long as the index is present.


Dantemss:
From my testing, seems about 5x faster. We could reduce the delay for sending info to Accounts, if desired. Maybe every 10 minutes would be a good number.

PR #534, opened 2017-07-28
Accounts login performance improvements

Dantemss:
- Changed Doorkeeper config so AccessTokens are reused
- Added rake task and cron to remove expired AccessGrants
- One-time cleanup of old AccessTokens

### DIFF
--- a/db/migrate/20170728203226_remove_old_access_tokens.rb
+++ b/db/migrate/20170728203226_remove_old_access_tokens.rb
@@ -1,0 +1,27 @@
+class RemoveOldAccessTokens < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    ActiveRecord::Base.transaction do
+      Rake::Task['doorkeeper:cleanup'].invoke
+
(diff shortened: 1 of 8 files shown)


db/migrate/20170728203226_remove_old_access_tokens.rb
```diff
@@ -0,0 +1,27 @@
...
+      Rake::Task['doorkeeper:cleanup'].invoke
+
+      Doorkeeper::AccessToken.where(
+        <<-SQL.strip_heredoc
+          EXISTS (
+            SELECT *
+              FROM "oauth_access_tokens" "oat"
+              WHERE "oat"."resource_owner_id" = "oauth_access_tokens"."resource_owner_id"
+                AND "oat"."application_id" = "oauth_access_tokens"."application_id"
+                AND "oat"."created_at" > "oauth_access_tokens"."created_at"
```
jpslav: I should probably learn more SQL :-) This looks like it is self-referential?

PR #535, opened 2017-08-03
Focus first form field on signup page

mwharrison:
From https://trello.com/c/GnqWNehB/2-bug-account-accessibility-issues
I just wanted to get Accounts up and running and do one PR. :) Feel free to tell me this isn't the correct way to do this.

### DIFF
--- a/app/assets/javascripts/signup/index.js
+++ b/app/assets/javascripts/signup/index.js
@@ -5,6 +5,7 @@
 //= require_self
 
 $(document).ready(function(){
+  $('form:first *:input[type!=hidden]:first').focus();
   OX.Signup.TypeSelector.initialize();
   $('[data-toggle="tooltip"]').tooltip()
 });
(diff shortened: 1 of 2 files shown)


jpslav:
Thanks @mwharrison!  Can you add to the happy path scenarios in the log in and sign up feature specs that the intended field gets the focus?  e.g. around here https://github.com/openstax/accounts/blob/master/spec/features/user_signs_up_spec.rb#L20 for sign up and probably similar for log in.  Might need something like https://stackoverflow.com/a/13177528 but with the `expect` syntax, e.g. `expect(page.evaluate_script("document.activeElement.id")).to eq "some_id"`

jpslav:
If you wanted to be fancy, you could even add a helper method for this ugly looking test to the `spec/feature_helpers.rb` file:

```ruby
def expect_focus_on_id(id)
  expect(page.evaluate_script("document.activeElement.id")).to eq id
end
```

mwharrison:
@jpslav - thanks for pointing me in the right place, I wasn't finding what was failing. :)

jpslav:
@mwharrison the failures are actually unrelated -- looks like we have some failures on master.

mwharrison:

PR #539, opened 2017-08-23
Auto-create account using trusted params from LMS

nathanstitt:
This accepts a set of signed parameters from an application and auto-signs in or drops a user into the sign-up flow with fields pre-populated


app/handlers/signup_start.rb
```diff
@@ -51,6 +56,27 @@ def handle
...
+      value: signup_state.contact_info_value
+    )
+    run(ConfirmContactInfo, ci)
+    transfer_errors_from(ci, {type: :verbatim}, true)
+    options[:session].sign_in! user
+
+    uuid = UserAlternativeUuid.create(user: user, uuid: trusted_state['uuid'])
+    transfer_errors_from(uuid, {type: :verbatim}, true)
+
+    outputs.redirect_action = :password
```
jpslav: Seems like maybe we want the calling app to indicate whether humans sent to accounts should be asked for a password or not... i.e. for LMS students Tutor can say "doesn't need a password", but for LMS teachers we would want a password so they can use their account for other OX uses.

app/handlers/trusted_parameters_launch.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  protected
+
+  def authorized?
+    true
+  end
+
+  def setup
+    @attrs = request.session['trusted']
+    @user_state = options[:user_state]
```
jpslav: this is totally fine here... but just fyi normally stuff in `setup` is that which you need to be available when `authorized?` is called.  Otherwise it can just go into `handle`.

app/handlers/trusted_parameters_launch.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    if attrs['email']
+       user = LookupUsers.by_verified_email(attrs['email']).first
+       if user
+         UserAlternativeUuid.create(user: user, uuid: attrs['uuid'])
+         user_state.sign_in!(user)
+         outputs[:user] = user
+         return
+       end
+    end
```
jpslav: I think we still need the part of the flow where a user can log in to an existing account, right (where the trusted email doesn't match an existing email)? (in which case we would use the trusted parameters to add to their existing account and we'd also create a `UserAlternativeUuid` for them.

app/controllers/sessions_controller.rb
```diff
@@ -67,6 +69,21 @@ def authenticate
     redirect_to root_path if signed_in?
   end
 
+  def trusted_launch
+    handle_with(TrustedParametersLaunch,
+                user_state: self,
```
jpslav: Pass in `trusted_state: session[:trusted]` here so that don't have to say `@attrs = request.session['trusted']` in the handler below.  Makes testing easier too -- don't have to create or mock an entire request.

app/models/user.rb
```diff
@@ -22,7 +22,7 @@ class User < ActiveRecord::Base
   has_many :message_recipients, inverse_of: :user, dependent: :destroy
   has_many :received_messages, through: :message_recipients, source: :message
   has_many :sent_messages, class_name: 'Message'
-
+  has_many :alternative_uuids, class_name: 'UserAlternativeUuid'
```
jpslav: `dependent: :destroy`

app/models/user_alternative_uuid.rb
```diff
@@ -0,0 +1,6 @@
+class UserAlternativeUuid < ActiveRecord::Base
```
jpslav: In another comment talked about maybe calling the trusted uuid an "external UUID" - if you like that, maybe change this to match ... `UserExternalUuid`

app/models/user_alternative_uuid.rb
```diff
@@ -0,0 +1,6 @@
...
+
+  belongs_to :user, inverse_of: :alternative_uuids
+
+  validates :uuid, :user, presence: true
```
jpslav: add uniqueness check, also in DB

db/migrate/20170824174144_create_user_alternative_uuids.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUserAlternativeUuids < ActiveRecord::Migration
+  def change
+    create_table :user_alternative_uuids do |t|
+      t.references :user, null: false
+      t.string :uuid, null: false, index: true
```
jpslav: add uniqueness constraint

app/handlers/signup_start.rb
```diff
@@ -51,6 +56,27 @@ def handle
     outputs.signup_state = new_signup_state
   end
 
+  def confirm_trusted_account(signup_state)
```
jpslav: I have a non-firm feeling that this code should be somewhere else :-) like on its own somewhere.... but it isn't clear in my brain.

Like this is somehow a "calling app said just make a user and don't get a password" routine or something, and maybe not called from `SignupStart`... dunno

app/models/user.rb
```diff
@@ -135,6 +135,12 @@ def full_name
     guess.blank? ? nil : guess
   end
 
+  def full_name=(name)
+    names = name.split(/\s+/)
```
jpslav: to be paranoid, you might `name.strip.split...`

lib/lookup_users.rb
```diff
@@ -21,21 +21,26 @@ def self.by_email_or_username(email_or_username)
...
     end
 
+    return self.by_verified_email(email_or_username)
+  end
+
+  def self.by_verified_email(email)
     # Case-sensitive email search
     ContactInfo.verified
-               .where(value: email_or_username)
+               .where(value: email)
```
jpslav: wow I'm guessing this code was not being used :-)

app/controllers/sessions_controller.rb
```diff
@@ -296,4 +313,22 @@ def maybe_skip_to_sign_up
...
 
+  def maybe_launch_with_trusted_params
+    return unless params[:go] == 'trusted_launch'
+
+    base_string = OAuth::Helper.normalize(
+      params.except(:controller, :action, :client_id, :signature)
+    )
+    secret_key = Doorkeeper::Application.find_by_uid!(params[:client_id]).secret
+    signature = OpenSSL::HMAC.hexdigest('sha1', secret_key, base_string)
+    if signature != params[:signature] ||
```
jpslav: being paranoid here... can `hexdigest` ever return `nil`?  If so we need a separate error for `params[:signature]` being `nil`

app/controllers/sessions_controller.rb
```diff
@@ -296,4 +313,22 @@ def maybe_skip_to_sign_up
...
+       !(2.minutes.ago..2.minutes.from_now).cover?(Time.at(params[:timestamp].to_i))
+
```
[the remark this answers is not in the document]
nathanstitt: Yes, what'd you think about moving this and the accompanying encoder from tutor's `lms_controller` to `openstax_api` ?

lib/user_session_management.rb
```diff
@@ -154,4 +163,14 @@ def get_alternate_signup_url
     session[:alt_signup]
   end
 
+  # called when the user has trusted paramters for creating or sign in
+  def set_trusted_parameters(params)
+    session[:trusted] = {
```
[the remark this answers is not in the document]
nathanstitt: sure :grin:

http://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html says:
> If you have secret_key_base set, your cookies will be encrypted. This goes a step further than signed cookies in that encrypted cookies cannot be altered or read by users. This is the default starting in Rails 4.

secret_key_base is set in `secrets.yml`  - so should be present unless deployment scripts isn't setting it for some reason.  I checked production and my `_accounts_session` cookie is encrypted

lib/user_session_management.rb
```diff
@@ -154,4 +163,14 @@ def get_alternate_signup_url
...
+      email: params[:email],
+      name:  params[:name],
+      uuid:  params[:uuid],
```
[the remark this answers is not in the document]
nathanstitt: it's the uuid that is sent from the LMS - would probably be better named `user_uuid` in the params, I'll rename it.

app/controllers/sessions_controller.rb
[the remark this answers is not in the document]
jpslav: Interesting! nice one

lib/user_session_management.rb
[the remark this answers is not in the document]
jpslav: great - I think I've checked this many times and will probably do it again in the future ;-)

lib/user_session_management.rb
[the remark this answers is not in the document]
jpslav: how about `external_user_uuid`?  `user_uuid` sounds like it is talking about the account's "user uuid"

app/handlers/signup_start.rb
nathanstitt: Yes, I've moved the trusted_data storage onto signup-state, and have added logic where needed to consult it, rather than creating users in multiple spots

app/handlers/signup_start.rb
nathanstitt: Actually i kinda wonder if we want to give students a password.  We'd decided not to, but that was when we planned to just auto-create an account for them.  Maybe it makes sense to do so now?

lib/lookup_users.rb
nathanstitt: Oh that was because I added a new `by_verified_email(email)` to  lookup_users

nathanstitt:
Whew, thanks for an excellent review @jpslav 

I've changed the code around quite a bit, it now stores the trusted params on the `signup_state` record and the various controller actions/handlers consult it as needed.

That slims down the session and also gets rid of the ugly `setup_trusted..` stuff in the start handler.

app/controllers/signup_controller.rb
```diff
@@ -15,7 +15,7 @@ class SignupController < ApplicationController
   before_filter :restart_if_missing_signup_state, only: [:verify_email, :password, :social]
 
   # TODO spec this
-  before_filter :exit_signup_if_logged_in, only: [:start, :verify_email, :password, :social, :verify_by_token]
+  before_filter :exit_signup_if_logged_in, only: [:start, :social, :verify_by_token]
```
jpslav: I think we still need to exit signup if logged in for `:verify_email` and `:password`.  Because of the `# TODO spec this` comment nothing may have broken in the specs but I think we still need it.

app/handlers/signup_password.rb
```diff
@@ -20,7 +20,11 @@ def authorized?
   end
 
   def handle
-    outputs.user = User.create
+    outputs.user = User.build
+    if signup_state.trusted?
+      outputs.user.full_name = signup_state.trusted_data['name']
+    end
```
jpslav: I think we still need to send the user through the profile screen (so they can change their name from "Bob Jones" to "Professor Bob Jones".  If so, can we do the name there? Prepopulate the name field on that form?

app/handlers/signup_start.rb
```diff
@@ -19,30 +19,29 @@ def handle
...
     end
 
     # If email in use, want users to login with that email, not create another account
     fatal_error(code: :email_in_use, offending_inputs: [:signup, :email]) if email_in_use?
 
+    # Create a new one
+    new_signup_state = SignupState.email_address.create(
+      contact_info_value: email,
+      role: signup_params.role,
+      trusted_data: existing_signup_state.trusted_data,
```
jpslav: cool, I like it.

let's not forget too to use the trusted data to improve someone's account (e.g. add verified email if it doesn't exist, help mark faculty verified -- not sure how to do latter yet)

app/models/signup_state.rb
```diff
@@ -18,9 +18,38 @@ class SignupState < ActiveRecord::Base
...
+      }
+    )
+  end
+
+  def trusted?
+    trusted_data.present?
+  end
+
+  def skip_email_validation?
+    contact_info_value == trusted_data['email']
```
jpslav: Can we use existing `signup_state.verified?` in place of `skip_email_validation?`

app/handlers/signup_start.rb
nathanstitt: Yes, when `signup_state` is initialized from trusted data it's marked as verified, which in turn marks the contact as verified when it's created.

https://github.com/openstax/accounts/pull/539/files/de2966022584438e9a8769bd4bb7dcfa18dc3162..a48de9bfe35eb8672d686157b9f6d0ba7608fa75#diff-6f4447b02666c80e38966472bdc56145R25

app/models/signup_state.rb
nathanstitt: Actually what we need is a method to determine what should be rendered after the first role / email entry screen.  For normal users, it's email verification.  For verified instructors it's password, and for verified students it's direct to profile.  

Another wrinkle is that accounts are created when the password is, so for verified students we have to create the account first.

I added a `trusted_student` controller action/handler to accomplish this.

app/controllers/signup_controller.rb
nathanstitt: Looks to me like we do have a spec for the exit_signup case 
https://github.com/openstax/accounts/blob/master/spec/features/user_signs_up_spec.rb#L363-L371

I've removed TODOs

app/routines/user_from_signup_state.rb
```diff
@@ -0,0 +1,22 @@
...
+      user.full_name = signup_state.trusted_data['name']
+      user.self_reported_school = signup_state.trusted_data['school']
+      if signup_state.trusted_data['uuid']
+        user.external_uuids.build(uuid: signup_state.trusted_data['uuid'])
+      end
+    end
+    user.save
+    outputs.user = user
+    transfer_errors_from(user, {type: :verbatim}, true)
+  end
```
nathanstitt: @jpslav I wonder if we should add a json field to user and store the trusted data there so it's retained?  As we'd discussed we're setting the `self_reported_school` to something that's not really self_reported, instead it's trusted.  

Down the road it might be nice if we tell apart what was trusted and self-reported,

app/handlers/sessions_trusted_launch.rb
```diff
@@ -0,0 +1,32 @@
...
+  def authorized?
+    true
+  end
+
+  def handle
+    signup_state = options[:signup_state]
+    uuid_link = UserExternalUuid.find_by_uuid(signup_state.trusted_data['uuid'])
+    if uuid_link.present?
+      outputs[:user] = uuid_link.user
+      signup_state.sign_in!(outputs[:user])
```
jpslav: It is late, so I may be off my rocker .... but `signup_state` is an instance of `SignupState`, right?  It doesn't have a `sign_in!` method that I see?

app/handlers/sessions_trusted_launch.rb
```diff
@@ -0,0 +1,32 @@
...
+    if signup_state.trusted_data['email']
+       user = LookupUsers.by_verified_email(signup_state.trusted_data['email']).first
+       if user
+         UserAlternativeUuid.create(user: user, uuid: trusted_state['uuid'])
+         signup_state.sign_in!(user)
+         outputs[:user] = user
+         return
+       end
+    end
+
```
jpslav: See my other comment on moving code out of `maybe_launch_with_trusted_params` and into this handler...

Here at the bottom, after we have failed to log in an existing user, we could make and store the `SignupState`.  We could also use this spot to encode in the `SignupState` whether we want the user to be asked to choose a password, e.g.

```ruby
outputs.signup_state = 
  if params[:sp][:role] == :instructor
    # Make the SignupState have `require_password` set to true (really `require_authentication` b/c could be social)
  else
    # set `require_password` to false
  end
```
Basically I'm looking for a way to move `after_email_action` out of `SignupState` because that method seems out of the scope of that model (models shouldn't know names of controller actions, but ok in my mind for a model to encode what is logically needed during sign up -- e.g. whether a password is needed or not).

The `handle_with`'s `success` block already takes action based on if `outputs.user` is set.  Similarly, it can use the `outputs.signup_state`'s present or missing email address and `require_authentication` fields to determine where to send the user next (i.e. to collect and verify an email or not, then to collect a password/social auth or not).

With more code in this handler, it will become more of a one-stop shop to see what we do with incoming known users... see if we recognized their external UUID, if so log them in; see if we recognize their external email address, if so log them in; otherwise, send them along the login/signup flow with options set on what we need to collect and what we can skip.

app/controllers/signup_controller.rb
```diff
@@ -3,20 +3,16 @@ class SignupController < ApplicationController
...
 
   before_filter :check_ready_for_profile, only: [:profile]
 
-  # TODO spec this
   before_filter :restart_if_missing_signup_state, only: [:verify_email, :password, :social]
-
-  # TODO spec this
-  before_filter :exit_signup_if_logged_in, only: [:start, :verify_email, :password, :social, :verify_by_token]
-
+  before_filter :exit_signup_if_logged_in, only: [:start, :social, :verify_by_token]
```
jpslav: I'm still concerned that `:verify_email`, `:password`, and `:social` are no longer in this `only` list... sorry if you explained this in a prior review response... do we know these are safe to remove?

app/controllers/sessions_controller.rb
```diff
@@ -67,6 +68,22 @@ def authenticate
...
   end
 
+  def trusted_launch
+    handle_with(SessionsTrustedLaunch,
+                signup_state: signup_state,
+                success: lambda do
+                  if @handler_result.outputs.user
+                    redirect_back
+                  else
+                    redirect_to signup_url
```
jpslav: When we redirect a user to signup, how do we handle the case where we already have an email address for the user?  (what happens on the first start screen which currently requests a role (which we have) and an email address (which we may have)).  Seems like here we'd maybe need to skip `signup#start` or use a variant of it?  

See related comment later that talks about more decisions about where to send different states of users in this `success` block.

app/controllers/signup_controller.rb
```diff
@@ -25,10 +21,12 @@ def start
     if request.post?
```
jpslav: I suspect the `GET` of `start` needs to be able to handle if we already know user role or user email?

app/controllers/signup_controller.rb
```diff
@@ -25,10 +21,12 @@ def start
...
       handle_with(SignupStart,
                   existing_signup_state: signup_state,
+                  trusted_state: session[:trusted],
                   return_to: session[:return_to],
+                  session: self,
                   success: lambda do
                     save_signup_state(@handler_result.outputs.signup_state)
-                    redirect_to action: :verify_email
+                    redirect_to action: signup_state.after_email_action
```
jpslav: See other comment about making the choice of redirection based on more simple state in `SignupState` and not having `SignupState` know about action names.

app/controllers/signup_controller.rb
```diff
@@ -71,6 +69,18 @@ def verify_by_token
...
+    handle_with(SignupTrustedStudent,
+                signup_state: signup_state,
+                session: self,
+                success: lambda do
+                  redirect_to action: :profile
+                end,
+                failure: lambda do
+                  redirect_to action: :start
+                end)
+  end
```
jpslav: Instead of a dedicated `trusted_student` action and handler, can we just use the existing flow and look to the signup state to know whether we can skip email and password solicitation?

app/handlers/signup_trusted_student.rb
```diff
@@ -0,0 +1,26 @@
...
+    signup_state.trusted?
+  end
+
+  def handle
+    run(UserFromSignupState, signup_state)
+    contact_info = ContactInfo.new(type: 'EmailAddress', value: signup_state.contact_info_value)
+    contact_info.verified = true
+    contact_info.user = outputs.user
+    contact_info.save
+    transfer_errors_from(contact_info, {scope: :contact_info}, true)
```
jpslav: If we keep this handler, these lines can collapse to 

```ruby
run(AddEmailToUser, signup_state.contact_info_value, outputs.user, {already_verified: true})
```

app/models/signup_state.rb
```diff
@@ -18,6 +18,30 @@ class SignupState < ActiveRecord::Base
...
+  end
+
+  def after_email_action
+    return :verify_email unless trusted?
+    if role == 'instructor'
+      :password
+    else
+      :trusted_student
+    end
+  end
```
jpslav: would like to remove this method (see other comments)

app/routines/user_from_signup_state.rb
jpslav: I like that idea.

app/controllers/sessions_controller.rb
```diff
@@ -67,6 +68,22 @@ def authenticate
...
+                  end
+                end,
+                failure: lambda do
+                  flash[:alert] = I18n.t :"controllers.sessions.trusted_launch_failed"
+                  render :start
+                end)
```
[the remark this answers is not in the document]
nathanstitt: Yes my use of the term was for "start the signup process with known data"

After thinking a bit more, I think we can forgo the entire prefix and just call it  `SessionsTrustedSignup` and likewise `maybe_signup_with_trusted_params` below.

app/controllers/sessions_controller.rb
nathanstitt: That case should not happen, because we auto-log them in if there's a matching verified email 

https://github.com/openstax/accounts/pull/539/files#diff-4d6adcaa9d66afacdff074ac1a555244R21

jpslav: Sorry, by "we already have an email address for the user" I meant an email address that doesn't match an existing verified email.  E.g. someone has an account with "bob@gmail.com" but we get "bob@school.edu" in the trusted data.

app/controllers/sessions_controller.rb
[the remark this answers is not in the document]
jpslav: like the direction... signup is still a bit not right b/c we may just log the user in, not sign them up.  `new_with_trusted_data`... or we ditch this action altogether and the `maybe` filter and put this code into the `new` action?

```ruby
def new
  if params[:sp].present?
    # act on it
  else
    render
  end
end
```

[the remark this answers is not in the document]
jpslav: Actually everything could go into a `SessionsNew` handler.  That might make some decent sense.  Some people with trusted data are going to need to see the first login screen anyway.

app/controllers/sessions_controller.rb
```diff
@@ -296,4 +313,20 @@ def maybe_skip_to_sign_up
...
+             params: params[:sp],
+             secret: ::Doorkeeper::Application.find_by_uid!(params[:client_id]).secret)
+
+      Rails.logger.warn "Invalid signature for trusted parameters"
+      head :forbidden and return false
+    end
+
+    save_signup_state(SignupState.create_from_trusted_data(params[:sp]))
+    redirect_to action: :trusted_launch
+  end
```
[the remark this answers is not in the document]
nathanstitt: yep, that's exactly what turned out to be the case.  After I added the "SignupStart" handler it's able to immediately sign a user in and redirect back.  That'll be nice for the happy path where a user's returning

app/handlers/sessions_trusted_launch.rb
nathanstitt: yeah - that's a bug :(  Should be calling the method on the controller reference.  All OOB now of course.

app/controllers/signup_controller.rb
nathanstitt: I think we still have to have signup_state (or something else) decide what the "after_email_action" should be.

The user may have changed the email in the signup form, in which case we can no longer trust it and it needs to be verified.

Or the user may change the role on signup, in which case we do need to allow them to fill out password

app/controllers/signup_controller.rb
nathanstitt: I don't recall why I removed these..  And I thought these were specced, but It now appears they are not since the spec didn't fail.   I'll add add them back and add a spec for 'em.

app/controllers/signup_controller.rb
nathanstitt: no, because the regular flow creates the user account after a password is added (in https://github.com/openstax/accounts/blob/master/lib/omniauth/strategies/custom_identity.rb#L131-L134)

Since student's don't get a password that's never called.  I extracted the "create user from signup" into a routine and am calling it from both handlers.

app/controllers/signup_controller.rb
nathanstitt: i've moved the logic for this to the SignupStart handler

app/handlers/sessions_trusted_launch.rb
```diff
@@ -0,0 +1,32 @@
...
+  def authorized?
+    true
+  end
+
+  def handle
+    signup_state = options[:signup_state]
+    uuid_link = UserExternalUuid.find_by_uuid(signup_state.trusted_data['uuid'])
+    if uuid_link.present?
+      outputs[:user] = uuid_link.user
```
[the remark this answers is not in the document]
nathanstitt: I don't know for sure but it seems like it might be a bad idea to change user's info after they've signed up.  I'm envisioning a case where an institution has bad info for a user in the LMS (misspelled name or something).  

The user might not be able to edit their records there to correct it, but they would fix it in Tutor, only to have their changes overwritten.

Of course I have no idea how likely that is, but seems like it at least possible.

app/controllers/signup_controller.rb
nathanstitt: I've changed my mind on this, and have extracted some of the logic into a `SignupTrustedStudent` routine that I'm calling from the `SignupStart` handler.  That seems to work well, but I'm not super happy with how large `SignupStart` is now 😁

app/handlers/sessions_start.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    user = nil
+
+    uuid_link = UserExternalUuid.find_by_uuid(secure_params['uuid'])
+    if uuid_link.present?
+      user = uuid_link.user
+    elsif secure_params['email'] # maybe there's a verified email that matches?
+      user = LookupUsers.by_verified_email(secure_params['email']).first
+      if user && secure_params['uuid']
+        UserAlternativeUuid.create!(user: user, uuid: secure_params['uuid'])
```
jpslav: missed changing this to new `UserExternalUuid`

db/migrate/20170824174144_create_user_alternative_uuids.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUserExternalUuids < ActiveRecord::Migration
```
jpslav: change file name to make class name

spec/factories/user_alternative_uuids.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :user_alternative_uuid do
```
jpslav: user_external_uuid

mwharrison:

app/controllers/signup_controller.rb
```diff
@@ -88,17 +85,20 @@ def profile
...
                   success: lambda do
-                    clear_signup_state
-
-                    if current_user.student?
+                    if current_user.student? || current_user.created_from_trusted_data?
                       redirect_back
                     else
                       redirect_to action: :instructor_access_pending
                     end
+                    clear_signup_state
```
jpslav: why does it matter if `clear_signup_state` is called here or at the start of the block?  Will happen either way, and the redirects don't depend on it?

lib/user_session_management.rb
```diff
@@ -58,7 +64,6 @@ def authenticate_admin!
   alias_method :admin_authentication!, :authenticate_admin!
 
   def set_login_state(username_or_email: nil, matching_user_ids: nil, names: nil, providers: nil)
-    clear_signup_state
```
jpslav: we now don't delete this because the `signup_state` is now holding the signed data?

nathanstitt: yes, and it's needed if the user switches from sign-in to sign-up by clicking the link at bottom of sign-in modal.

Previously this was cleared here and then the sign-up wouldn't be trusted

app/controllers/signup_controller.rb
nathanstitt: I'm guessing a previous incarnation of the code read `signup_state` when deciding where redirect.  I'll move it back

app/controllers/sessions_controller.rb
```diff
@@ -8,28 +8,46 @@ class SessionsController < ApplicationController
...
 
   # Login form
-  def new; end
+  def start
+    handle_with(SessionsStart,
+                signup_state: signup_state,
+                session_management: self,
+                success: lambda do
+                  if @handler_result.outputs.user
+                    redirect_back
```
jpslav: I think this needs to be `redirect_to action: :redirect_back` so that the `before_filter` that checks for terms signature is run.  Can you add a feature spec where a LMS user logs in but the terms have changed (so that they have unsigned terms) and the spec shows that they have to sign terms before getting sent back to the app?  See `make_new_contract_version` in `spec/feature_helpers.rb`.  I think the general case spec is missing for this too, which is my bad.

Gemfile
```diff
@@ -48,7 +48,7 @@ gem 'quiet_assets'
 gem 'openstax_utilities', '~> 4.2.0'
 
 # API versioning and documentation
-gem 'openstax_api', '~> 8.1.0'
+gem 'openstax_api', git: 'https://github.com/openstax/openstax_api', branch: 'sign-params'
```
jpslav: I released `openstax_api` `8.2.0` - can you update the gemfile?

db/migrate/20170824174144_create_user_alternative_uuids.rb
jpslav: @nathanstitt please change the filename to talk about external uuids, not alternative uuids (still reflects old model name)

spec/factories/user_alternative_uuids.rb
jpslav: ditto on this having old name

spec/features/trusted_login_flow_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  background do
+    load 'db/seeds.rb'
+    create_default_application
+  end
+  let(:role) { 'instructor' }
+  let(:payload) {
+    {
+      role:  role,
+      external_user_uuid: SecureRandom.uuid,
+      name:  'Tester McTesterson',
```
jpslav: hehe

spec/features/trusted_login_flow_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      email: 'test@test.com',
+      school: 'Testing U'
+    }
+  }
+  let(:signed_params) {
+    { sp: OpenStax::Api::Params.sign(params: payload, secret: @app.secret) }
+  }
+
+  describe 'instructors' do
```
jpslav: I think we need a spec to show a returning LMS user gets logged in immediately? (probably don't need instructor and student variants)

spec/features/trusted_login_flow_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      expect_back_at_app
+      expect_validated_records(params: payload.merge(email: email))
+    end
+
+  end
+
+  describe 'students' do
+    let(:role) { 'student' }
+
+    it 'signs in and links' do
```
jpslav: think this title is wrong - this is sign up, probably just copy/paste from above

app/models/user_external_uuid.rb
```diff
@@ -0,0 +1,6 @@
+class UserExternalUuid < ActiveRecord::Base
+
+  belongs_to :user, inverse_of: :external_uuids
+
+  validates :uuid, :user, presence: true
```
jpslav: uniqueness too? (know it is also in db constraint)

jpslav: random separate question - in Tutor should we prepend the tool guid so that we can be more certain that the user "uuid" is actually unique?

app/controllers/sessions_controller.rb
nathanstitt: I added the spec as you suggested (good call, doing so uncovered a different bug).

However it still didn't prompt for terms - when I went to track down why I discovered that the `redirect_back` action is set to skip fine_print https://github.com/openstax/accounts/blob/master/app/controllers/sessions_controller.rb#L29

That's contrary to the comment before the action though.  Is the `fine_print_skip` in error?

spec/features/trusted_login_flow_spec.rb
nathanstitt: yeah probably copy & paste, or maybe I meant that the user is signed in at the end.  Either way it's not well named 😁

spec/features/trusted_login_flow_spec.rb
nathanstitt: yep, added that and one for the terms still showing up

PR #541, opened 2017-09-21
Added Google Analytics for production only

edwoodward:


app/views/layouts/_google_analytics.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if Rails.env == "production"  %>
```
jpslav: this syntax totally fine, but also common to see it as `if Rails.env.production?` (and similarly `if Rails.env.development?` and `if Rails.env.testing?`)

jpslav: On the Rails sites, all deployed environments use the "production" settings, so `Rails.env.production?` is true for `tutor-dev`, `tutor-qa`, and `tutor-staging` in addition to the "real" production at tutor.openstax.org.  

So there are a few options here.  

(1) You could make a helper function like [this one](https://github.com/openstax/tutor-server/blob/master/app/helpers/application_helper.rb#L13-L15) that lets you say `if is_real_production_site?`.  The disadvantage here is that this code won't run at all until it is on real production, but that probably isn't the end of the world since it is just copy and pasted code from Google.

(2) You could put the GA codes in vaulted secrets and then only set them for production (and write the code to not insert the snippet if the vault has blank codes

(3) You can put the GA tracking codes in the admin settings and load them from there.  Obviously this is easier to change vs in the vault where changes require deploys.  Here is [code to support multiple GA tracking codes](https://github.com/openstax/tutor-server/pull/1410/commits/d9af36711fd52f84ecc1ee5b24137436922d9515) then a [change to get tracking codes from the admin database settings instead of the vault](https://github.com/openstax/tutor-server/commit/bcfe51eb83060d9ded3d27b4e86cbb8fff2437e6#diff-149ca0692b5819eaa241f19f9ae8e79a)

You might check with marketing to see if we need to support multiple tracking codes like we are doing in Tutor.  The multiple codes are to support a new GA property in addition to a legacy property, I believe.

edwoodward: I created a GA flag in settings so it is testable in QA. By default everything is turned off except production.

app/views/layouts/_google_analytics.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if Settings::GoogleAnalytics.send_google_analytics == true  %>
```
jpslav: Now that you have the GA code in settings, this could be `if !Settings::GoogleAnalytics.google_analytics_code.blank?` -- that way you don't get into a situation where this flag is true, but the code is blank (leading to JS errors, etc).

edwoodward: The toggle is to allow QA to test easily, so I need to check this value. Otherwise they will have to know what the GA code is and enter it. I can add an "and !Settings::GoogleAnalytics.google_analytics_code.blank?" to the if statement

PR #545, opened 2017-10-06
Specs were missing 'spec' suffix so were not in full runs

jpslav:
Some files were missing `_spec` at the end of the filename so were not getting included in full rspec runs.  And one of these had some issues that needed to be fixed.


spec/features/log_out_inactive_admins_spec.rb
```diff
@@ -25,7 +25,6 @@
         Timecop.travel(login_time + 29.minutes)
         visit admin_feature_url
 
-        expect(page).to have_http_status(:success)
```
jpslav: pages were getting 304 Not Modified, but not getting sent to login (which this spec is worried about), so removed this check and the one below

PR #548, opened 2017-10-09
Handle LMS launches when user is already signed in

jpslav:
I started testing LMS launches on QA and happened to already be logged in as a user on Accounts.  I expected to be taken to log in or sign up pages (regardless of my log in state), but I wasn't -- I was just swooshed back to Tutor as my logged-in user and no external UUIDs, etc, were tracked.

Basically the problem is that we had code to do stuff in the `SessionsController`, but we were never even getting there.  `authenticate_user!` calls in before filters were catching us before then and sending us on our merry way back to Tutor.  

So this PR basically moves code that was in `SessionsStart` to `authenticate_user!`.  In addition to that move, it makes the following changes:

* If there is a user who is logged in during a signed parameter "launch", we log that user out if (1) the signed params don't point us to a user to log in or (2) if they do point us to a user to log in and it differs from the currently logged in user
* We add security log information so we can track who got logged in because of cases (1) and (2) above and so we know if signed parameter users are getting logged in via external UUID match or external email match.  
* Somewhat unrelatedly, I started skipping `authenticate_user!` on `SessionsController#redirect_back` calls, as I think we pretty much always have a signed-in user on those calls already, and even if we don't, redirecting back doesn't necessarily imply that we must have a signed-in user.
* The decision point for sending signed parameter students to sign up instead of log in also moved to `authenticate_user!`
* I moved the verification of the signed parameters up to the earliest `before_filter` on every request.  This helped me not worry in other places that the parameters were in fact valid.  I can know that they are valid because an early check would have 400'd if they weren't.
* Added ability to log extra data for `:sign_out` security log events.
* Once signed parameters are "used", we remove them from the query strings so that they aren't used again.  Means we can't rely on `params[:sp]` any other time -- need to get stuff from places like the `SignupState`.
* Now that `authenticate_user!` has grown in this PR, I no longer really makes sense in `UserSessionManagement`.  With its redirect calls and url helpers it was kind of out of place even before this PR.  So I moved it and `authenticate_admin!` into their own module.  Might be helpful to look at the [first commit's](https://github.com/openstax/accounts/pull/548/commits/c92a987576c45874a8c147eecdd57f33cf26eabf#diff-faeb6794c1e97d4ca64baa4753ce06e6L45) `authenticate_admin!` code to see what changed substantively before the move.


lib/authenticate_methods.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    if incoming_user.present?
+      if signed_in? && incoming_user != current_user
+        # Sign out the current user; signing in the new user will effectively sign out
+        # the current user, but we choose to sign out explicitly so we can record
+        # information in the current user's security log
+
+        sign_out!(security_log_data: {type: 'different external user'})
+      end
```
nathanstitt: I think there's a small loophole here.

* If the user is signed out, we search for a matching validated email and use that if found.
* However if they're signed in, we force them to go through the sign in/up flow, regardless of email matching.

I think we also need to check the validated emails of a signed-in user and link/redirect back in that case also.

lib/authenticate_methods.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  def external_email
+    signed_params['email']
+  end
+
+  def request_url_without_signed_params
+    url = request.url || ""
+    query_hash = Rack::Utils.parse_nested_query(URI.parse(url).query).except("sp")
+    "#{url.split('?').first}#{'?' + query_hash.to_query if query_hash.present?}"
+  end
```
nathanstitt: https://stackoverflow.com/a/4272050/7000032 has a really neat way to remove query params from a url.  Uses the addressable gem but we're already pulling that in.

jpslav: Cool.  I will change!

lib/authenticate_methods.rb
jpslav: discussed with @nathanstitt and we think this is ok.

PR #553, opened 2017-10-24
Add UUID to user search

mwharrison:


mwharrison:
[Add Student UUID to their Accounts profile information](https://trello.com/c/OPTNjcdU/468-add-student-uuid-to-their-accounts-profile-information)

app/views/admin/users/index.html.erb
```diff
@@ -100,10 +100,14 @@
...
               <i>State:</i> <%= user.state %>
-            </span>
+            </span> |
+
+            <span class="uuid">
+              <i>UUID:</i> <%= user.uuid || '(no uuid)' %>
+            </span> |
 
             <% if user.self_reported_school.present? %>
-              | <span class="school">
```
jpslav: The `|` still needs to be here and not after UUID in case there's no `self_reported_school`

mwharrison: That makes sense. Putting it back there. :)

PR #554, opened 2017-10-24
Require email verification even if using trusted signup

nathanstitt:
Also require sig-in before auto-linking account

### DIFF
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -56,20 +56,6 @@ class SignupState < ActiveRecord::Base
     UserExternalUuid.find_by_uuid(trusted_external_uuid)
   end
 
-  def on_user_authentication(user)
-    if trusted_external_uuid
-      existing_link = linked_external_uuid
-      # Update uuid to point to new account if it's linked to a different one
-      if existing_link
-        if existing_link.user_id != user.id
-          existing_link.update_attributes!(user: user)
-        end
-      else
-        UserExternalUuid.create!(user: user, uuid: trusted_external_uuid)
-      end
-    end
-  end
-
   protected
 
   def prepare
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -45,10 +45,6 @@ class SignupState < ActiveRecord::Base
     role_trusted? && role == 'instructor'
   end
 
-  def trusted_email?
-    trusted? && contact_info_value == trusted_data['email']
-  end
-
   def trusted_external_uuid
     trusted? ? trusted_data['external_user_uuid'] : nil
   end
@@ -58,12 +54,6 @@ class SignupState < ActiveRecord::Base
 
   def linked_external_uuid
     UserExternalUuid.find_by_uuid(trusted_external_uuid)
-  end
-
-  def suggested_username
-    if trusted_email? && LookupUsers.by_verified_email(trusted_data['email']).any?
-      trusted_data['email']
-    end
   end
 
   def on_user_authentication(user)
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -20,6 +20,7 @@ class SignupState < ActiveRecord::Base
 
   def self.create_from_trusted_data(data)
     role = User.roles[data[:role]] ? data['role'] : nil
+    data['external_user_uuid'] = data.delete('uuid')
     SignupState.create!(
       role: role,
       verified: false,
@@ -49,7 +50,7 @@ class SignupState < ActiveRecord::Base
   end
 
   def trusted_external_uuid
-    trusted? ? trusted_data['uuid'] : nil
+    trusted? ? trusted_data['external_user_uuid'] : nil
   end
 
   def confirmed;  verified;  end
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -49,11 +49,35 @@ class SignupState < ActiveRecord::Base
   end
 
   def trusted_external_uuid
-    trusted? ? trusted_data['external_user_uuid'] : nil
+    trusted? ? trusted_data['uuid'] : nil
   end
 
   def confirmed;  verified;  end
   def confirmed?; verified?; end
+
+  def linked_external_uuid
+    UserExternalUuid.find_by_uuid(trusted_external_uuid)
+  end
+
+  def suggested_username
+    if trusted_email? && LookupUsers.by_verified_email(trusted_data['email']).any?
+      trusted_data['email']
+    end
+  end
+
+  def on_user_authentication(user)
+    if trusted_external_uuid
+      existing_link = linked_external_uuid
+      # Update uuid to point to new account if it's linked to a different one
+      if existing_link
+        if existing_link.user_id != user.id
+          existing_link.update_attributes!(user: user)
+        end
+      else
+        UserExternalUuid.create!(user: user, uuid: trusted_external_uuid)
+      end
+    end
+  end
 
   protected
 
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -22,7 +22,7 @@ class SignupState < ActiveRecord::Base
     role = User.roles[data[:role]] ? data['role'] : nil
     SignupState.create!(
       role: role,
-      verified: true,
+      verified: false,
       contact_info_value: data['email'],
       trusted_data: data.merge(role: role)
     )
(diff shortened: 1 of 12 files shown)


lib/authenticate_methods.rb
```diff
@@ -38,66 +38,41 @@ def use_signed_params
...
-      incoming_user = LookupUsers.by_verified_email(external_email).first
-
       if incoming_user.present?
-        found_incoming_user_by = "external email"
-
-        if external_user_uuid.present?
-          # Go ahead and track the external UUID so can use in future
-          incoming_user.external_uuids.create!(uuid: external_user_uuid)
-        end
+        sign_out!(security_log_data: {type: 'new external user'}) if signed_in?
```
jpslav: This is not necessarily a "new external user" because the incoming user could be the same as the currently-logged-in user... so I think we still need the `if signed_in? && incoming_user != current_user` logic and if that triggers we sign out for a `different external user` reason.  (if it is the same user logging in again, we just don't do the `sign_out!` but we do repeat the `sign_in!` so that a security log is entered.)

lib/authenticate_methods.rb
```diff
@@ -38,66 +38,41 @@ def use_signed_params
     incoming_user = nil
     found_incoming_user_by = nil
```
jpslav: Lines 38 and 39 not needed.

lib/authenticate_methods.rb
```diff
@@ -38,66 +38,41 @@ def use_signed_params
...
+    sign_out!(security_log_data: {type: 'new external user'}) if signed_in?
 
-      signup_state = SignupState.create_from_trusted_data(params[:sp])
-      save_signup_state(signup_state)
-    end
+    # Save the secure params data to facilitate sign up if that's what the user
+    # chooses to do
+    signup_state = SignupState.create_from_trusted_data(params[:sp])
+    save_signup_state(signup_state)
   end
```
jpslav: The `use_signed_params` method has become two unconnected pieces of code (before line 52 and after).  Might be clearer if it became

```ruby
def use_signed_params
  auto_login_external_user || prepare_for_new_external_user_signup
end

def auto_login_external_user
  return false if external_user_uuid.blank?

  incoming_user = UserExternalUuid.find_by_uuid(external_user_uuid).try(:user)
  return false if incoming_user.nil?

  # other lines

  true
end

def prepare_for_new_external_user_signup
  # lines from bottom of current `use_signed_params` method
end
```

lib/omniauth/strategies/custom_identity.rb
```diff
@@ -69,6 +69,8 @@ def callback_phase
         return fail_with_log!(:too_many_login_attempts) if too_many_login_attempts?
 
         if identity.present?
+          # Give signup_state a chance to record the link if one exists
+          signup_state.on_user_authentication(identity.user) if signup_state
```
jpslav: Can you talk to me about why this is here?  I didn't expect to see this here.

app/views/sessions/start.html.erb
```diff
@@ -35,6 +35,7 @@
 
       <div class="card-body">
         <%= fh.text_field name: :username_or_email,
+            value: signup_state ? signup_state.suggested_username : '',
```
jpslav: does specifying this value break the form if a user enters a different email but there's an error with it? e.g. the LMS tells us about the email `bob@example.com`, but they think, no my email is `bobby@example.com` and use that, but then we don't find it, we'll clear `bobby@example.com` and replace it with `bob@example.com`.  Can we initialize if not set?

app/models/signup_state.rb
```diff
@@ -49,12 +49,36 @@ def trusted_email?
...
   def confirmed?; verified?; end
 
+  def linked_external_uuid
+    UserExternalUuid.find_by_uuid(trusted_external_uuid)
+  end
+
+  def suggested_username
+    if trusted_email? && LookupUsers.by_verified_email(trusted_data['email']).any?
+      trusted_data['email']
+    end
```
jpslav: This and the `on_user_authentication` code steer us away from skinny models.  So that this model doesn't have to know about the `LookupUsers` library method, can this code go elsewhere?  The view or somewhere.  A helper maybe.

app/models/signup_state.rb
```diff
@@ -49,12 +49,36 @@ def trusted_email?
...
+    if trusted_external_uuid
+      existing_link = linked_external_uuid
+      # Update uuid to point to new account if it's linked to a different one
+      if existing_link
+        if existing_link.user_id != user.id
+          existing_link.update_attributes!(user: user)
+        end
+      else
+        UserExternalUuid.create!(user: user, uuid: trusted_external_uuid)
+      end
```
jpslav: There are two other places where we set the user's external UUID (`TransferSignupState` and `UserFromSignupState`) -- what situation does this code handle that those don't?

spec/features/trusted_login_flow_spec.rb
```diff
@@ -19,15 +19,38 @@
...
+    it 'pre-fills email on sign in when there is a match' do
+      user = create_user 'user'
+      create_email_address_for(user, payload[:email])
+      arrive_from_app(params: signed_params)
+      expect(page).to have_field('login_username_or_email', with: payload[:email])
+      click_button(t :"signup.start.next")
+      complete_login_password_screen 'password'
+      expect_back_at_app
+      expect(user.external_uuids.where(uuid: payload[:uuid])).to exist
+    end
```
jpslav: Can you write a spec where `payload[:role] = 'student'`?  Want to make sure it does this same pre-filled login page instead of signup.

spec/features/trusted_login_flow_spec.rb
```diff
@@ -19,15 +19,38 @@
...
+
+    it 'auto signs in and returns when linked' do
```
jpslav: Is this spec a dupe of `describe 'coming from app when already linked'` below?

app/routines/transfer_signup_state.rb
```diff
@@ -4,9 +4,9 @@ class TransferSignupState
 
   def exec(signup_state:, user:)
     fatal_error(code: :no_signup_email) if signup_state.nil?
-    fatal_error(code: :not_verified) if !signup_state.verified?
 
-    run(AddEmailToUser, signup_state.contact_info_value, user, {already_verified: true})
+    run(AddEmailToUser, signup_state.contact_info_value, user, {already_verified: signup_state.verified?})
```
jpslav: I know we don't have verified emails now...but this is old code... wondering why it is ok to now to have unverified emails transferred to the user... when from way back (when we didn't have trusted data) we only transferred verified emails?

lib/omniauth/strategies/custom_identity.rb
nathanstitt: This accommodates the flow where a user arrives with trusted data and then signs in.   After they sign in we need to link the account to the provided uuid so we can auto-redirect them back in subsequent visits.

I believe `custom_identity` is the only code that touches the user after they authenticate and before they're redirected back.  Maybe this should go into a routine though given your comment above about not fattening the `SignupState` model.

spec/features/trusted_login_flow_spec.rb
nathanstitt: yes it is. I'll remove it and move the terms spec up

spec/features/trusted_login_flow_spec.rb
nathanstitt: yep, good idea

lib/omniauth/strategies/custom_identity.rb
jpslav: I am fairly sure that a user who logs in will pass back through `SessionsCreate`, specifically the [handle_during_login](https://github.com/openstax/accounts/blob/master/app/handlers/sessions_create.rb#L62-L76) method.  This handler also has access to the `signup_state` (passed in from the controller), so it is probably a good place to add the `UserExternalUuid` if it doesn't exist yet.  Also, I believe that the `CustomIdentity` stuff modified here is only used for password logins, so social logins may miss out on the current way of linking external UUIDs (whereas `SessionsCreate` should not).  

It would be nice to have a spec that shows that someone logging in socially gets their external UUID added.  Unfortunately looking around, we may not have any specs that test successful social login, tho we do have some unsuccessful ones.  Search `with_omniauth_test_mode`.

nathanstitt: Yep, you're correct!  the code there isn't needed since `SessionsCreate` calls the `TransferSignupState`  routine

app/routines/transfer_signup_state.rb
nathanstitt: Hm, well we could force the user through verification but that seems like it'd be cumbersome, especially since the FE will now prompt the user to verify the email using the navbar nags.

Seems to me like it's best to just transfer what we've got and let the UI verify as needed

jpslav: ok, I will ponder this in the AM.

PR #556, opened 2017-10-26
placeholder PL translations for "submitting" disabled button state; submitting state for add/reset PW

jpslav:
With LMS Accounts work, we changed one locale heading from `new` to `start` that didn't make it into the Polish locale.

* Gets a locale heading change (`new` to `start`) from the LMS PR into the Polish locale
* Adds `submitting` translations to the Polish translation that just copy the text from the button that is being disabled
* Move the `submitting` translations in the English translation next to the translations in the button that are being disabled with the submitting message.
* Add button disabling and translations for adding and resetting passwords

### DIFF
--- a/app/views/identities/_set.html.erb
+++ b/app/views/identities/_set.html.erb
@@ -1,4 +1,6 @@
 <%
+  submitting_text ||= submit_text
+
   error_codes = @handler_result.errors.map(&:code)
 
   big_time_error_message =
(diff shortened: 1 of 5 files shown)


jpslav:
@aiwenar - we added placeholder Polish translations for buttons that are disabled after they have been pressed.  While the English translations change a clicked "Submit" button text to "Submitting...", I just copied the existing button text into the submitting state for the Polish translations.  We have a code freeze tomorrow if you want to update the pl locale before then, otherwise I suspect these are ok until your team can update.

apienk:
@jpslav Just pushed updates to the Polish "submitting" translations.

PR #559, opened 2017-11-03
suggested mod to authenticate_methods methods

jpslav:


lib/authenticate_methods.rb
```diff
@@ -34,14 +34,14 @@ def authenticate_admin!
...
-  # When the external site provides secure params the're
+  # When the external site provides signed params they're
   # requesting the person either be
-  # (1) automatically logged in if their UUID is known to us
-  # (2) Their UUID is remembered after they login so they're automatically logged in on subsequent visits
-  # (3) directed through a partially pre-populated sign up process, with the UUID being remembered at the end.
+  # (1) automatically logged in if their UUID is known to us, or
+  # (2) allowed to log in or sign up where we pre-populate some fields using
+  #     the signed data and remember their external UUID so that future requests
+  #     with signed params with this UUID can be automatically logged in
```
jpslav: I'm glad you noted the (2) option, but since the implementation was `something || another_thing` I changed the comment to just have two points too.

lib/authenticate_methods.rb
```diff
@@ -34,14 +34,14 @@ def authenticate_admin!
...
   def use_signed_params
-    auto_login_external_user || prepare_new_external_user_signup
+    auto_login_external_user || prepare_for_new_external_user
```
jpslav: Also, you rightfully pointed out that the `prepare_for...` method is not just for signup, so I took off `_signup`

lib/authenticate_methods.rb
```diff
@@ -50,35 +50,35 @@ def auto_login_external_user
...
+      # Sign out the current user; signing in the new user will effectively sign out
+      # the current user, but we choose to sign out explicitly so we can record
+      # information in the current user's security log
+
+      sign_out!(security_log_data: {type: 'different external user'})
     end
-    false
+
+    sign_in!(incoming_user, security_log_data: {type: 'external uuid'})
+    true
```
jpslav: This is the original logic that I think we still need.  We only want to `sign_out!` if the incoming user is different than the current user.

lib/authenticate_methods.rb
```diff
@@ -50,35 +50,35 @@ def auto_login_external_user
...
     # If we didn't find a user with a linked account to automatically log in,
-    # we do not want to assume that any already-logged-in user owns this secure
+    # we do not want to assume that any already-logged-in user owns this signed
     # params information.
     # Therefore at this point we sign out whoever is signed in.
-    if signed_in?
-      type = incoming_user == current_user ?
-               'untrusted external user' : 'different external user'
-      sign_out!(security_log_data: { type: type })
-    end
```
jpslav: if the incoming user is the current user, I don't think we should sign them out, but I'm open to hearing the thoughts behind `untrusted external user`.

PR #560, opened 2017-11-04
fix logins with signup state that were following signup flow

jpslav:
LMS launches send signed data to Accounts when authenticating.  That signed data is put into a `SignupState` object.  An authenticating user can try to log in in this situation, but the code was detecting the presence of the `SignupState` object and sending the user through some signup-related code.  

This PR has a quick fix to use the login flow instead.  Some other issues were found that have been put into a Trello card.

### DIFF
--- a/app/handlers/sessions_create.rb
+++ b/app/handlers/sessions_create.rb
@@ -47,12 +47,12 @@ class SessionsCreate
 
   def handle
     outputs[:status] =
-      if signing_up?
-        handle_during_signup
-      elsif signed_in?
+      if signed_in?
         handle_while_logged_in
       elsif logging_in?
         handle_during_login
+      elsif signing_up?
+        handle_during_signup
       else
         fatal_error(code: :unknown_callback_state)
       end
@@ -69,6 +69,10 @@ class SessionsCreate
          options[:login_providers][authentication.provider]['uid'] != authentication.uid
        )
       return :mismatched_authentication
+    end
+
+    if signup_state.present?
+      run(TransferSignupState, signup_state: signup_state, user: authentication_user)
     end
 
     sign_in!(authentication_user)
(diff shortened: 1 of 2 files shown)


app/handlers/sessions_create.rb
```diff
@@ -47,12 +47,12 @@ def authorized?
...
     outputs[:status] =
-      if signing_up?
-        handle_during_signup
-      elsif signed_in?
+      if signed_in?
         handle_while_logged_in
       elsif logging_in?
         handle_during_login
+      elsif signing_up?
+        handle_during_signup
```
jpslav: By changing the order of the checks we are able to avoid the case where a login attempt goes down the signup path.

app/handlers/sessions_create.rb
```diff
@@ -71,6 +71,10 @@ def handle_during_login
       return :mismatched_authentication
     end
 
+    if signup_state.present?
+      run(TransferSignupState, signup_state: signup_state, user: authentication_user)
+    end
```
jpslav: @nathanstitt your earlier idea that we needed to add code to set a user's UUID during login was correct. (the stuff you put in  `SignupState.on_authentication...`).  In that PR you saw that the UUIDs were getting added and took that code out.  They were getting added because of this bug where users were going down the signup flow during login.  Because they are now correctly going through this `handle_during_login` method, we need to transfer the UUIDs once again.

nathanstitt: What a great description of a trail of errors that somehow was still working 😭

Dantemss:
I think this is good but then again the login code is very confusing.

jpslav:
Yeah it is because adding authentications, logging in, and signing up all come back through the one oauth callback controller action.  I bet we could pass parameters in the oauth URL that get appended to the callback URL, and use that to determine which flow we are in.  Would help split up the code.

PR #562, opened 2017-11-13
Change users/uuid to not null and populate existing rows

mwharrison:


mwharrison:
[Accounts: Give all user records a uuid and make the field NOT NULL, making sure updates reach Tutor](https://trello.com/c/Mr0TcXZW/507-accounts-give-all-user-records-a-uuid-and-make-the-field-not-null-making-sure-updates-reach-tutor)

jpslav:
@Dantemss can you comment on why you take the [approach here](https://github.com/openstax/tutor-server/blob/master/db/migrate/20171109210427_assign_missing_uuids_for_local_accounts.openstax_accounts.rb#L6-L10) instead?

mwharrison:
@Dantemss @jpslav? Any thoughts on this?

Dantemss:
@mwharrison could you change this migration to be like the one JP linked? Right now the column type is string which is not ideal. The linked migration also sets the default for the column to a randomly-generated uuid, which is not strictly necessary but would be nice to have.

mwharrison:
@Dantemss - just so I understand better:
1. What is the type of field that is better suited in Rails for storing a UUID? I couldn't find mention of a specific UUID field type (like we have in Django).
2. What is the difference between what is happening in the linked migration vs. `SecureRandom.uuid`? It looks like they both would create a random UUID.

mwharrison:
@Dantemss ping :)

Dantemss:
@mwharrison changing the column type would make the column in postgres a uuid type instead of string. Both work, but uuid is more efficient (especially since we are doing searches on uuid). Having it as a uuid type instead of string also allows postgres to set the uuid by itself, which is marginally more efficient than having Rails do it.

app/routines/search_users.rb
```diff
@@ -83,9 +83,11 @@ def exec(query, options={})
       end
 
       with.keyword :uuid do |uuids|
-        sanitized_uuids = sanitize_strings(uuids, append_wildcard: options[:admin],
-                                            prepend_wildcard: options[:admin])
-        users = users.where{uuid.like_any sanitized_uuids}
+        if not uuids[0].include? '-' and uuids[0].to_s.length == 8
```
Dantemss: Can you replace this condition with `if uuids.size == 1 and not (partial_uuid = uuids[0].to_s.chomp('-')).include? '-' and partial_uuid.length == 8`? (and use the partial_uuid variable below instead of uuids[0])

Just a minor thing to make it not have the `between` behavior for arrays with more than 1 element and to ignore `-` at the end of the uuid.

app/routines/search_users.rb
```diff
@@ -83,9 +83,11 @@ def exec(query, options={})
...
+            users = users.where(:uuid => uuids[0] + '-0000-0000-0000-000000000000'..uuids[0] + '-ffff-ffff-ffff-ffffffffffff')
```
Dantemss: Can you get rid of the hash rocket `=>`? I've personally been advocating for using the new hash syntax when possible: (`uuid: partial_uuid + ...`)

app/routines/search_users.rb
Dantemss: This might work better actually:

```
uuids_queries = uuids.map do |uuid|
  partial_uuid = uuid.to_s.chomp('-')
  next uuid if partial_uuid.include? '-' or partial_uuid.length != 8

  partial_uuid + '-0000-0000-0000-000000000000'..partial_uuid + '-ffff-ffff-ffff-ffffffffffff'
end
users = users.where(uuid: uuids_queries)
```

PR #564, opened 2017-12-19
Show message if user is coming from OSweb Tutor page

mwharrison:


mwharrison:
[SF.onboarding.04: Add copy to the Accounts page if user comes from OXT marketing page](https://trello.com/c/6Z7QMQY1/525-sfonboarding04-add-copy-to-the-accounts-page-if-user-comes-from-oxt-marketing-page)

mwharrison:
@jpslav How is that sent with the request? I could use it - if I could figure it out.
FWIW, we don't really care about dependability here.

mwharrison:
Oooh - also @jpslav - this should only be displayed if they come from the Tutor marketing page, not all of OSWeb.

jpslav:
@mwharrison see https://stackoverflow.com/a/6023980 for some explanation.  Basically browsers set this field and they don't always set it to what you'd want.  We've tried to use referrers in the past but it has always been unsatisfactory.  Sometimes it works, sometimes it doesn't and you end up with bug reports getting filed, devs scratching their heads, time wasted, yadda yadda.

I think you're talking about the links on https://openstax.org/openstax-tutor ?  Log yourself out of Accounts and out of Tutor and then click one of those buttons, e.g. "Access Your Course".  When you arrive to Accounts, you'll see a `client_id` in the URL.  This is the ID of the osweb OAuth application on Accounts.  You have several options here.  One question for marcom/ux is whether it is bad for all traffic from osweb to get a message.  It might not be in which case seeing that the client is osweb might be all you need.  (this seems like a good option to me).  If we really need it to only be shown from this one page, the more reliable way would be to pass another parameter (e.g. `show_arrival_msg=1` or similar) that gets forwarded on to Accounts.  This is non-trivial because we don't forward all parameters to Accounts, so it'd have to be whitelisted in the code.  

Or you can use the referrer but have to know that it will be hit and miss.

Also, you may think about whether the message being shown can be internationalized, since all (or almost all) other text content in Accounts is available in English and Polish.

mwharrison:
@jpslav - turns out the link from osweb goes to Tutor, and if the user is not logged in it redirects to accounts. So the client is actually Tutor. I spoke with Marketing, and they agreed to show the message to all users coming from Tutor. @Dantemss also helped me do this in a rather nice way!

Dantemss:
Looks almost good to me:

- I think the application name is actually Tutor, so I suggest using `downcase`: `@client_app.name.downcase.include?` etc..
- Indentation is slightly off

mwharrison:
@Dantemss - thanks. Both items fixed, I guess `include?` is case sensitive then. 😄

app/views/sessions/start.html.erb
```diff
@@ -36,6 +36,12 @@
                                   errors: @handler_result.try(:errors),
                                   error_field_classes: "alert alert-danger") %>
 
+      <% if @client_app  %>
+          <% if @client_app.name.downcase.include? "tutor" %>
```
Dantemss: Could avoid the double `if` by using `&&` or `try!()` but it's very minor.

Dantemss:
Oh by the way I suggest testing on accounts-dev to make sure it all works (since it involves interactions between accounts and tutor, the specs might not catch everything)

mwharrison:
@Dantemss @jpslav - I tried adding translations (still waiting on the official translated text), can you confirm I did it correctly?

app/views/sessions/start.html.erb
```diff
@@ -38,7 +38,7 @@
 
       <% if @client_app  %>
           <% if @client_app.name.downcase.include? "tutor" %>
-            <div class="tutor-message">Log in or sign up for an account to access OpenStax Tutor</div>
```
Dantemss: To me calling the message just `tutor` is a bit too generic (though the `sessions` namespace does help I guess), I would suggest something like `access_tutor` at least.

app/views/sessions/start.html.erb
```diff
@@ -38,7 +38,7 @@
...
+            <div class="tutor-message"><%= t '.tutor' %></div>
```
Dantemss: To me calling the message just `tutor` is a bit too generic (though the `sessions` namespace does help I guess), I would suggest something like `access_tutor` at least.

app/views/sessions/start.html.erb
mwharrison: I actually cringed writing this - but it was erroring out without it. I didn't know about the `&&`, let me update this.

mwharrison:
@Dantemss - is there a spec for testing the translations somewhere?

Dantemss:
We've been slacking with those translation tests, so there are none in accounts currently, but I believe you would set them up like these view specs in Tutor: https://github.com/openstax/tutor-server/blob/master/spec/views/webview/index.html.erb_spec.rb

Dantemss:
Instead of testing that the page is blank, you could test that it includes some specific translation (or just test that the page is not blank and rely on the fact that missing translations will raise exceptions in the test environment)

RoyEJohnson:
Is this good to merge?

Dantemss:
@RoyEJohnson yes, I think so, unless @mwharrison wanted to make more specs first or something

mwharrison:
I’m good. Free to merge.

PR #566, opened 2017-12-21
Add is_test boolean field to users

Dantemss:
For https://trello.com/c/57bX4YFN

JP said it could be useful to have this field live in Accounts.

Defaults to `nil` which means we don't know if it's a test account or not, but can be set to true or false by admins.

The `is_test?` method (with a `?`) will always return a boolean (returns false for nil)

### DIFF
--- a/app/representers/api/v1/user_representer.rb
+++ b/app/representers/api/v1/user_representer.rb
@@ -45,6 +45,12 @@ module Api::V1
              readable: true,
              writeable: false
 
+    property :is_test?,
+             as: :is_test,
+             type: :boolean,
+             readable: true,
+             writeable: false
+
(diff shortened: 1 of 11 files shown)


app/representers/api/v1/user_representer.rb
```diff
@@ -45,6 +45,12 @@ class UserRepresenter < Roar::Decorator
              readable: true,
              writeable: false
 
+    property :is_test?,
+             as: :is_test,
+             type: :boolean,
+             readable: true,
+             writeable: false
```
jpslav: Can you add a `schema_info` description that says: "If true, the user is not a real OpenStax end user but instead an internal test user."

Dantemss:
Change made

PR #567, opened 2017-12-26
Push application source with SF leads

jpslav:
Lets admins set an "application source" for each oauth app.  When users sign up from one of those apps, this new source field is sent along with the lead to Salesforce.  

Will require us to set Tutor's source value to "Tutor Signup" after deployment

### DIFF
--- a/app/routines/push_salesforce_lead.rb
+++ b/app/routines/push_salesforce_lead.rb
@@ -4,7 +4,8 @@ class PushSalesforceLead
 
   protected
 
-  def exec(user:, email: nil, role:, phone_number:, school:, num_students:, using_openstax:, url:, newsletter:, subject:)
+  def exec(user:, email: nil, role:, phone_number:, school:, num_students:,
+           using_openstax:, url:, newsletter:, subject:, source_application:)
 
     raise(IllegalArgument, "Cannot push SF Leads for student roles") if role.match(/student/i)
 
@@ -33,7 +34,8 @@ class PushSalesforceLead
       website: url,
       adoption_status: using_openstax,
       num_students: num_students.to_i,
-      os_accounts_id: user.id
+      os_accounts_id: user.id,
+      application_source: source_application.try(:lead_application_source) || ''
     )
 
     lead.save
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -33,6 +33,12 @@ Rails.application.configure do
   # Raises helpful error messages.
   config.assets.raise_runtime_errors = true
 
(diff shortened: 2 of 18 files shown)


config/environments/development.rb
```diff
@@ -33,6 +33,12 @@
   # Raises helpful error messages.
   config.assets.raise_runtime_errors = true
 
+  # Tell Action Mailer not to deliver emails to the real world.
+  # The :test delivery method accumulates sent emails in the
+  # ActionMailer::Base.deliveries array.  If SMTP used instead
+  # but without good credentials, results in 30 second pauses.
+  config.action_mailer.delivery_method = :test
```
Dantemss: Not sure if the test mailer is usually used in the dev environment?

db/migrate/20171226134252_add_lead_source_to_oauth_applications.rb
```diff
@@ -0,0 +1,6 @@
+class AddLeadSourceToOauthApplications < ActiveRecord::Migration
+  def change
+    add_column :oauth_applications, :lead_application_source, :string,
+                                    null: false, default: ''
```
Dantemss: Not a huge fan of the blank string as a default. I'd maybe leave the column as nullable, if some apps have no lead_application_source. But I guess if Salesforce expects a blank string that may be ok.

app/routines/push_salesforce_lead.rb
```diff
@@ -33,7 +34,8 @@ def exec(user:, email: nil, role:, phone_number:, school:, num_students:, using_
       website: url,
       adoption_status: using_openstax,
       num_students: num_students.to_i,
-      os_accounts_id: user.id
+      os_accounts_id: user.id,
+      application_source: source_application.try(:lead_application_source) || ''
```
Dantemss: Is a blank string what salesforce expects when we don't know the source?

Dantemss: Also, does SF accept any string or just some specific strings?

jpslav: specific strings defined by SF admin (denver et al)

jpslav: blank string works when we don't know the string

config/environments/development.rb
jpslav: without this we get 30 second hang times when there's an exception in dev mode (I fixed this in Tutor some time ago and just copied the fix over here).  If someone actually needs to test mailing they can change this to `:smtp` and provide some credentials.

PR #570, opened 2018-01-05
Do not explode when trusted data missing email

jpslav:
In some situations, the trusted data that Accounts receives can be missing email.  We convert this data into a `SignupState` record which requires the email to be present.  In the case where it is missing, this PR allows the missing data up until the point where the user starts filling in forms, after which we require the email

### DIFF
--- a/app/handlers/signup_start.rb
+++ b/app/handlers/signup_start.rb
@@ -34,6 +34,7 @@ class SignupStart
 
     # Create a new one
     new_signup_state = SignupState.email_address.create(
+      is_partial_info_allowed: false,
       contact_info_value: email,
       role: signup_params.role,
       trusted_data: existing_signup_state.try(:trusted_data),
--- a/app/models/signup_state.rb
+++ b/app/models/signup_state.rb
@@ -1,5 +1,5 @@
 class SignupState < ActiveRecord::Base
-  attr_accessible :contact_info_value, :role, :return_to, :trusted_data, :verified
+  attr_accessible :contact_info_value, :role, :return_to, :trusted_data, :verified, :is_partial_info_allowed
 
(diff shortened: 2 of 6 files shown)


app/handlers/signup_start.rb
```diff
@@ -34,6 +34,7 @@ def handle
 
     # Create a new one
     new_signup_state = SignupState.email_address.create(
+      is_partial_info_allowed: false,
```
jpslav: Don't allow partial info (incomplete contact info) once a user actually fills out the signup start form

app/models/signup_state.rb
```diff
@@ -22,6 +22,7 @@ def self.create_from_trusted_data(data)
     role = User.roles[data[:role]] ? data['role'] : nil
     data['external_user_uuid'] = data.delete('uuid')
     SignupState.create!(
+      is_partial_info_allowed: true,
```
jpslav: do allow partial info (missing email) when building from trusted data

db/migrate/20180105205907_allow_partial_signup_state.rb
```diff
@@ -0,0 +1,8 @@
+class AllowPartialSignupState < ActiveRecord::Migration
+  def change
+    change_column_null :signup_states, :contact_info_value, true
+    change_column_null :signup_states, :contact_info_kind, true
+
+    add_column :signup_states, :is_partial_info_allowed, :boolean, default: false, null: false
+  end
```
nathanstitt: I'm not sure if we need to store `is_partial_info_allowed`.  It could just be an `attr_accessor` for so the validation could read it while saving.

However it's just a bool column and doesn't hurt anything really

PR #571, opened 2018-01-08
Tutor message

mwharrison:


mwharrison:
[SF.onboarding.04: Add copy to the Accounts page if user comes from OXT marketing page](https://trello.com/c/6Z7QMQY1/525-sfonboarding04-add-copy-to-the-accounts-page-if-user-comes-from-oxt-marketing-page)

PR #574, opened 2018-01-19
LMS students who have a pre-existing account

nathanstitt:
When there's a matching email for a student we send them to sign-in and if they do switch to sign-up we display a warning and don't proceed to sending a validation email


nathanstitt:
[Bug: Student with previous email gets caught in LMS signup loop](https://trello.com/c/Zogs5f8F/1199-bug-student-with-previous-email-gets-caught-in-lms-signup-loop)

PR #576, opened 2018-02-01
Make screenreader read login errors

RoyEJohnson:


RoyEJohnson:

app/views/layouts/_attention.html.erb
```diff
@@ -1,6 +1,6 @@
 <% errors = collect_errors %>
 
-<div class="errors">
+<div class="errors" aria-live="polite">
   <%= alert_tag(errors[:alerts])   %>
   <%= notice_tag(errors[:notices]) %>
 </div>
```
nathanstitt: Hey @RoyEJohnson would `alert` be a better role for these?  I'm not sure but that seems to fit the purpose of the errors div

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

RoyEJohnson: Reading about it, I think so. I'll put that in.

nathanstitt: Yeah, I'd never used anything other than `polite` before but it seemed appropriate here 😁

README.md
```diff
@@ -29,6 +29,20 @@ link to sign up.
...
+```sh
+$ brew install postgresql
+$ brew services start postgresql
+$ psql postgres
+CREATE ROLE ox_accounts WITH LOGIN;
+ALTER USER ox_accounts WITH SUPERUSER;
+\q
+```
+
 ### Running as a normal Rails app on your machine
```
nathanstitt: Yay!

PR #577, opened 2018-02-05
Fix trusted? being called on nil

RoyEJohnson:
Added checks to test that variable is non-null where there were `trusted?` method calls.

### DIFF
--- a/app/routines/user_from_signup_state.rb
+++ b/app/routines/user_from_signup_state.rb
@@ -6,7 +6,7 @@ class UserFromSignupState
 
   def exec(signup_state)
     user = User.new
-    if signup_state.trusted?
+    if signup_state && signup_state.trusted?
       user.role = signup_state.role
       user.full_name = signup_state.trusted_data['name']
       user.self_reported_school = signup_state.trusted_data['school']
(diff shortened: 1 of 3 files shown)


RoyEJohnson:
[Accounts 500 error: A NoMethodError occurred in exceptions#rescue_from: undefined method `trusted?' for nil:NilClass](https://trello.com/c/4lDayJZo/592-accounts-500-error-a-nomethoderror-occurred-in-exceptionsrescuefrom-undefined-method-trusted-for-nilnilclass)

app/routines/user_from_signup_state.rb
```diff
@@ -6,7 +6,7 @@ class UserFromSignupState
 
   def exec(signup_state)
     user = User.new
-    if signup_state.trusted?
+    if signup_state && signup_state.trusted?
       user.role = signup_state.role
```
nathanstitt: These work and are arguably the way to go.  There is a rubyism that uses `try` to work around possibly nil values. 
For instance this could be:
```
signup_state.try(:trusted?)
```
Personally I find the <value> && <value>.<method> equally clear though.

PR #578, opened 2018-02-05
Add faculty verification to user security log

mwharrison:


mwharrison:
[Write event to Accounts security log when user becomes faculty verified.](https://trello.com/c/uMi1zB3X/474-write-event-to-accounts-security-log-when-user-becomes-faculty-verified)

app/routines/update_user_salesforce_info.rb
```diff
@@ -191,6 +191,13 @@ def cache_contact_data_in_user!(contact, user)
 
       user.faculty_status = case contact.faculty_verified
       when "Confirmed"
+        SecurityLog.create!(
+            user: user.try(:is_anonymous?) ? nil : user,
```
Dantemss: I think the anonymous check is not needed here, the user should never be anonymous if they are getting faculty-verified. You should be able to just pass in the user directly.

app/routines/update_user_salesforce_info.rb
```diff
@@ -191,6 +191,13 @@ def cache_contact_data_in_user!(contact, user)
...
+            application: nil,
+            remote_ip: nil,
+            event_type: :faculty_verified,
+            event_data: user
```
Dantemss: This field should be a json object, not a ruby object. We don't usually pass in a user to event_data (except apparently in a few tests for some reason). I recommend you set the event_data to something like `{ user_id: user.id }`.

app/routines/update_user_salesforce_info.rb
```diff
@@ -191,6 +191,13 @@ def cache_contact_data_in_user!(contact, user)
 
       user.faculty_status = case contact.faculty_verified
       when "Confirmed"
+        SecurityLog.create!(
+            user: user,
+            application: nil,
+            remote_ip: nil,
+            event_type: :faculty_verified,
+            event_data: { user_id: user.id, salesforce_id: contact.id }
```
mwharrison: @Dantemss - what do you think about also having the Salesforce ID, so you can easily go look them up if needed?

Dantemss: That would be great. The more info you can provide the better IMO.

mwharrison: Cool - so does this look ready @Dantemss?

PR #581, opened 2018-02-09
Test for nil array in custom_identity

RoyEJohnson:


RoyEJohnson:
[Bug: Accounts sends an error email to devs if user tries to login with an invalid password and params['login'] is not present (need to investigate when this happens)](https://trello.com/c/v3G7nO4V/598-bug-accounts-sends-an-error-email-to-devs-if-user-tries-to-login-with-an-invalid-password-and-paramslogin-is-not-present-need-to)

PR #582, opened 2018-02-09
Fix: "Pin can't be blank" message appears twice

RoyEJohnson:


RoyEJohnson:
[email pin validation message appears twice](https://trello.com/c/g5e1dCC6/622-email-pin-validation-message-appears-twice)

PR #586, opened 2018-02-28
Updated omniauth gem

Dantemss:
Security


Dantemss:
Codecov failure is bogus. Gemfile.lock is not part of the coverage.

PR #587, opened 2018-02-28
Get school type info from Salesforce

Dantemss:
Requires https://github.com/openstax/openstax_salesforce/pull/12 (already merged and pushed to rubygems)
Required for https://trello.com/c/nwTcoFWn

### DIFF
--- a/app/routines/update_user_salesforce_info.rb
+++ b/app/routines/update_user_salesforce_info.rb
@@ -1,4 +1,9 @@
 class UpdateUserSalesforceInfo
+  COLLEGE_TYPES = [
+    'College/University (4)',
+    'Technical/Community College (2)',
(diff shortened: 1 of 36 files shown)


app/routines/update_user_salesforce_info.rb
```diff
@@ -203,10 +209,20 @@ def cache_contact_data_in_user!(contact, user)
         :pending_faculty
       when /Rejected/
         :rejected_faculty
-      when nil
+      when NilClass
```
jpslav: I should know why this is important but do not - fill me in?

Dantemss: It's not important, because in ruby `nil == nil` but I feel bad about it because that's not true in SQL for example.

PR #594, opened 2018-03-13
Pass param to auth providers

Dantemss:
Last paragraph of https://trello.com/c/1F820U9N

Includes https://github.com/openstax/accounts/pull/593

Also has code that tries to separate the 2 flows of adding a new auth provider vs signing in/signing up by passing a parameter to the OAuth provider that should make it back to Accounts.

Tested on accounts-dev and seemed to work fine.

### DIFF
--- a/app/handlers/sessions_create.rb
+++ b/app/handlers/sessions_create.rb
@@ -120,23 +120,36 @@ class SessionsCreate
   end
 
   def handle_while_logged_in
-    # Normally this happens when adding authentications on the profile screen.
+    # Attempt to login when already logged in
 
+    # Same user: don't do anything
     return :no_action if authentication_user == current_user
 
-    return :authentication_taken if authentication_user && authentication_user.is_activated?
+    # Check if they are trying to add a new authentication to their account
+    if request.env['omniauth.params'].try!(:[], 'add') == 'true'
+      # Add the new authentication
+      return :authentication_taken if authentication_user && authentication_user.is_activated?
 
-    return :same_provider \
-      if current_user.authentications.map(&:provider).include?(authentication.provider)
+      return :same_provider \
+        if current_user.authentications.map(&:provider).include?(authentication.provider)
 
-    return :new_signin_required if user_signin_is_too_old?
+      return :new_signin_required if user_signin_is_too_old?
 
-    return :email_already_in_use if ContactInfo.verified.where(value: @data.email)
-                                                        .where{user_id != my{current_user.id}}.any?
+      return :email_already_in_use if ContactInfo.verified.where(value: @data.email)
+                                                          .where{user_id != my{current_user.id}}
+                                                          .exists?
 
-    run(TransferAuthentications, authentication, current_user)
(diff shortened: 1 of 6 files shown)


app/handlers/sessions_create.rb
```diff
@@ -120,23 +120,36 @@ def handle_during_signup
...
-    return :authentication_added
+      run(TransferAuthentications, authentication, current_user)
+      run(TransferOmniauthData, @data, current_user) if authentication.provider != 'identity'
+      return :authentication_added
+    # Fallback to one of the other flows
+    elsif logging_in?
+      handle_during_login
+    elsif signing_up?
+      handle_during_signup
+    else
```
jpslav: Instead of having a fallback here, can we change the `handle` method to control which sub "handle" we call?

```ruby
def handle
  outputs.status = 
    if adding_authentication? # new method that is `signed_in? && see the add flag`
      handle_add_authentication
    elsif # actually is there a separate `handle_while_logged_in` case?
    elsif ...
...
```

app/handlers/sessions_create.rb
```diff
@@ -120,23 +120,36 @@ def handle_during_signup
   end
 
   def handle_while_logged_in
-    # Normally this happens when adding authentications on the profile screen.
+    # Attempt to login when already logged in
 
+    # Same user: don't do anything
     return :no_action if authentication_user == current_user
```
jpslav: Can this be moved out to the main `handle`?

app/handlers/sessions_create.rb
jpslav: Also is the implementation of `signing_up?` still valid?  `pre_auth_state` can exist during login too?

spec/handlers/sessions_create_spec.rb
```diff
@@ -1,199 +1,96 @@
 # coding: utf-8
 require 'rails_helper'
+require_relative 'sessions_create_shared_examples'
```
jpslav: Why did all this code move to shared examples if only used here?

Dantemss: It's used in here in 2 places: https://github.com/openstax/accounts/pull/594/files/2be24f2eb634048763f81397d754701f890a0b67#diff-c80901fc9ff3f8b10d4b0564dd4e2b85R7 and https://github.com/openstax/accounts/pull/594/files/2be24f2eb634048763f81397d754701f890a0b67#diff-c80901fc9ff3f8b10d4b0564dd4e2b85R27

Idk do you want me to move the shared examples to the top/bottom of this file?

app/handlers/sessions_create.rb
Dantemss: I think it makes more sense here because current_user will be nil elsewhere?

app/handlers/sessions_create.rb
Dantemss: OK I will give combinining them a try.

I think signing_up? is fine because we check it last, so it only triggers if they are not logging in instead.

PR #596, opened 2018-03-19
Add tests for special email situations

RoyEJohnson:
@Dantemss determined that the bug associated with the card has been fixed. So I wrote some new tests.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -92,7 +92,7 @@ en:
       temporarily_unavailable: >-
         The terms are temporarily unavailable. Check back soon.
 
-    lost_user: 'Sorry, we lost you. Please try again.'
+    lost_user: Sorry, we lost you. Please try again.
 
   handlers:
     authentications_delete:
--- a/spec/features/user_cant_sign_in_spec.rb
+++ b/spec/features/user_cant_sign_in_spec.rb
@@ -83,6 +83,23 @@ feature "User can't sign in", js: true do
 
       screenshot!
     end
+
+    scenario "user tries to sign up with used oauth email" do
+      user = create_user 'user'
+      authentication = FactoryGirl.create :authentication, provider: 'google_oauth2', user: user
+
+      arrive_from_app
+      click_sign_up
+      complete_signup_email_screen "Instructor", "unverified@example.com", screenshot_after_role: true
+
+      with_omniauth_test_mode(uid: authentication.uid) do
+        # Found link from back button or some other shenanigans
+        visit '/auth/google_oauth2?login_hint='
+      end
+
+      screenshot!
+      expect(page).to have_content('This is a fake external application')
+    end
   end
 
   context "we find one user", js: true do


RoyEJohnson:
[Accounts Illegal State 2 (not a dupe)](https://trello.com/c/3AN1eNpd/599-accounts-illegal-state-2-not-a-dupe)

app/controllers/sessions_controller.rb
```diff
@@ -157,6 +160,16 @@ def create
...
+        elsif unverified
+            Rails.logger.warn do
+              oauth = request.env['omniauth.auth']
+              errors = @handler_result.errors.inspect
+
+              "[SessionsCreate] Unverified account: " +
+              "OAuth data: #{oauth}; status: #{status}; errors: #{errors}"
+            end
+
+            redirect_to root_path, alert: I18n.t(:'controllers.lost_user')
```
jpslav: Now that we have two things we are looking for, can we pull this error code into a helper method?

```ruby
def log_and_send_to_root(dev_message, user_alert)
  ...
end
```

```ruby
error_codes = errors.map(&:code)

if error_codes.include?(:unknown_callback_state) || error_codes.include?(:invalid_omniauth_data)
  log_and_send_to_root("Lost user on failure", I18n.t(:'controllers.lost_user'))
elsif (error_codes.include?(:not_verified)
  log_and_send_to_root("Unverified account", I18n.t(:'controllers.lost_user')) # need a different message here.
else
 ...
end
```

jpslav: Need a different message displayed to the user, because I think we know more about how this happens.  That verbiage is in `config/locales/en.yml`.

RoyEJohnson: Is this test roughly what we're looking for? I think it's what's described on the card.
https://github.com/openstax/accounts/blob/fix-email-validation-message-appears-twice/spec/features/user_signs_up_spec.rb#L697-L719

jpslav: Looks similar except user wouldn't complete verification and may or may not need the 2nd browser.

RoyEJohnson:
I have added a spec, but I don't know what `expect` call to make.

app/controllers/sessions_controller.rb
```diff
@@ -141,22 +141,31 @@ def create
...
           Rails.logger.warn do
             oauth = request.env['omniauth.auth']
             errors = @handler_result.errors.inspect
 
-            "[SessionsCreate] Lost User on failure: " +
+            "[SessionsCreate] #{dev_message}: " +
             "OAuth data: #{oauth}; status: #{status}; errors: #{errors}"
           end
+          redirect_to root_path, alert: user_alert
+        end
```
jpslav: I suspect that defining the method inside the lambda (inside the `-> do ... end`) won't work.  If it does let me know! :-)

config/locales/pl.yml
```diff
@@ -82,6 +82,7 @@ pl:
         Warunki użytkowania są czasowo niedostępne. Spróbuj ponownie później.
 
     lost_user: Przepraszamy, nie mamy już Twoich danych. Spróbuj ponownie.
+    unverified_user: Adres e-mail Twojego konta nie został zweryfikowany.
```
jpslav: NICE!!

spec/features/user_cant_sign_in_spec.rb
```diff
@@ -59,6 +59,13 @@
       expect_authenticate_page
     end
 
+    scenario "user tries to log in with unverified account" do
+      user1 = create_user 'user1'
+      email1 = create_email_address_for(user1, 'user-1@example.com')
+      ContactInfo.where(id: email1.id).update_all(verified: false)
+      complete_login_username_or_email_screen('user1')
```
jpslav: The expectation is probably...

```ruby
expect(page).to have_content(t(:'controllers.unverified_user'))
expect_authentication_page # defined in `feature_helpers.rb`
```

I think the code redirects users to the root path, but I think that redirects to the login page (not 100% sure tho)

And if it doesn't pass, useful commands to call would be

`save_and_open_page`

Or somewhere before the last line of the spec scenario put in a `debugger` statement and then poke around in `page`.

app/controllers/sessions_controller.rb
RoyEJohnson: Tests pass, including one for lost_user, so I think it works.

app/controllers/sessions_controller.rb
```diff
@@ -141,22 +141,31 @@ def create
...
       end,
       failure: -> do
-        errors = @handler_result.errors
-        lost_user = errors.any? do |error|
-          [:unknown_callback_state, :invalid_omniauth_data].include? error.code
-        end
 
-        if lost_user
-          # Something weird happened to the user's session or omniauth data and we lost their info
+        def log_and_send_to_root(dev_message, user_alert)
```
Dantemss: Can you move this method definition to outside other methods? Having a method define another is not ideal.

app/controllers/sessions_controller.rb
```diff
@@ -141,22 +141,31 @@ def create
...
+          redirect_to root_path, alert: user_alert
+        end
 
-          redirect_to root_path, alert: I18n.t(:'controllers.lost_user')
+        errors = @handler_result.errors
+        lost_user = errors.any? do |error|
+          [:unknown_callback_state, :invalid_omniauth_data].include? error.code
+        end
+        unverified = errors.any? do |error|
+          error.code == :not_verified
```
Dantemss: I don't think this will happen anymore. Maybe remove/comment out since it's impossible to test?

app/controllers/sessions_controller.rb
```diff
@@ -141,22 +141,31 @@ def create
...
+        end
+
+        if lost_user
+          # Something weird happened to the user's session or omniauth data and we lost their info
+          log_and_send_to_root("Lost User on failure", I18n.t(:'controllers.lost_user'))
+        elsif unverified
+          log_and_send_to_root("Unverified account", I18n.t(:'controllers.unverified_user'))
```
Dantemss: Same as above.

config/locales/en.yml
```diff
@@ -92,7 +92,8 @@ en:
       temporarily_unavailable: >-
         The terms are temporarily unavailable. Check back soon.
 
-    lost_user: 'Sorry, we lost you. Please try again.'
+    lost_user: Sorry, we lost you. Please try again.
+    unverified_user: The email address for your account has not been verified.
```
Dantemss: These might not be used anymore since we determined the error doesn't happen.

config/locales/pl.yml
Dantemss: Same as above.

spec/features/user_cant_sign_in_spec.rb
```diff
@@ -83,6 +83,45 @@
 
       screenshot!
     end
+
+    scenario "user tries to login with unconfirmed email" do
```
Dantemss: This test is a dupe of https://github.com/openstax/accounts/blob/ff73ded343e11f578a3bfbf22436b9c6dbbe104a/spec/features/user_cant_sign_in_spec.rb#L10

spec/features/user_cant_sign_in_spec.rb
```diff
@@ -83,6 +83,45 @@
...
+      expect(page).to have_content(t(:'sessions.start.unknown_email'))
+    end
+
+    scenario "oauth user tries to login with unconfirmed email" do
+      user = create_user 'user'
+      email = create_email_address_for user, 'user@example.com', confirmation_code='not yet'
+      FactoryGirl.create :authentication, provider: 'google_oauth2', user: user
+      complete_login_username_or_email_screen('user')
+
+      with_omniauth_test_mode(email: 'user@example.com') do
```
Dantemss: This test is really testing that the uid provided by google does not match the uid of the authentication object created above (since you did not pass a uid to with_omniauth_test_mode, it uses a random one). Because of this, it is a dupe of https://github.com/openstax/accounts/blob/ff73ded343e11f578a3bfbf22436b9c6dbbe104a/spec/features/user_cant_sign_in_spec.rb#L205

PR #606, opened 2018-06-05
Update claiming account invite flow

nathanstitt:
This way it will either set or reset the password and will redirect back to the
application that created the account once terms are agreed to


nathanstitt:
[AN.Tech.14 Allow a power user to invite other users into the AN](https://trello.com/c/ASCK8gB7/1249-antech14-allow-a-power-user-to-invite-other-users-into-the-an)

PR #611, opened 2018-07-18
Add ability to create banners

BryanEliDimas:
This PR adds the ability to create alert-like banners to the main page/login form, to be used when parts of the system are down, like Facebook login.

Note: we (meaning I) may still have to add more more tests and we may want to move all the timezone logic to a shared location like application helper.

screen shot 2018-07-19 at 12 05 04 pm
[Create]
===
screen shot 2018-07-19 at 12 06 26 pm
[View all]
===
screen shot 2018-07-19 at 12 06 12 pm

### DIFF
--- a/app/controllers/admin/banners_controller.rb
+++ b/app/controllers/admin/banners_controller.rb
@@ -1,0 +1,30 @@
+module Admin
+  class BannersController < BaseController
+    layout 'admin'
+
+    def index
+      @banners = Banner.where('expires_at > ?', DateTime.now)
+    end
+
+    def new; end
+
+    def create
+      handle_with(BannersCreate,
+        success: lambda do 
+          flash[:success] = 'Banner created'
+          redirect_to admin_banners_path
+        end,
+        failure: lambda do
+          flash[:alert] = 'Banner created'
+          redirect_to :back
+        end
+      )
+    end
+
+    def destroy
+      banner = Banner.where(id: params[:id]).first
+      banner.destroy
+      redirect_to :back
+    end
+  end
+end
--- a/app/controllers/admin/banners_controller.rb
+++ b/app/controllers/admin/banners_controller.rb
@@ -9,13 +9,30 @@ module Admin
     def new; end
 
     def create
-      handle_with(BannersCreate,
-        success: lambda do 
+      handle_with(BannersManage,
+        success: lambda do
           flash[:success] = 'Banner created'
           redirect_to admin_banners_path
         end,
         failure: lambda do
-          flash[:alert] = 'Banner created'
+          flash[:alert] = 'Error in saving. Please try again.'
+          redirect_to :back
+        end
+      )
+    end
+
+    def edit
+      @banner = Banner.where(id: params[:id]).first
+    end
+
+    def update
+      handle_with(BannersManage,
+        success: lambda do
+          flash[:success] = 'Banner updated'
+          redirect_to admin_banners_path
+        end,
+        failure: lambda do
+          flash[:alert] = 'Error in saving. Please try again.'
           redirect_to :back
         end
       )
--- a/app/controllers/admin/banners_controller.rb
+++ b/app/controllers/admin/banners_controller.rb
@@ -1,9 +1,10 @@
 module Admin
   class BannersController < BaseController
     layout 'admin'
+    before_filter :delete_expired_banners, only: [:index]
 
     def index
-      @banners = Banner.where('expires_at > ?', DateTime.now)
+      @banners = Banner.active
     end
 
     def new; end
@@ -11,29 +12,25 @@ module Admin
     def create
       handle_with(BannersManage,
         success: lambda do
-          flash[:success] = 'Banner created'
-          redirect_to admin_banners_path
+          redirect_to admin_banners_path, notice: 'Banner created'
         end,
         failure: lambda do
-          flash[:alert] = 'Error in saving. Please try again.'
-          redirect_to :back
+          render :new
         end
       )
     end
 
     def edit
-      @banner = Banner.where(id: params[:id]).first
+      @banner = Banner.find_by(id: params[:id])
     end
 
     def update
       handle_with(BannersManage,
         success: lambda do
-          flash[:success] = 'Banner updated'
-          redirect_to admin_banners_path
+          redirect_to admin_banners_path, notice: 'Banner updated'
         end,
         failure: lambda do
-          flash[:alert] = 'Error in saving. Please try again.'
-          redirect_to :back
+          render :edit
         end
       )
     end
@@ -42,6 +39,13 @@ module Admin
       banner = Banner.where(id: params[:id]).first
       banner.destroy
       redirect_to :back
+    end
+
+    protected
+
+    def delete_expired_banners
+      expired_banners = Banner.where('expires_at < ?', DateTime.now)
+      expired_banners.delete_all
     end
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -35,7 +35,7 @@ class SessionsController < ApplicationController
 
   # Login form
   def start
-    @banners = Banner.where('expires_at > ?', DateTime.now)
+    @banners = Banner.active
   end
 
   def lookup_login
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -35,6 +35,7 @@ class SessionsController < ApplicationController
 
   # Login form
   def start
+    @banner_message = Banner.where('expires_at > ?', DateTime.now).first.try(:message)
   end
 
   def lookup_login
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -35,7 +35,7 @@ class SessionsController < ApplicationController
 
   # Login form
   def start
-    @banner_message = Banner.where('expires_at > ?', DateTime.now).first.try(:message)
+    @banners = Banner.where('expires_at > ?', DateTime.now)
   end
 
   def lookup_login
--- a/app/handlers/admin/banners_manage.rb
+++ b/app/handlers/admin/banners_manage.rb
@@ -21,12 +21,13 @@ module Admin
         banner.message = params[:banner][:message]
         banner.expires_at = expires_at
         banner.save
+        transfer_errors_from(banner, { type: :verbatim })
       end
     end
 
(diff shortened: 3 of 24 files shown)


BryanEliDimas:
[Be able to set banners and notification through accounts](https://trello.com/c/vQIMadW6/458-be-able-to-set-banners-and-notification-through-accounts)

app/controllers/admin/banners_controller.rb
```diff
@@ -0,0 +1,47 @@
+module Admin
+  class BannersController < BaseController
+    layout 'admin'
+
+    def index
+      @banners = Banner.where('expires_at > ?', DateTime.now)
```
Dantemss: Will need an index on expires_at. Might also be a good idea to automatically delete expired banners when they are accessed (if there's no way to see them) even for admins.

app/controllers/admin/banners_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+        end,
+        failure: lambda do
+          flash[:alert] = 'Error in saving. Please try again.'
+          redirect_to :back
+        end
+      )
+    end
+
+    def edit
+      @banner = Banner.where(id: params[:id]).first
```
Dantemss: Could use `.find_by` instead of `.where().first`

app/controllers/sessions_controller.rb
```diff
@@ -35,6 +35,7 @@ class SessionsController < ApplicationController
 
   # Login form
   def start
+    @banners = Banner.where('expires_at > ?', DateTime.now)
```
Dantemss: Maybe make this into a scope to avoid repeating it.

app/controllers/admin/banners_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    def new; end
+
+    def create
+      handle_with(BannersManage,
+        success: lambda do
+          flash[:success] = 'Banner created'
+          redirect_to admin_banners_path
+        end,
+        failure: lambda do
+          flash[:alert] = 'Error in saving. Please try again.'
```
Dantemss: This error message is not too explicit. Are the fields that have errors highlighted at least when this happens?

app/handlers/admin/banners_manage.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def handle
+      expires_at_central = string_from_date_select_params(params[:banner], :expires_at)
+      expires_at = central_to_utc(expires_at_central)
+
+      outputs[:banner] = Banner.where(id: params[:id]).first_or_initialize.tap { |banner|
+        banner.message = params[:banner][:message]
+        banner.expires_at = expires_at
+        banner.save
+      }
```
Dantemss: I prefer `do...end` for multiline blocks but you don't have to change this because a lot of people don't care anyway :)

app/handlers/admin/banners_manage.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    # consider moving this method into a shared location
+    # because it's something that Rails should've included
+    # as it is necessary for parsing dates from date_select tags
+    def string_from_date_select_params(params, key)
+      date_parts = params.select { |k,v| k.to_s =~ /\A#{key}\([1-6]{1}i\)/ }.values
+      date_parts[0..2].join('-') + ' ' + date_parts[3..-1].join(':')
```
Dantemss: It's not very clear what this code is doing. How about trying the approach that is in https://stackoverflow.com/a/13606990? Then you will have a date that is already parsed and won't need to parse the string again (in the method below).

app/helpers/application_helper.rb
```diff
@@ -154,13 +155,21 @@ def ox_card(classes: "", heading: "", &block)
         end
       end
 
+      banners_divs = if banners.present?
```
Dantemss: I would suggest using `unless banners.empty?` just because `present?` is a Rails extension, not standard Ruby.

But actually you should be able to just remove this check because an empty array should make banners_divs a perfectly fine blank string.

app/models/banner.rb
```diff
@@ -0,0 +1,2 @@
+class Banner < ActiveRecord::Base
```
Dantemss: I recommend adding the message presence validation here too.

config/routes.rb
```diff
@@ -202,6 +202,8 @@
 
     resources :pre_auth_states, only: [:index]
 
+    resources :banners
```
Dantemss: You don't have a show action in the controller, so would be good to use `except: :show` here.

spec/factories/banners.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :banner do
+    message "MyString"
+    expires_at "2018-07-17 17:30:39"
```
Dantemss: Might be a good idea to default to an expiration date in the future to avoid surprises, for example using `Time.now + 1.day` or something.

spec/models/banner_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Banner, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
Dantemss: Since the model is so simple, maybe just add a `it { should validate_presence_of :message }` (after adding the validation in the model above)

app/controllers/sessions_controller.rb
BryanEliDimas: Done.

app/helpers/application_helper.rb
BryanEliDimas: True. Done. I removed that check as you said.

app/models/banner.rb
BryanEliDimas: Ah. Done!

config/routes.rb
BryanEliDimas: Oh yeah, I missed that.

spec/factories/banners.rb
BryanEliDimas: Fixed.

app/handlers/admin/banners_manage.rb
BryanEliDimas: Oh. Yeah, I also prefer that, I just changed it because I wanted to paste it into the console all as one line and forgot to change it back. Fixed.

app/controllers/admin/banners_controller.rb
BryanEliDimas: Fixed, as we discussed, just by adding `render :edit`.

app/handlers/admin/banners_manage.rb
BryanEliDimas: Done.

spec/models/banner_spec.rb
BryanEliDimas: Done.

app/controllers/admin/banners_controller.rb
BryanEliDimas: Done, added an index on expires_at. And I added a `before_filter :delete_expired_banners, only: [:index]` as I can't think of any other way to do it automatically without creating a scheduled rake task.

app/controllers/admin/banners_controller.rb
BryanEliDimas: Yeah. True. Done.

BryanEliDimas:
@Dantemss Thanks Dante. I've made the suggested changes and I hope that the tests I've added are enough.

app/views/admin/banners/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <th style="width: 20%">Expires at</th>
+    <th>Message</th>
+    <th>Edit</th>
+    <th>Delete</th>
+  </tr>
+
+  <% @banners.each do |banner| %>
+    <% zebra = cycle('odd', 'even') %>
+    <tr class="<%= zebra %>">
+      <td><%= banner.expires_at.in_time_zone('UTC').strftime("%m/%d/%Y %I:%M%p") %></td>
```
BryanEliDimas: @Dantemss I don't know why this works this way, seems counterintuitive to me, but it does seem to work fine. I'd consider moving this timezone logic and `strftime` display logic somewhere else but not sure where or if it's really necessary.

BryanEliDimas:
@Dantemss I think the failing specs are unrelated to my code.

app/models/banner.rb
```diff
@@ -0,0 +1,10 @@
...
+  scope :active, -> { where('expires_at > ?', DateTime.now) }
+
+  validates :message, presence: true
+  validates :expires_at, presence: true
+
+  def expires_at=(expires_at)
+    super(expires_at).try(:in_time_zone, 'Central Time (US & Canada)')
```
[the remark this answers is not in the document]
Dantemss: Should be fine, but do you really care about the return value of `expires_at=`?

app/views/admin/banners/index.html.erb
BryanEliDimas: Right... that's right. Like I said, I don't know why this works this way but it looks right in the frontend. @Dantemss

app/models/banner.rb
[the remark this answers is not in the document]
BryanEliDimas: No I don't care about the return value, but I think what that code does is *parse* the time as Central time because that's what it is. Otherwise, it thinks that it's in UTC which it isn't. @Dantemss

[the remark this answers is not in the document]
Dantemss: What this code is doing is setting the expires_at to whatever time you pass in, then return that value as Central Time. I don't think the return part is needed.

app/views/admin/banners/index.html.erb
Dantemss: Hm can you move this code to the banner model (`def expires_at`) and write a few tests? Just to be safe.

Dantemss: Actually maybe call it something else since it returns a string.

app/models/banner.rb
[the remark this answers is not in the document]
Dantemss: Maybe just remove this method, I don't think it's actually doing much.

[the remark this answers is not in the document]
BryanEliDimas: Ohh. Yeah. Done.

BryanEliDimas:
All done.

BryanEliDimas:
Added a feature test and a model test to check for correctness in timezone.

victoria-cooke:
@BryanEliDimas Banner screen: https://invis.io/KJUHX9TD7BQ

PR #618, opened 2018-08-16
Add Adjunct Faculty role

RoyEJohnson:


RoyEJohnson:
[Add roles to faculty verification form leads](https://trello.com/c/cOPmkWIK/804-add-roles-to-faculty-verification-form-leads)

app/models/user.rb
```diff
@@ -34,7 +34,7 @@ class User < ActiveRecord::Base
   has_many :security_logs
 
   enum faculty_status: [:no_faculty_info, :pending_faculty, :confirmed_faculty, :rejected_faculty]
-  enum role: [:unknown_role, :student, :instructor, :administrator, :librarian, :designer, :other]
+  enum role: [:unknown_role, :student, :instructor, :adjunct, :administrator, :librarian, :designer, :other]
```
jpslav: Rails stores these enums as integers in the database, with the value corresponding to position in this array.  So `:adjunct` needs to be added at the end of the array, otherwise all roles from `:administrator` on will be incorrectly mapped to users.

https://api.rubyonrails.org/v4.1.0/classes/ActiveRecord/Enum.html

>Note that when an Array is used, the implicit mapping from the values to database integers is derived from the order the values appear in the array. In the example, :active is mapped to 0 as it's the first element, and :archived is mapped to 1. In general, the i-th element is mapped to i-1 in the database.
>
>Therefore, once a value is added to the enum array, its position in the array must be maintained, and new values should only be added to the end of the array.

jpslav: File this under things you can't know until you know them :-)

RoyEJohnson:
[Add Adjunct Faculty as a role on adoption, interest, and faculty verification forms.](https://trello.com/c/F3M791Eo/762-add-adjunct-faculty-as-a-role-on-adoption-interest-and-faculty-verification-forms)

PR #619, opened 2018-08-16
Create HTML email to confirm user account

mwharrison:


mwharrison:
The plan is to put this in the admin so it can be changed easily, but for now - this will hopefully help our pin confirmation emails from going into spam so frequently.

app/mailers/signup_confirmation_mailer.rb
```diff
@@ -6,7 +6,7 @@ def instructions(pre_auth_state:)
 
     res = mail to: "#{pre_auth_state.contact_info_value}",
          subject: @show_pin ?
-                    "Use PIN #{pre_auth_state.confirmation_pin} to confirm your email address" :
+                    "Please confirm your OpenStax account" :
```
jpslav: Why take the code out of the subject? I like it when confirmation codes are in the subject of emails or the first X characters of a text message so I can just read it directly.

jpslav:
@mwharrison re "hopefully help" - which part of these changes are expected to help us and why?

mwharrison:
@jpslav - well, we don't have a UX team right now... so this is all Allison and myself. Since she's the product manager, I'm assuming we've got a clearance to go ahead. Is there someone else that should be in the loop on this?

What will this "hopefully help"? We have extensive complaints from the CS team about users not receiving our emails - so this is an attempt to help that. Even those in the office complain about them being sent to spam sometimes.

mwharrison:
screen shot 2018-08-16 at 2 22 57 pm

mwharrison:
img_3700

mwharrison:
Also, here is the text-based email:

You're almost there!

We've finished setting up your OpenStax account, but you still
need you to confirm your email address.

Enter your unique six-digit PIN in your browser to confirm:
336990

Or, click the link below.

Confirm email 
( http://localhost:2999/signup/verify_by_token?code=e66703eca8b7e8f4c63f447d25983f4119558356958aaa9e90ffa27c0381f958 )


CONNECT WITH US:

Facebook ( https://www.facebook.com/OpenStax )

Twitter ( https://twitter.com/OpenStax )

Instagram ( https://instagram.com/OpenStax )

YouTube ( https://www.youtube.com/user/cnxorg )

jpslav:
Sorry guess I was asking if converting it to styled HTML is going to prevent it from going to spam?  It was already HTML (albeit very simple HTML).

The new page also promotes clicking the link over using the code; previously the link was there as a fallback and isn't the intended way for users to proceed.

I'm nervous about changing this without UX in the loop (and I know that is an impossibility at the moment).

mwharrison:
@jpslav Well, we're not really sure. We thought we'd give it a try... and assumed having a better-looking email couldn't hurt anything. We have no scientific evidence to back up the decision though. 😬 

Is there a reason the pin code is the preferred method? That seems to be opposite to the experience on most sites/confirmation emails.

I'm going to propose that the Allison, you and myself meet about this - we could also include others you think might be appropriate.

allisonhowell:
@jpslav I'm happy to talk with you about this a little bit more, but I feel pretty strongly that this change needs to be made. While I certainly understand having UX influence the accounts flow, etc., I don't think that's relevant here. These emails are going to spam and we are attempting to stop that from happening. I'll admit this is not a long-term fix, and we'll certainly need UX input on making any bigger changes, but this is not a UX decision.

edwoodward:
Allison is planning to have the emails reviewed by others. Based on the review, this PR will be updated or merged.

PR #622, opened 2018-09-11
Send Analytics if in Production

RoyEJohnson:
Previous PR https://github.com/openstax/accounts/pull/608
turned off Google Analytics unless the environment was production _and_ `send_google_analytics` was `true`. But it set `send_google_analytics` to `false`.
This PR sends Analytics if in Production, regardless of the `send_google_analytics` setting.

### DIFF
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -1,4 +1,4 @@
-<% if Rails.env.production? and Settings::GoogleAnalytics.send_google_analytics == true and !Settings::GoogleAnalytics.google_analytics_code.blank?  %>
+<% if (Rails.env.production? or Settings::GoogleAnalytics.send_google_analytics == true) and !Settings::GoogleAnalytics.google_analytics_code.blank?  %>
 <script>
 (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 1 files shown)


RoyEJohnson:
[Bug: Accounts no longer reporting in Google Analytics](https://trello.com/c/o9nLnNCx/931-bug-accounts-no-longer-reporting-in-google-analytics)

jpslav:
Note @mwharrison and @RoyEJohnson that `Rails.env.production?` is true in all deployed environments, including `-dev`, `-qa`, and `-staging` and what we call production.  Is that ok?

RoyEJohnson:
@jpslav Ugh. Is there any way to determine which environment it is deployed in? Does this become a devops issue for setting `send_google_analytics` per deployment?

jpslav:
https://github.com/openstax/accounts/blob/master/config/application.rb#L55-L57

https://github.com/openstax/accounts/blob/master/app/helpers/application_helper.rb#L125-L127

I think `send_google_analytics` is just set in the DB.... The original PR initialized `send_google_analytics` to false (because in general we don't want all deployments sending analytics, tho we may at times to test things).  Was this flag not set to true in production?  If it wasn't, do we just need to go into the admin settings and change it?

RoyEJohnson:
According to the Trello card (see earlier card in the thread), Analytics stopped receiving anything from Accounts as of July 17. Sounds like you can revert this PR and go into admin settings in the production environment.

PR #625, opened 2018-09-20
Add Introduction to Business

RoyEJohnson:


RoyEJohnson:
[Add Introduction to Business to Accounts for Faculty Verification](https://trello.com/c/kmOHdR2h/909-add-introduction-to-business-to-accounts-for-faculty-verification)

config/app.yml
```diff
@@ -59,6 +59,9 @@ defaults: &defaults
     economics:
       title: Principles of Economics
       sf: Economics
+    introduction_to_business:
```
edwoodward: Since this same info is in the CMS, it seems like we should be pulling it from there. I'll create a card.

PR #630, opened 2018-10-02
Add role to fields in Salesforce lead

RoyEJohnson:


RoyEJohnson:
[Add roles to faculty verification form leads](https://trello.com/c/cOPmkWIK/804-add-roles-to-faculty-verification-form-leads)

app/routines/push_salesforce_lead.rb
```diff
@@ -37,7 +37,8 @@ def exec(user:, email: nil, role:, phone_number:, school:, num_students:,
       adoption_status: using_openstax,
       num_students: num_students.to_i,
       os_accounts_id: user.id,
-      application_source: source_application.try(:lead_application_source) || ''
+      application_source: source_application.try(:lead_application_source) || '',
+      role: role
```
Dantemss: Hmm will salesforce accept the exact same roles that accounts accepts (things like instructor, administrator, librarian, student, other)? Otherwise you will need to do some role translation here.

Dantemss: You will also need to get the openstax_salesforce PR merged and released then do a `bundle update openstax_salesforce` and commit Gemfile.lock.

Dantemss: OK I checked the salesforce sandbox and it seems like it would accept the same values that accounts uses for role. In this case the only thing left to do is to `bundle update openstax_salesforce` to get the newest version (You should see version `2.1.0` which I just released and includes your PR to add the role field)

RoyEJohnson: Updated and committed

PR #632, opened 2018-10-11
Update faculty verification form

RoyEJohnson:
As far as I can tell, this is working properly. But I cannot verify what Salesforce is receiving.

Also it looks like I need to update about 100 tests.


RoyEJohnson:
I'm going to split this into two PRs: one that lays the groundwork without changing the app, and the other where all the substantive changes are.

PR #633, opened 2018-10-25
Properly expire SSO cookie on logout

mwharrison:


mwharrison:
Fixed with help from @Dantemss, see https://openstax.slack.com/archives/C6NRD0CCQ/p1540418762000100 for details.

PR #634, opened 2018-10-31
Update Faculty Verification form, phase 1

RoyEJohnson:
This PR makes changes to prepare for the changes in the forms, but does not include the forms changes. Everything works as it did previously, and all specs pass. There will be a separate PR with all the changed forms and updated specs.


RoyEJohnson:
Doing it all in one PR after all.

PR #635, opened 2018-11-02
Update faculty verification form

RoyEJohnson:
@nathanstitt I think I got all the changes you mentioned in the previous incarnation of this PR. I replaced the PR because labeling all the fields was not working out for me.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -202,8 +202,6 @@ group :development do
   gem 'web-console', '~> 2.0'
 
   gem  'i18n-tasks', '~> 0.9.6'
-  # Speedup and run specs when files change
-
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,10 @@ end
 gem 'rails', '4.2.7.1'
 gem 'rails-i18n', '~> 4'
 gem 'sprockets', '~> 2.12.5'
-gem 'pattern-library', git: 'https://github.com/openstax/pattern-library.git'
+gem 'pattern-library', git: 'https://github.com/openstax/pattern-library.git', branch: 'master'
+
+# Knockout
+gem 'knockoutjs-rails'
 
 # Bootstrap front-end framework
 gem 'bootstrap-sass', '~> 3.1.1'
@@ -199,6 +202,8 @@ group :development do
   gem 'web-console', '~> 2.0'
 
   gem  'i18n-tasks', '~> 0.9.6'
+  # Speedup and run specs when files change
+
 end
 
 group :test do
--- a/app/controllers/signup_controller.rb
+++ b/app/controllers/signup_controller.rb
@@ -96,11 +96,15 @@ class SignupController < ApplicationController
                     end
                   end,
                   failure: lambda do
-                    render :profile
+                      Rails.logger.warn '=== Here is your problem ==='
+                      Rails.logger.warn all_errors
+                      Rails.logger.warn '=== end of error listing ==='
+                      render :profile
                   end)
     else
       params[:profile] = {
-        school: current_user.self_reported_school
+        school: current_user.self_reported_school,
+        newsletter: "1"
       }
     end
   end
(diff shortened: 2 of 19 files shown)


spec/features/skipped_terms_spec.rb
```diff
@@ -37,31 +37,28 @@ def run_skipped_terms_test(app:, should_skip:)
...
     expect(ContactInfo.where(value: "bob@bob.edu").verified.count).to eq 1
     expect(PreAuthState.count).to eq 0
 
-    complete_instructor_access_pending_screen
+    #save_and_open_page
+    # if should_skip
+    #   expect(page).to have_no_content(terms_content)
+    # else
+    #   expect(page).to have_content(terms_content)
+    # end
```
nathanstitt: debug content 😁

spec/features/user_signs_up_spec.rb
```diff
@@ -22,38 +25,26 @@
...
-        expect(lead_in_sf).not_to be_nil
-
-        method.call(*args)
-      end
+      # expect_any_instance_of(PushSalesforceLead).to receive(:log_success).and_wrap_original do |method, *args|
+      #   lead_in_sf = OpenStax::Salesforce::Remote::Lead.where(id: args[0].id).first
+      #   expect(lead_in_sf).not_to be_nil
+      #
+      #   method.call(*args)
+      # end
```
nathanstitt: I think `PushSalesforceLead` is part of openstax_salesforce. Seems like the test should still be enabled or maybe moved to the appropriate location

spec/features/user_signs_up_spec.rb
```diff
@@ -417,26 +408,25 @@
...
+      # )
+      #
+      # [:first_name, :last_name, :phone_number, :school, :url, :using_openstax].each do |var|
+      #   expect(page).to have_content(error_msg SignupProfileInstructor, var, :blank)
+      # end
+      # expect(page).to have_content(error_msg SignupProfileInstructor, :num_students, :not_a_number)
+      # expect(page).to have_content(error_msg SignupProfileInstructor, :subjects, :blank_selection)
+      #
+      # screenshot!
     end
```
nathanstitt: we should still test that errors are shown when fields are left blank

Gemfile
```diff
@@ -199,6 +202,8 @@ group :development do
   gem 'web-console', '~> 2.0'
 
   gem  'i18n-tasks', '~> 0.9.6'
+  # Speedup and run specs when files change
+
```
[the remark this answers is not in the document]
RoyEJohnson: I don't know what that's from. Careless copying, maybe. The same comment appears in the previous block of code. It's also in the Tutor-server Gemfile.

spec/features/user_signs_up_spec.rb
RoyEJohnson: There's no active Next button when required fields are blank, so the attempt to submit (in complete_signup_profile_screen) fails.

RoyEJohnson: What I have done is tested that you can't click Next, or that Next does not change pages.

app/controllers/signup_controller.rb
```diff
@@ -96,11 +96,15 @@ def profile
                     end
                   end,
                   failure: lambda do
-                    render :profile
+                      Rails.logger.warn '=== Here is your problem ==='
+                      Rails.logger.warn all_errors
+                      Rails.logger.warn '=== end of error listing ==='
+                      render :profile
                   end)
```
nathanstitt: I think this is leftover dev code, since Rails.logger isn't really used on production. if you want to retain the errors, they should be logged to sentry, right?

spec/features/user_signs_up_spec.rb
```diff
@@ -447,31 +426,29 @@
...
-      expect(last).to eq('Not Listed')
-      expect(subjects).to eq(subjects.sort)
-    end
+    # scenario "subjects list is sorted correctly" do
+    #   subjects = all('.subjects .subject label').map(&:text)
+    #   last = subjects.pop
+    #   expect(last).to eq('Not Listed')
+    #   expect(subjects).to eq(subjects.sort)
+    # end
   end
```
nathanstitt: commented spec should be re-enbled or removed if it's no longer relevant (seems like the latter)

PR #638, opened 2018-11-14
Remove faculty verification from "Other" role

mwharrison:


spec/features/apply_for_faculty_access_spec.rb
```diff
@@ -7,36 +7,7 @@
...
-      url: "http://www.rice.edu",
-      subjects: ["Biology", "Principles of Macroeconomics"],
-      newsletter: true,
-    )
-
-    expect(page).to have_content(t :"faculty_access.pending.page_heading")
-    click_button(t :"faculty_access.pending.ok")
-
-    expect(page.current_url).to eq(capybara_url(external_app_for_specs_path))
-  end
```
Dantemss: I think we should leave this test in, just change the role that they pick to anything else. The other tests don't seem to test the flow where they visit the faculty then click to login (they login before going to the faculty page).

mwharrison: @Dantemss - when I add the test back and change the role, it fails the test. Any ideas? I did try this before you mentioned it, I felt it wouldn't be favorable to remove tests. :)

mwharrison: @Dantemss - Do you have a recommendation on how to fix the test besides changing the user role?

PR #644, opened 2019-01-25
Remove all the exernal uuids from an account

nathanstitt:
This is needed so that an LMS launch can be re-associated with a different user

image

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -221,7 +221,7 @@
     <%= f.label :trusted_launch_user, 'External authentications?', class: "col-sm-2 control-label" %>
     <div class="col-sm-10">
       <div class="checkbox">
-        <%= f.check_box :trusted_launch_user, checked: true, style: 'margin-left: 0px'  %>
+        <%= f.check_box :keep_external_uuids, checked: true, style: 'margin-left: 0px'  %>
         <i style="margin-left: 1rem">(Unchecking will remove all links between this account and external systems such as LMS)</i>
       </div>
     </div>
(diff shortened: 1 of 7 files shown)


app/views/admin/users/_form.html.erb
```diff
@@ -210,6 +216,17 @@
...
   </div>
 
+  <% if @user.external_uuids.any? %>
+  <div class="form-group">
+    <%= f.label :trusted_launch_user, 'Trusted Launch User?', class: "col-sm-2 control-label" %>
+    <div class="col-sm-10">
+      <div class="checkbox">
+        <%= f.check_box :trusted_launch_user, checked: true, style: 'margin-left: 0px'  %>
+      </div>
+    </div>
```
jpslav: Accounts doesn't know anything about launches so I'm not super enthused about putting that terminology here (could be in a help text I guess).  Is it clear that unchecking this removed external UUIDs?  Is there another way to do this that is clearer?

nathanstitt: yeah, I struggled with this.  I didn't want to explicitly say "Remove LMS launch data" since as you say it's not really for LMS only.  

How about about adding some help text that says "Unchecking will remove all links between this account and external systems such as LMS"

jpslav: yes - and make the code more about external uuids

jpslav: can you make the checkbox `:keep_external_uuids` or something?

nathanstitt:
@jpslav I can't recall for sure where we left this one, but I think you'd given it a 👍; can you confirm that it's ok to merge for next release?

PR #645, opened 2019-01-29
generate sso cookie using it's own key and salt

nathanstitt:
Previously it was actually using the same key as the main cookie.

This also sets it's options directly so it'll be more stable 🤞 not change until we need it to


nathanstitt:
@jpslav verbally approved 👀

PR #649, opened 2019-02-05
Created a 'validate role' method to restrict access to students and a…

KarinaMendez2:
…ny other than admins to control applications.

### DIFF
--- a/app/controllers/oauth/applications_controller.rb
+++ b/app/controllers/oauth/applications_controller.rb
@@ -4,7 +4,7 @@ module Oauth
     before_action :validate_role
 
     def validate_role
-      raise SecurityTransgression if !@user.is_administrator?
+      raise ActionController::RoutingError, 'not found' unless @user.is_administrator?
     end
 
     def index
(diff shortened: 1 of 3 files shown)


app/controllers/oauth/applications_controller.rb
```diff
@@ -1,6 +1,11 @@
 module Oauth
   class ApplicationsController < Doorkeeper::ApplicationsController
     before_action :set_user
+    before_action :validate_role
```
jpslav: Was going to suggest that you rename this and the related method to `before_action :authenticate_admin!` to make it consistent with other places where we do this.  Then I went in the code to check that that was the right method name, and I saw the other places [use this code that can be included](https://github.com/openstax/accounts/blob/master/lib/authenticate_methods.rb#L25-L31)

I _think_ that because this `AuthenticateMethods` module is [already included](https://github.com/openstax/accounts/blob/master/config/initializers/controllers.rb#L10) in `ActionController::Base`, that you can maybe just change this line 4 to `before_action :authenticate_admin!` and delete the `validate_role` method.  I think that should work.  The good thing is your specs should continue to work if I'm right :-)

KarinaMendez2:
@mwharrison ready to merge :)

PR #651, opened 2019-02-13
Restyle untrusted SSO authorization form

RoyEJohnson:
View comes from doorkeeper gem, but styling is added locally, scoped to `.authorizations.new`

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -61,3 +61,30 @@
     padding: 0px 10px;
   }
 }
+
+.authorizations.new {
+    #application-body {
+        background-color: ui-color(white);
+        border-radius: 0.2rem;
+        padding-bottom: $normal-margin;
+
+        .actions {
+            display: flex;
+            flex-direction: row;
+            margin-top: $normal-margin;
+
+            > form {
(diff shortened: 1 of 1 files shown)


RoyEJohnson:
image

app/assets/stylesheets/application.scss
```diff
@@ -61,3 +61,30 @@
     padding: 0px 10px;
   }
 }
+
+.authorizations.new {
+    #application-body {
```
nathanstitt: You might consider adding the `ox-card` style to the view, that way it'll match the other "card type" views and can use the built-in styles https://github.com/openstax/accounts/blob/master/app/assets/stylesheets/controls/_card.scss

Will probably still have to style the buttons a bit though

RoyEJohnson: The view seems to be owned by the doorkeeper gem, as you suggested, so unless I make a local view (and I suspect that's beyond my skills), I can't add a class. If `ox-card` were defined as a placeholder type, though, I could use that.

nathanstitt: ah gotcha.  Yep this makes sense then

PR #654, opened 2019-03-20
Created rack middleware to recode urls with "/accounts" and specs

KarinaMendez2:


spec/features/recode_url_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  
+    context "when called with a GET request" do
+        before(:each) do
+            request.get("/accounts/api/user", 'CONTENT_TYPE' => 'text/plain')
+        end
+        it "changes PATH_INFO" do
+            expect(app["PATH_INFO"]).to eq('text/plain')
+        end
+    end
+end
```
nathanstitt: Yay!  glad `Rack::MockRequest` worked

lib/recode_url.rb
```diff
@@ -0,0 +1,10 @@
+class RecodeUrl
+  def initialize(app, options = {})
+    @app = app
+  end
+  
+  def call(env)
+    env['PATH_INFO'] = env['PATH_INFO'].gsub('/accounts','') if env['REQUEST_METHOD'] == 'GET'
+    @app.call(env)
+  end
+end
```
nathanstitt: you could use the bang version of gsub that modifies in place, but probably should also check that PATH_INFO exists:
```ruby
env['PATH_INFO'].gsub!('/accounts','') if env['REQUEST_METHOD'] == 'GET' && env['PATH_INFO'].present?
```

Also this is completely subjective but I prefer putting the constant on the left so that it'll break if you use a single equal sign instead of accidentally assigning.
```ruby
if 'GET' == env['REQUEST_METHOD']
```

config/application.rb
```diff
@@ -1,7 +1,7 @@
 require File.expand_path('../boot', __FILE__)
 
 require 'rails/all'
-
+    
 # Require the gems listed in Gemfile, including any gems
```
nathanstitt: stray whitespace addition 👻

lib/recode_url.rb
jpslav: +1 on `if constant == variable` for safety

jpslav: I think the gsub line will incorrectly change a path of `/something/accounts/blah` to `/something/blah`.  Can you add a spec that shows this happens?  Then change the gsub command to use a regular expression in the first argument that requires the `/accounts` to be at the start of the path.  rubular.com is a great tool to play with for this btw.

lib/recode_url.rb
```diff
@@ -0,0 +1,10 @@
+class RecodeUrl
```
jpslav: Can we use a more specific class name?  E.g. `RemoveAccountsPathPrefix`.

spec/features/recode_url_spec.rb
jpslav: @KarinaMendez2 please use two spaces for indentation and make sure to use spaces instead of tabs.

spec/features/recode_url_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    subject { described_class.new(app) }
+
+    let(:request) { Rack::MockRequest.new(subject) }
+  
+    context "when called with a GET request" do
+        before(:each) do
+            request.get("/accounts/api/user", 'CONTENT_TYPE' => 'text/plain')
+        end
+        it "changes PATH_INFO" do
+            expect(app["PATH_INFO"]).to eq('text/plain')
```
jpslav: isn't the path info supposed to be equal `/api/user`, not `text/plain`?

spec/features/remove_accounts_path_prefix_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it "should be start to sign up" do
+      expect_any_instance_of(SignupController).to receive(:start)
+      request.get("/accounts/signup")
+    end
+  end
+
+  context "Request through /accounts should only be re routed when applicable" do
+    it "shouldn\'t replace /accounts" do
```
jpslav: add "when not at the start of the path" to the spec description

spec/features/remove_accounts_path_prefix_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      app_object = DummyApp.new
+      expect(app_object).to receive(:call).with({'PATH_INFO' => '/blah/accounts', 'REQUEST_METHOD'=>'GET'})
+      RemoveAccountsPathPrefix.new(app_object).call({'PATH_INFO' => '/blah/accounts', 'REQUEST_METHOD'=>'GET'})
+    end
+
+    it "should replace /accounts" do
```
jpslav: "... when at the start of the path"

spec/features/remove_accounts_path_prefix_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      app_object = DummyApp.new
+      expect(app_object).to receive(:call).with({'PATH_INFO' => '/blah/accounts', 'REQUEST_METHOD'=>'GET'})
+      RemoveAccountsPathPrefix.new(app_object).call({'PATH_INFO' => '/accounts/blah/accounts', 'REQUEST_METHOD'=>'GET'})
+    end
+
+    it "should render image" do
+      app_object = DummyApp.new
+      expect(app_object).to receive(:call).with({'PATH_INFO' => '/assets/bg-login.jpg', 'REQUEST_METHOD'=>'GET'})
+      RemoveAccountsPathPrefix.new(app_object).call({'PATH_INFO' => '/accounts/assets/bg-login.jpg', 'REQUEST_METHOD'=>'GET'})
+      #expect(page).to have_content("img[src*='/assets/bg-login.jpg']")
```
jpslav: you can only expect pages to have content when they go through the real app (and probably also in a feature spec, which this is not).  The DummyApp is sending all requests to a dead end, so won't return anything real, which is in part why this expectation shouldn't work.   Also the JPG content would be binary content, not the img tag you're going for.

I'd just take this commented out expectation out.

spec/features/remove_accounts_path_prefix_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it "should render js" do
+      app_object = DummyApp.new
+      expect(app_object).to receive(:call).with({'PATH_INFO' => '/assets/admin.js', 'REQUEST_METHOD'=>'GET'})
+      RemoveAccountsPathPrefix.new(app_object).call({'PATH_INFO' => '/accounts/assets/admin.js', 'REQUEST_METHOD'=>'GET'})
+      #expect(page).to have_content("(function( global, factory )")
```
jpslav: If you used `app` instead of the dummy, this _might_ work but again the Rails asset pipeline could mean this doesn't work... I dunno and I'm not that worried about it.  I'd leave one of the assets specs in and remove the other one.

PR #655, opened 2019-03-28
Started spec for testing caching headers and fixed translation errors.

KarinaMendez2:


config/application.rb
```diff
@@ -18,6 +18,16 @@ class Application < Rails::Application
...
     # config.i18n.default_locale = :de
+    config.i18n.default_locale = :en
+    config.i18n.fallbacks = {
+      en_us: :en,
+      en_ca: :en,
+      en_uk: :en,
+      fr_lu: :fr,
+      fr_ca: :fr,
+      fr:    :en,
+      es:    :en,
```
jpslav: What's the difference between a default locale and fallbacks?  And why these fallbacks?

spec/features/cache_headers_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    it 'should have headers with public caching assets' do
+      get '/assets/bg-login.jpg'
+      expect(response.headers["Cache-Control"]).to include("public")
+    end
+
+    it 'should have headers with private caching in login' do
+      get '/login'
+      expect(response.headers["Cache-Control"]).to include("max-age=0", "private")
```
jpslav: any idea why `/login` and `/profile` have different cache-control headers?

spec/requests/api/v1/cache_header_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let!(:untrusted_application_token) do
+    FactoryGirl.create :doorkeeper_access_token,
+                       application: untrusted_application,
+                       resource_owner_id: nil
+  end
+  let!(:trusted_application_token) do
+    FactoryGirl.create :doorkeeper_access_token,
+                       application: trusted_application,
+                       resource_owner_id: nil
+  end
```
jpslav: I know I wrote this, but can you remove the unused `let`s?  I think 3 of them can go.

config/application.rb
KarinaMendez2: Locale is for default when there are more than 1 existing translations, fallbacks are for when the translation doesn't exist and falls back to an existing one.

spec/features/cache_headers_spec.rb
KarinaMendez2: I think it might be because we're not mocking a user...

mwharrison:
@KarinaMendez2 - Is this card ready for release?

jpslav:
@mwharrison - I think this card is on me to do something.  I keep snoozing a Slack reminder about it :-) Not ready for release.

mwharrison:
@jpslav - any updates on this?
Do you need anything from BIT?

jpslav:
The returned cache headers are good given that we are setting a minimum TTL of 0 in Cloudfront.

BryanEliDimas:
There appears to be duplicate issues for this PR. I think we're tracking the other one so... I don't know. it's confusing.

PR #659, opened 2019-04-16
Prefix assets and add prefix to hardcoded paths

jpslav:
## DONE

* Updates openstax_path_prefixer gem to a version that supports prefixing assets (precompiles them into an `/accounts/assets` folder and makes references within the JS and CSS use `/accounts` prefixes when referring to fonts, images, etc).  The PR configures this library to only do the asset stuff on production b/c it really only works on production (only place it matters anyway).
* Goes through and changes some hardcoded paths in ERB to path helpers (which get the `/accounts` prefix as needed)
* Changes some of the Javascript hardcoded paths to use the prefix (does this by embedding the prefix in the HTML and then having the JS pull that `data` field out). **There are still other JS URLs to handle and @nathanstitt is going to have to do those b/c they are beyond me :-)**
* You can run the rails server with a `SIMULATE_CLOUDFRONT=true` environment variable and the server will raise an exception if it ever receives a URL without the `/accounts` prefix (similarly Cloudfront will fail because routes without an `/accounts` prefix will never even make it to Accounts).  This is useful for clicking around and making sure we have accounted for all of the routes.  

## TODO

* [x] Adding emails on the profile page uses Javascript which is not yet updated to use the `/accounts` prefix.  We need to modify the `app/assets/javascripts/profile/email.coffee` file to have logic similar to https://github.com/openstax/accounts/pull/659/commits/2f61db2a22411ceadf734c1f19a0a5f9bd35b1c1 -- NB: the `/accounts` prefix isn't added all the time -- only when the page is requested with it! (which is when the `request` has a populated `script_name` attribute)

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -202,7 +202,11 @@ class SessionsController < ApplicationController
       url ||= begin
         referrer_uri = URI(request.referer)
         request_uri = URI(request.url)
-        "#{referrer_uri.scheme}://#{referrer_uri.host}:#{referrer_uri.port}/?#{request_uri.query}"
+        if referrer_uri.host == request_uri.host
+          "#{root_url}?#{request_uri.query}"
+        else
+          "#{referrer_uri.scheme}://#{referrer_uri.host}:#{referrer_uri.port}/?#{request_uri.query}"
+        end
       rescue # in case the referer is bad (see #179)
         root_url
       end
--- a/Gemfile
+++ b/Gemfile
@@ -149,7 +149,7 @@ gem 'scout_apm', '~> 3.0.x'
 gem 'openstax_healthcheck'
 
 # Allow Accounts routes to be accessed under an /accounts prefix (for use in CloudFront)
-gem "openstax_path_prefixer", github: "openstax/path_prefixer", ref: "9d4f40693529"
+gem "openstax_path_prefixer", github: "openstax/path_prefixer", ref: "c74c04fd6c1"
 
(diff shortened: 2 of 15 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -202,7 +202,11 @@ def destroy
       url ||= begin
         referrer_uri = URI(request.referer)
         request_uri = URI(request.url)
-        "#{referrer_uri.scheme}://#{referrer_uri.host}:#{referrer_uri.port}/?#{request_uri.query}"
+        if referrer_uri.host == request_uri.host
+          "#{root_url}?#{request_uri.query}"
+        else
+          "#{referrer_uri.scheme}://#{referrer_uri.host}:#{referrer_uri.port}/?#{request_uri.query}"
+        end
```
jpslav: if Accounts is the referrer, we want to use `root_url` so that we automatically get the path prefixed URL.

nathanstitt: hm, this is kinda dangerous.  `referer` is getting more and more unsupported due to browser security restrictions.  https://stackoverflow.com/questions/6880659/in-what-cases-will-http-referer-be-empty

It does seem like it should be set if it's coming from the same site, but don't know if we can rely on that in all cases.  

However I can't think of any other method to use here 🤔

jpslav: I agree this isn't the best, but at least this PR doesn't change the referrer here (that line was already here), it just adds a new case for when Accounts is the referrer

PR #661, opened 2019-05-03
add docker development environment

TomWoodward:


config/database.yml
```diff
@@ -1,5 +1,6 @@
 default: &default
   adapter: postgresql
+  host: <%= ENV['OXA_DB_HOST'] || nil %>
```
jpslav: The `|| nil` only executes if the `ENV` call returns falsy stuff. If that env var is not defined the env call should return nil, which is falsy. So the change here shouldn’t be needed. Did this change make something work?

TomWoodward: no i just put that in there because others had default values, but i really want it to be unset if you don't pass the param. will remove.

PR #662, opened 2019-05-03
allow login redirects without a host

TomWoodward:
please nitpick my ruby, i don't ruby great.

login redirects without a host will be processed using the accounts host

~if there is concern over using the `referrer` header it would also work for me to default it to the `host` header set by cloudfront. `referrer` is more flexible and i'm _**pretty sure**_ that there is no way to manipulate referrer when actually sending a user somewhere~

### DIFF
--- a/config/initializers/controllers.rb
+++ b/config/initializers/controllers.rb
@@ -71,11 +71,11 @@ ActionController::Base.class_exec do
   end
 
   def save_redirect
-    return true if request.format != :html || request.options?
+    return true if request.format != :html || request.options? || params["r"].blank?
 
-    url = params["r"]
-
-    return true if url.blank? || !Host.trusted?(url)
+    url = Host.default_host(params["r"], request.referer)
+    
+    return true if !Host.trusted?(url)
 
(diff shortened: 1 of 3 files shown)


TomWoodward:
talked with @philschatz, there might be some concern over using the referrer header, we seem unclear on whether the concern is founded on anything.

an alternative that might(?) be cleaner is to simply allow storing a redirect with no host, and then the redirect will process using the accounts host (the `host` header i mentioned in my description). in the alternative it may or may not be confusing for developers to know that redirect with no host will use the accounts host, not the referrer host.

jpslav:
We have had problems with referrer behavior in the past. Memory is fuzzy on
specifics.

What is the underlying goal here? (What isn’t working)

TomWoodward:
@jpslav the actual goal is to be able to render the login link without knowledge of the current hostname, because when rex templates are compiled they don't know what host they will be served on. 

we could compensate with javascript after pageload, but handling redirects with no host would be useful to others i'm sure, and simpler in general for services using accounts login.

PR #664, opened 2019-06-07
Upgrade rails and bootstrap-sass

RoyEJohnson:
Per warnings in
https://github.com/openstax/accounts/network/alerts

Note: no fix yet for omniauth package, the highest severity problem.


BryanEliDimas:
I'm working on the Rails upgrade in a different PR now so this one can be closed. I addressed the security issues separately as well.

PR #669, opened 2019-06-18
Prevent revealing name upon entering email

BryanEliDimas:
Resolve issue https://github.com/openstax/business-intel/issues/358

### DIFF
--- a/app/views/sessions/authenticate.html.erb
+++ b/app/views/sessions/authenticate.html.erb
@@ -1,9 +1,9 @@
 <%
   login_state = get_login_state
-  names = login_state[:names] || []
+  username_or_email = login_state[:username_or_email]
 %>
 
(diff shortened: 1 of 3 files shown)


BryanEliDimas:
Ready for review. It still amazes me how much design there is in Accounts, and it makes it so difficult to choose where is the best place to put things or what to name things. Of course, feel free to suggest any changes at all no matter how small they seem. I mean, for now I committed the smallest possible changes in order to solve the issue at hand, but I could've maybe made things betters. For example, I could've (or I could still) rename `name_greeting` to something else. Email greeting? Sounds deceiving, for lack of a better word. Email or username greeting? Meh. Authenticate greeting? Maybe. But instead of name what do I pass in? `identifier`, so `authenticate_greeting: Hello, %{identifier}!`  Maybe 🤷‍♂.

PR #670, opened 2019-06-18
Fix code coverage reports

BryanEliDimas:
Seems like when I cleaned up the `rails_helper` I introduced an issue which caused code coverage to go down by like 45%. It seemed totally odd and even unrelated so we merged that PR like that anyway. This PR fixes it.

The problem was that codecov/simplecov absolutely must be set up before any other code is even `require`d 🙄  (one reason why the `rails_helper` needed cleaning up).

In this PR I opted to create a `simplecov_helper.rb` file in order for everything to remain clean and for coverage to work well at the same time. I upgraded the gem `simplecov` as well because why not. As for the line

```
config.eager_load = ENV['CI'] == 'true'
```

I believe it helps simplecov to work correctly, according to their documentation:

> If you're using Spring to speed up test suite runs and want to run SimpleCov along with them, you'll find that it often misreports coverage with the default config due to some sort of eager loading issue. 
>
> One solution is to explicitly call eager load in your test_helper.rb / spec_helper.rb after calling SimpleCov.start.
>
> require 'simplecov'
SimpleCov.start 'rails'
Rails.application.eager_load!
Or you could remove gem 'spring' from your Gemfile.

source: https://github.com/colszowka/simplecov#want-to-use-spring-with-simplecov

(Notice that their suggestion is only one option and I did it a little differently.)

**Most importantly, this PR makes codecov go up by 37% which brings it up to 81%.**


BryanEliDimas:
@mwharrison Note that this PR is not connected to an issue (because there is none for it). I don't know if there is a process for this type of PRs without a prior _issue_.

mwharrison:
🎉

mwharrison:
@BryanEliDimas - it's okay if there is no card, the PR creates a card that you can move along.. which is fine for small changes.

PR #671, opened 2019-06-18
Enable cache compression

BryanEliDimas:
Resolve issue https://github.com/openstax/business-intel/issues/147

Note: the default compression threshold for Rails [according to the docs](https://api.rubyonrails.org/v4.2.7/classes/ActiveSupport/Cache/Store.html) is **16K** [bytes], and for Rails 5 it's **1kB** according to the [docs here](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html). The threshold can be configured with this setting: `compress_threshold: 1600`. I didn't specify that, so the default will apply.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -48,7 +48,8 @@ module Accounts
     config.cache_store = :redis_store, {
       url: redis_secrets['url'],
       namespace: redis_secrets['namespaces']['cache'],
-      expires_in: 90.minutes
+      expires_in: 90.minutes,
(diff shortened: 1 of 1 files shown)


BryanEliDimas:
@mwharrison 😄 Yes, hoping that does the job! If anything, it's _possible_ that it'll need fine-tuning in the future. In that case all we'd have to change is add `compress_threshold`, so also would be an easy fix.

PR #673, opened 2019-06-25
Replace factory_girl with factory_bot

BryanEliDimas:
The gem has been renamed. This PR upgrades to the gem with the new name. Also, this PR fixes flaky tests in Travis.

Notice that the diff shows that I added as many lines of code as I removed, and the only one exception is this line: https://github.com/openstax/accounts/pull/673/commits/509a434f3f0fcbd4c22ad78804cb7a382145d9a1#diff-8212d962e93a22af7645b80bb901fca8R5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -173,7 +173,7 @@ group :development, :test do
   gem 'rspec-rails', '~> 3.5'
 
   # Fixture replacement
-  gem 'factory_girl_rails'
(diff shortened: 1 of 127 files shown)


BryanEliDimas:
Closing this PR because it's not necessary to separate these changes into a branch with PR. I'll keep them as commits instead.

PR #682, opened 2019-08-09
Fix reset password loop – Do not ask to re-authenticate when forgot password

BryanEliDimas:
Fixes issue https://github.com/openstax/business-intel/issues/550

### DIFF
--- a/spec/features/reset_password_spec.rb
+++ b/spec/features/reset_password_spec.rb
@@ -11,6 +11,7 @@ feature 'Password reset', js: true do
     # issue: https://github.com/openstax/business-intel/issues/550
     user = create_user 'user'
     create_email_address_for user, 'user@example.com'
+    login_token = generate_login_token_for 'user'
     log_in('user','password')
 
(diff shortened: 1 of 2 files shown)


spec/features/reset_password_spec.rb
```diff
@@ -2,10 +2,30 @@
...
+    Timecop.freeze(Time.now + RequireRecentSignin::REAUTHENTICATE_AFTER) do
+      find('[data-provider=identity] .edit').click
+      expect(page).to have_content(t :"sessions.reauthenticate.page_heading")
+      click_link(t :"sessions.authenticate_options.reset_password")
+      expect(page).to have_content(t(:'identities.send_reset.we_sent_email', emails: 'user@example.com'))
+
+      open_email('user@example.com')
+      password_reset_link = get_path_from_absolute_link(current_email, 'a')
+      visit password_reset_link
+      expect(page).not_to(have_current_path(reauthenticate_path))
```
Dantemss: Would be nice to also assert that you are in the path/url that renders the "reset password" form.

BryanEliDimas: Added ✅

PR #683, opened 2019-08-14
Suggest email domain on suspected mistype

BryanEliDimas:
Addresses https://github.com/openstax/business-intel/issues/471

Also fixes a test that I believe was wrong: `failure because suspected non-school email then success`

### DIFF
--- a/config/locales/pl.yml
+++ b/config/locales/pl.yml
@@ -405,7 +405,7 @@ pl:
         Czy to na pewno Twój szkolny adres-email? Jeżeli tak, wciśnij %{button} poniżej.
         Jeżeli nie, wpisz adres e-mail przyznany Ci przez Twoją szkołę.
       did_you_mean:
-        Czy to literówka?
+        Czy to literówka
       proceed_with_likely_mistyped_email:
         Jeśli tak, popraw adres e-mail. Jeśli nie ma błędu, kliknij %{button}.
 
--- a/config/locales/pl.yml
+++ b/config/locales/pl.yml
@@ -404,6 +404,10 @@ pl:
       teacher_school_email_warning: >-
         Czy to na pewno Twój szkolny adres-email? Jeżeli tak, wciśnij %{button} poniżej.
         Jeżeli nie, wpisz adres e-mail przyznany Ci przez Twoją szkołę.
+      did_you_mean:
+        TODO Did you mean
+      proceed_with_likely_mistyped_email:
+        TODO If yes, please edit your email. If not, click %{button} below.
 
     verify_email:
       page_heading_pin: Sprawdź skrzynkę e-mail
--- a/config/locales/pl.yml
+++ b/config/locales/pl.yml
@@ -405,9 +405,9 @@ pl:
         Czy to na pewno Twój szkolny adres-email? Jeżeli tak, wciśnij %{button} poniżej.
         Jeżeli nie, wpisz adres e-mail przyznany Ci przez Twoją szkołę.
       did_you_mean:
-        TODO Did you mean
+        Czy to literówka?
       proceed_with_likely_mistyped_email:
-        TODO If yes, please edit your email. If not, click %{button} below.
+        Jeśli tak, popraw adres e-mail. Jeśli nie ma błędu, kliknij %{button}.
 
     verify_email:
       page_heading_pin: Sprawdź skrzynkę e-mail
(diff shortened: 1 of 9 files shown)


Dantemss:
Not a huge fan of hardcoding a bunch of email domains in the JS. Wouldn't it be better to make an API that returns a list of common emails from the DB?

We also warn people if they use non-edu addresses etc.

spec/features/signed_login_flow_spec.rb
```diff
@@ -178,16 +178,16 @@
 
     it 'requires email validation when edited' do
       arrive_from_app(params: signed_params, do_expect: false)
-      fill_in (t :"signup.start.email_placeholder"), with: 'test-modified@test.com'
+      fill_in (t :"signup.start.email_placeholder"), with: 'test-modified@example.com'
```
BryanEliDimas: Reviewer can pretty much ignore this type of change. `mailcheck`er was showing the likely mistype warning for the domain `@test.com`. Should be no big deal.

BryanEliDimas:
@Dantemss Yes, I agree. The question is, should I do that now, in this PR?

Dantemss:
I would vote yes but I'm not in BIT so I don't vote... Your pm/pdm has to decide.

mwharrison:
@Dantemss, are you talking about the domains in mailcheck.js?

BryanEliDimas:
@mwharrison Yeah, I think so. I mean, my hands hurt while coding that ugly baby but I'm expecting that I'll be forced to do *a lot* of refactoring while working on the big redesign (plus code freeze ~is~ was this Friday) and I relented.

🤷‍♂

mwharrison:
@BryanEliDimas - are you using this? https://github.com/mailcheck/mailcheck/

BryanEliDimas:
Actually, yes, Dante is referring to the copy and paste I did in `app/assets/javascripts/signup/email-value.coffee` not in mailcheck.js (because that is source code directly ported over to `accounts`)

@mwharrison Yes, I'm using that lib.

BryanEliDimas:
That readme has a link to the JS version: https://raw.githubusercontent.com/mailcheck/mailcheck/d25dc9a119ca844bb35b1baf341cca0a634e4ac9/src/mailcheck.js

mwharrison:
Gotcha, is it possible to import the domain list from the library in `email-value.coffee`?

BryanEliDimas:
I don't think so. I think Dante is suggesting to create an API endpoint 😓, which to me sounds like a lot of work for relatively minor gains.

BryanEliDimas:
Actually, I've just realized that I shouldn't have hardcoded the exact same email domains in the JS/coffee script because the most popular domains are `.edu` and we don't even get to that `mailcheck`ing code in that case (see `IS_EDU.test(@email.val())`).

So with the DRY rule out of the question, I think there's nothing wrong at all with hardcoding the most popular email addresses at this time. The current top domains will probably remain the top domains in the foreseeable future, probably only new ones will come to life and at that point all we'd have to do is make that tiny change in our code. Creating an API endpoint is really unnecessary.

As for the hardcoded (whitelisted) domains in the Ruby code, it is not that difficult to improve it and probably worth the effort because it'll only cost an extra DB query, but I know _that_ is definitely outside the scope of this PR :).

Dantemss:
Hmm looks better. But why not just leave the `domain` argument out? It defaults to `Mailcheck.defaultDomains`.

BryanEliDimas:
@Dantemss Main reason is one feature test entered a `@myspace.com` email and `emailcheck`er detected it as likely a `@me.com` email which led me to believe that it's going to detect every email not in its list as likely one of the default ones :/. I'd rather only pass in the ones which **we** believe is likely a typo.

BryanEliDimas:
`@mac.com` email address aren't even in use anymore. Apple replaced them with `@icloud.com`.

Dantemss:
I think if it transforms myspace.com to mac.com, then the max edit distance is too high. Lowering it should fix this, I hope.

BryanEliDimas:
How would I do that? I have an idea from looking at the source code but it's JS and there's nothing about that in the docs/readme.

BryanEliDimas:
I just left the defaults for now. In that case, this PR is ready for code review.

BryanEliDimas:
I just left the defaults for now. In that case, this PR is ready for code review. @Dantemss @mwharrison

config/locales/pl.yml
```diff
@@ -404,6 +404,10 @@ pl:
       teacher_school_email_warning: >-
         Czy to na pewno Twój szkolny adres-email? Jeżeli tak, wciśnij %{button} poniżej.
         Jeżeli nie, wpisz adres e-mail przyznany Ci przez Twoją szkołę.
+      did_you_mean:
+        TODO Did you mean
```
mwharrison: Can we get the correct translations in here?

BryanEliDimas: Done. Please review the updated text.

PR #685, opened 2019-08-16
Rails 5 SSO

Dantemss:
I can rebase again after the Rails 5 PR passes Travis to fix the errors (or you can just merge this into your branch any time you want @BryanEliDimas).


config/application.rb
```diff
@@ -1,5 +1,6 @@
 require_relative 'boot'
 require 'rails/all'
+require 'rails/test_unit/railtie'
```
Dantemss: I think this is not necessary actually... let me try removing it.

PR #686, opened 2019-08-19
Create csv export of last login dates with emails

BryanEliDimas:
Based off of app/routines/export_users_info_to_match_with_consent_forms.rb, its task, and its tests.

### DIFF
--- a/app/routines/export_users_last_login_date.rb
+++ b/app/routines/export_users_last_login_date.rb
@@ -16,7 +16,8 @@ class ExportUsersLastLoginDate
       output_users << Hashie::Mash.new({
         emails: user.contact_infos.map(&:value).join(", "),
         last_login_at: SecurityLog.sign_in_successful
-                                    .where(user_id: user.id).try(:last).try(:created_at).try(:strftime, "%m/%d/%Y %I:%M%p %Z")
+                                    .where(user_id: user.id)
+                                    .first&.created_at&.strftime("%m/%d/%Y %I:%M%p %Z")
       })
     end
     output_users
@@ -43,7 +44,7 @@ class ExportUsersLastLoginDate
   end
 
   def exports_folder
-    File.join 'tmp'
+    'tmp'
   end
 
   def remove_exported_files
--- a/app/routines/export_users_last_login_date.rb
+++ b/app/routines/export_users_last_login_date.rb
@@ -14,7 +14,7 @@ class ExportUsersLastLoginDate
     output_users = []
     User.preload(:contact_infos).find_each do |user|
       output_users << Hashie::Mash.new({
-        emails: user.contact_infos.map(&:value).join(", "),
+        emails: user.contact_infos.verified.map(&:value).join(", "),
         last_login_at: SecurityLog.sign_in_successful
                                     .where(user_id: user.id)
                                     .first&.created_at&.strftime("%m/%d/%Y %I:%M%p %Z")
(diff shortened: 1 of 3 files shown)


BryanEliDimas:
Not the cleanest code but it's an urgent task so as long as _it works_.

app/routines/export_users_last_login_date.rb
```diff
@@ -0,0 +1,52 @@
...
+    remove_exported_files if delete_after
+  end
+
+  def info
+    output_users = []
+    User.preload(:contact_infos).find_each do |user|
+      output_users << Hashie::Mash.new({
+        emails: user.contact_infos.map(&:value).join(", "),
+        last_login_at: SecurityLog.sign_in_successful
+                                    .where(user_id: user.id).try(:last).try(:created_at).try(:strftime, "%m/%d/%Y %I:%M%p %Z")
```
Dantemss: I suggest using `&.`: `&.last&.created_at&.strftime(...)` etc.
Also I think you want `first` instead of `last` because the `default_scope` for SecurityLog is `created_at DESC`.

app/routines/export_users_last_login_date.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+  end
+
+  def filename
+    File.join exports_folder, "users_last_login_date.csv"
+  end
+
+  def exports_folder
+    File.join 'tmp'
```
Dantemss: `join` with 1 argument is odd. Maybe just `tmp`.

app/routines/export_users_last_login_date.rb
```diff
@@ -0,0 +1,52 @@
+class ExportUsersLastLoginDate
+  lev_routine
+
+  protected
+
+  def exec(delete_after: true)
```
Dantemss: Instead of `delete_after: true`, maybe make this `write_csv: false` (or something similar, just don't give it the same name as the method). Skip generating the CSV if set to false. Then you can also remove the delete part.

app/routines/export_users_last_login_date.rb
BryanEliDimas: Copy & pasta 🍝 . Fixed.

app/routines/export_users_last_login_date.rb
BryanEliDimas: **Great** catch with the ordering!! I didn't think too hard about that.

BryanEliDimas: I totally did not know about that Safe Navigation Operator_. 💯 Fixed.

app/routines/export_users_last_login_date.rb
```diff
@@ -0,0 +1,53 @@
...
+    outputs[:info] = info
+    generate_csv
+    remove_exported_files if delete_after
+  end
+
+  def info
+    output_users = []
+    User.preload(:contact_infos).find_each do |user|
+      output_users << Hashie::Mash.new({
+        emails: user.contact_infos.map(&:value).join(", "),
```
jpslav: IIRC this will include unverified emails, which will not be what we want.

BryanEliDimas: Fixed :)

app/routines/export_users_last_login_date.rb
BryanEliDimas: I think we want to also test the csv generation, so we add an expectation for `remove_exported_files` to get called and stuff. So, leaving it as it is to play it safe although it seems like a great refactor.

PR #697, opened 2019-09-04
update lev to 9.0.3 which supports ActiveRecord 5x

nathanstitt:
the monkey patch is no longer needed since https://github.com/lml/lev/pull/67


BryanEliDimas:
* correction to staxly bot: This PR landed in v2.3.2 because I made a mistake and upped the minor version instead. Renamed the release + tag.

PR #699, opened 2019-09-11
Make SSO cookie authoritative for current user; share session cookie across domains

jpslav:
* Make the SSO cookie the one place to look for (and set) the current user
* Make the Accounts session cookie shared across domains so that OX sites that proxy Accounts can make use of it
* Make the Accounts session cookie name environment-specific so that cookies from different environments don't clobber each other.

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,5 +1,14 @@
 # Be sure to restart your server when you modify this file.
 
 # Permanent cookie jar also uses 20 years
-Rails.application.config.session_store :cookie_store, key: '_accounts_session',
-                                                      expire_after: 20.years
+
+class SessionStoreCookieName
+  def self.to_s
+    environment_name = Rails.application.secrets.environment_name
+    "_accounts_session#{'_' + environment_name if environment_name != 'prodtutor'}"
+  end
+end
+
+Rails.application.config.session_store :cookie_store, key: SessionStoreCookieName.to_s,
+                                                      expire_after: 20.years,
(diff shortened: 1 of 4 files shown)


config/secrets.yml.example
```diff
@@ -148,7 +148,7 @@ test:
       name: ox
       options:
         domain: ''
-        secure: true
+        secure: false
```
jpslav: because the test environment doesn't use HTTPS

config/initializers/session_store.rb
```diff
@@ -1,5 +1,14 @@
...
 # Permanent cookie jar also uses 20 years
-Rails.application.config.session_store :cookie_store, key: '_accounts_session',
-                                                      expire_after: 20.years
+
+class SessionStoreCookieName
+  def self.to_s
+    environment_name = Rails.application.secrets.environment_name
+    "_accounts_session#{'_' + environment_name if environment_name != 'prodtutor'}"
+  end
+end
```
jpslav: split this out so I could write a spec

config/initializers/session_store.rb
```diff
@@ -1,5 +1,14 @@
...
+
+Rails.application.config.session_store :cookie_store, key: SessionStoreCookieName.to_s,
+                                                      expire_after: 20.years,
+                                                      domain: :all
```
Dantemss: `domain: :all` will need some QA testing, because I remember looking into this for the SSO cookie and it not working properly in all cases. But I don't remember what was wrong.

PR #701, opened 2019-09-11
Export user uuid instead of email address

BryanEliDimas:
For security reasons, we've decided that it's better to not include Personally Identifiable Information in a CSV (namely, email addresses). Instead, we can work with Unique User IDs.


BryanEliDimas:
@mwharrison Ready for code review. Don't know who else can or should review.

PR #704, opened 2019-09-15
Finish removing Doorkeeper::Application.trusted

jpslav:
Followup to #700 that completes the removal of Doorkeeper::Application.trusted


jpslav:
I have no idea what's up with the `codecov/patch` check

PR #709, opened 2019-09-26
Remove domain from sso cookie on dev and test

reedstrm:
Sending a cookie with "domain=''" is undefined behavior, according to the spec. Browsers treat it as no domain attribute (which the spec says it SHOULD do). Unfortunately, the python requests library chooses to ignore the cookie completely.  Instead, don't set it at all if it's not defined in the environment.

### DIFF
--- a/config/secrets.yml.example
+++ b/config/secrets.yml.example
@@ -80,7 +80,9 @@ development:
     cookie:
       name: <%= ENV["SSO_COOKIE_NAME"] || 'ox' %>
       options:
-        domain: <%= ENV["SSO_COOKIE_DOMAIN"] || "''" %>
+        <% if ENV.key?("SSO_COOKIE_DOMAIN") %>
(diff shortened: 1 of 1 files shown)


Dantemss:
Some weird spec is failing (GDPR stuff?)

config/secrets.yml.example
```diff
@@ -147,7 +147,7 @@ test:
     cookie:
       name: ox
       options:
-        domain: ''
+        domain: :all
         secure: false
```
nathanstitt: I think you don't want the leading `:` on all.  Since this is a yaml file you can't use ruby symbols.  Maybe it's better to just let domain be unset here which would be `nil` when it's read in?

reedstrm: Ah, I mindlessly copied the change I'd made in the ERB context for development, into the static test context. Will fix.

reedstrm: ok changed it to make the domain attribute completely conditional - don't set it if it's not requested.

PR #710, opened 2019-10-01
Add adoptions

edwoodward:
Adding adoption status from Salesforce to user model and user API

This is dependent on [PR #20](https://github.com/openstax/openstax_salesforce/pull/20) on openstax-salesforce repo


app/representers/api/v1/user_representer.rb
```diff
@@ -72,7 +72,7 @@ class UserRepresenter < Roar::Decorator
              }
 
     property :salesforce_contact_id,
-             if: ->(user_options:, **) { user_options.try(:fetch, :include_private_data, false) },
+             if: ->(user_options:, **) { user_options.try(:fetch, :include_private_data, false) }
```
BryanEliDimas: Missing a comma, this is going to break.

app/routines/update_user_salesforce_info.rb
```diff
@@ -113,6 +113,8 @@ def call
               :pending_faculty
             end
 
+          user.using_openstax = leads.map[&:adoption_status].uniq
```
BryanEliDimas: `leads.map[&:adoption_status].uniq` returns an array, but ` user.using_openstax` needs a string.

BryanEliDimas: I really don't know which lead you should store in/as `using_openstax`. I _think_ it would something similar as we're doing with `user.faculty_status` right above.

BryanEliDimas:
First tried to resolve the git conflict via GH's UI but it didn't look too good, so I rebased master into this branch locally and force-pushed.

app/routines/update_user_salesforce_info.rb
```diff
@@ -225,7 +225,7 @@ def cache_contact_data_in_user!(contact, user)
         :other_school_type
       end
 
-      if contact.adoption_status != nil
+      unless contact.adoption_status.nil? || contact.adoption_status.empty?
```
BryanEliDimas: You can call `.blank?` instead of `.nil?` and `.empty?`

BryanEliDimas: I haven't done a complete review, just saw this and thought I should mention it sooner rather than later.

edwoodward: I didn't think I could use `.blank?` because contact is not an ActiveRecord object. It is an object from the openstax-salesforce Gem.

BryanEliDimas: But you're checking `adoption_status` not `contact`, so it doesn't matter what kind of object `contact` is :).

db/migrate/20190930192109_add_using_openstax_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddUsingOpenStaxToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :using_openstax, :string
```
BryanEliDimas: I think `using_openstax` should be a boolean instead of a string that stores "Yes" or "No".

app/routines/update_user_salesforce_info.rb
edwoodward: So `.blank` is available on any field on any object? I didn't know that. I read it was only for ActiveRecords. I can make that change.

db/migrate/20190930192109_add_using_openstax_to_user.rb
edwoodward: If I made it a boolean, it would need to have a default value. Dante suggested I not have a default value so it only shows up in the API when data is available.

BryanEliDimas: By this "it only shows up in the API when data is available" — you mean the field `using_openstax` entirely, or the value of the field so when no data is available its value is `nil`?

Because if you mean the value of the field only shows up when data is available, then the value in that case could be `nil`, and if you mean the field `using_openstax` itself, then that's weird that it only shows some times. Keep in mind that `nil` ("no data available") can be represented as `null` in JSON.

BryanEliDimas: ** I see that you do mean the field's value.

So yeah, I think we can still make it a boolean and represent "No data available`, `nil` in Ruby, as `null` in JSON.

app/routines/update_user_salesforce_info.rb
BryanEliDimas: 👍 Rails adds it to every object. 
https://apidock.com/rails/Object/blank%3f

PR #713, opened 2019-10-07
Create rake task for sending PIN confirmation emails

BryanEliDimas:
Used for testing the deliverability of PIN confirmation emails.

Creates the rake command `rake send_pin_confirmation_emails`

It can be run this way:

`EMAILS_CSV_PATH='emails_list.csv' rake send_pin_confirmation_emails`

Needs a CSV file with only one column and no headers, for example: 

[emails_list.csv](#)
```
bryan@me.com
eli@me.com
bryaneli@me.com
```

### DIFF
--- a/app/mailers/signup_confirmation_mailer.rb
+++ b/app/mailers/signup_confirmation_mailer.rb
@@ -11,4 +11,16 @@ class SignupConfirmationMailer < ApplicationMailer
 
     pre_auth_state.update_column(:confirmation_sent_at, Time.now)
   end
+
+# primarily used for testing the deliverability of the PIN confirmation emails
+  def pin_confirmation_email(email_address:, confirmation_pin:, confirmation_token:, user_full_name:)
+    @email_address = email_address
+    @user_full_name = user_full_name
+    @confirmation_pin = confirmation_pin
+    @confirmation_token = confirmation_token
+
+    mail(to: "\"#{@user_full_name}\" <#{@email_address}>",
+      subject: "Use PIN #{@confirmation_pin} to confirm your email address"
+    )
+  end
 end
--- a/app/views/signup_confirmation_mailer/pin_confirmation_email.html.erb
+++ b/app/views/signup_confirmation_mailer/pin_confirmation_email.html.erb
@@ -1,0 +1,18 @@
+<p>Welcome!</p>
+
+  <p>Enter your 6-digit PIN in your browser to confirm your email address:</p>
+
+  <p>Your PIN: <b><%= @confirmation_pin %></b></p>
+
+  <p>If you have any trouble using this PIN, you can also click the link below to confirm your email address:</p>
+
+<%
+  token_url = signup_verify_by_token_url(code: @confirmation_token)
+%>
+
+<p><%= link_to token_url, token_url %></p>
+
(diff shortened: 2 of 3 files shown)


lib/tasks/test_pin_email_deliverability.rake
```diff
@@ -0,0 +1,17 @@
...
+desc "Send PIN confirmation emails to test their deliverability, needs EMAILS_CSV_PATH var"
+task :test_pin_email_deliverability => [:environment] do
+  CSV.foreach(ENV['EMAILS_CSV_PATH']) do |row|
+
+    SignupConfirmationMailer.for_testing_deliverability(
+        email_address: row[0],
+        confirmation_pin: TokenMaker.contact_info_confirmation_pin,
+        confirmation_token: TokenMaker.contact_info_confirmation_code,
+        user_full_name: Faker::Name.name
+    )
```
Dantemss: Something (either this task or the method) needs to call either `deliver_now` or `deliver_later` on the output of `mail`.

BryanEliDimas: DOH. How did I miss that. Fixed.

BryanEliDimas:
> re: Rather than making a new mailer and template, couldn't we just make the task reuse the existing ConfirmationMailer?

No, because that requires a `ContactInfo` for each email and each one requires a `User` and stuff. 

Addressed your comment, force-pushed the changes. Please review again with the changes?

Dantemss:
In that case I would suggest that inside a transaction we create the new Users/ContactInfos and use the real mailer, then ensure we rollback the transaction at the end. The goal is to minimize the amount of test-specific code so the test can be closer to the real thing.

But maybe @jpslav has other opinions?

BryanEliDimas:
Supa smart @Dantemss. With that suggestion, it turns out that we _will_ be able to run it directly through SSH instead of the release process! 🚀 

[EMAILS_CSV_PATH='emails_list.csv' bin/rails console](#)
irb >
```
CSV.foreach(ENV['EMAILS_CSV_PATH']) { |row|
  PreAuthState.transaction {
    pas = PreAuthState.create(is_partial_info_allowed: true, contact_info_value: row[0].try(:strip), role: 'student');
    puts SignupConfirmationMailer.instructions(pre_auth_state: pas).deliver_now;
    raise ActiveRecord::Rollback
  }
}
```

Dantemss:
Yes but test on staging etc first 😬This is a little scary to run directly on prod.

BryanEliDimas:
Closing because we ended up running the code I typed above directly via the rails console and it worked.

BryanEliDimas:
@scb6 New code because the analysts need all email to have the same subject line:

```
CSV.foreach(ENV['EMAILS_CSV_PATH']) { |row|
  PreAuthState.transaction {
    pas = PreAuthState.create(is_partial_info_allowed: true, contact_info_value: row[0].try(:strip), role: 'student');
    pas.update_attributes(confirmation_pin: 122608);
    puts(SignupConfirmationMailer.instructions(pre_auth_state: pas).deliver_now);
    raise ActiveRecord::Rollback
  }
}
```

Note that the PIN code for all email will be: 122608

PR #715, opened 2019-10-24
Make the new login and signup flows functional

BryanEliDimas:
Makes the new login flow (and signup flow, as a bonus) functional. Essentially, this PR is the backend portion for 90% or so of the new student login flow. Also includes code for sign up but it's not yet tested so that's WIP and will be finished in a different PR.


BryanEliDimas:
Travis is failing because I added a `before_save` callback to the ContactInfo in order to generate the confirmation code on creation. I don't think that it was really important for that field to empty on creation, so I plan on updating that expectation in the tests in the future. Besides that, I'm sending this PR up for code review.

mwharrison:
approving even though Travis is failing since this is not going to master yet

PR #717, opened 2019-11-05
Wrap up log in with facebook

BryanEliDimas:
Primarily adds the ability to login & signup with facebook. Also wrote tests and stuff. There's stuff left todo for signup with facebook so I'll open up a new PR to finalize that.


BryanEliDimas:
@mwharrison This is ready for review.

PR #726, opened 2019-11-15
New accounts flow / Add the ability to login with google

BryanEliDimas:
My implementation strategy is to create a new `omniauth` strategy called `googlenewflow` that works the same way as our `google` strategy but redirects instead to the newflow controller's callback action. 

I found that this is much easier and straightforward than the alternative—which, if I recall correctly, is to _somehow_ detect that the request came from the `NewflowController` and redirect back to the correct action.


BryanEliDimas:
@mwharrison It's working in accounts-dev.openstax.org/i/login now.

PR #736, opened 2019-12-03
Asymmetric SSO cookies

Dantemss:
Leaving the old SSO code in there for now but we should remove it once we update the other apps.

Needs https://github.com/openstax/tutor-deployment/pull/529


lib/user_session_management.rb
```diff
@@ -39,16 +42,17 @@ def sign_in!(user, security_log_data: {})
 
     if @current_user.is_anonymous?
       # Clear the SSO cookie
-      sso_cookies.delete(sso_cookie_name)
+      sso_cookie_jar.delete
+      old_sso_cookie_jar.delete('ox')
```
jpslav: ditto above

lib/user_session_management.rb
```diff
@@ -39,16 +42,17 @@ def sign_in!(user, security_log_data: {})
...
 
       security_log :sign_out, security_log_data
     else
       session[:last_admin_activity] = DateTime.now.to_s if @current_user.is_administrator?
 
       # Set the SSO cookie
-      sso_cookies.permanent[sso_cookie_name] = {
-        value: { 'user' => Api::V1::UserRepresenter.new(@current_user).to_hash }
-      }
+      cookie = { value: { 'user' => Api::V1::UserRepresenter.new(@current_user).to_hash } }
```
jpslav: Possible to go ahead and add faculty status here?  Will help other efforts.  I think it is not included in the UserRepresenter by default.

lib/user_session_management.rb
```diff
@@ -39,16 +42,17 @@ def sign_in!(user, security_log_data: {})
...
+      sso_cookie_jar.cookie = cookie
+      old_sso_cookie_jar.permanent['ox'] = cookie
```
jpslav: ditto above

lib/user_session_management.rb
Dantemss: I think I have to pass in include_private_data

lib/user_session_management.rb
```diff
@@ -1,32 +1,35 @@
...
+    cookie_jar.sso
   end
 
   # Always return an object
   def current_user
-    # The SSO cookie has sole responsbility for managing the current user
+    # The SSO cookie has sole responsibility for managing the current user
     @current_user ||= begin
-      sso_cookie = sso_cookies[sso_cookie_name]
+      sso_cookie = sso_cookie_jar.cookie || old_sso_cookie_jar['ox']
```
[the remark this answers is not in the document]
Dantemss: OK I'll leave that in the secrets

lib/user_session_management.rb
jpslav: Kind of don’t want other data in there eg all emails. I can look at this later. Would be cool if we could just have fac status. Would let other apps enable things without callback to accounts. Maybe faculty status isn’t really private?

jpslav: Kind of don’t want other data in there eg all emails. I can look at this later. Would be cool if we could just have fac status. Would let other apps enable things without callback to accounts. Maybe faculty status isn’t really private?

Dantemss: Yeah maybe faculty status is not private (being faculty somewhere is pretty much public knowledge anyway).

Dantemss:
The plan is to sign the cookie with the signature private key (readable with the signature public key) then encrypt it with the encryption public key (readable with the encryption private key).

Or we could just do asymmetric signature with symmetric encryption, since the public key can be generated from the private key anyway...

Dantemss:
Sure, but Travis exploded. I think it's just intermittent stuff, so I'm rerunning

PR #745, opened 2019-12-20
Added adoption status to user

edwoodward:
Data comes from Salesforce and is available via the user API

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -88,10 +88,6 @@ class User < ActiveRecord::Base
     !!is_test
   end
 
-  def using_openstax?
-    false
-  end
-
   def is_anonymous?
     false
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -88,6 +88,10 @@ class User < ActiveRecord::Base
     !!is_test
   end
 
+  def using_openstax?
+    false
+  end
+
   def is_anonymous?
     false
   end
--- a/app/routines/update_user_salesforce_info.rb
+++ b/app/routines/update_user_salesforce_info.rb
@@ -5,6 +5,13 @@ class UpdateUserSalesforceInfo
     'Career School/For-Profit (2)'
   ]
 
+  ADOPTION_STATUSES = {
+    "Current Adopter" => true,
+    "Future Adopter" => true,
+    "Past Adopter" => false,
+    "Not Adopter" => false
+  }
+
   def initialize(allow_error_email:)
     @allow_error_email = allow_error_email
     @errors = []
@@ -214,6 +221,10 @@ class UpdateUserSalesforceInfo
         :unknown_school_type
       else
         :other_school_type
+      end
+      
+      unless contact.adoption_status.blank?
+        user.using_openstax = ADOPTION_STATUSES[contact.adoption_status]
       end
     end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_12_04_193412) do
+ActiveRecord::Schema.define(version: 2019_12_18_192109) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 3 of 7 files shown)


app/models/user.rb
```diff
@@ -88,6 +88,10 @@ def is_test?
     !!is_test
   end
 
+  def using_openstax?
+    false
+  end
+
```
BryanEliDimas: Defining this is not necessary because, I believe, Rails dynamically defines it the moment you add the database column `using_openstax` to the users table. It actually might be problematic if you define it here as it will always be false regardless of the actual value in the database for `using_openstax`.

app/routines/update_user_salesforce_info.rb
```diff
@@ -5,6 +5,13 @@ class UpdateUserSalesforceInfo
     'Career School/For-Profit (2)'
   ]
 
+  ADOPTION_STATUSES = {
+    "Current Adopter" => true,
+    "Future Adopter" => true,
+    "Past Adopter" => false,
+    "Not Adopter" => false
+  }
+
```
BryanEliDimas: 🤔 How do I know that these exact strings are how they will come always come from Salesforce? If they are correct today, case-sensitive, separated by a space, what if they change in the future? I think there should be an integration test for this, at least; that would appease my concerns.

db/schema.rb
```diff
@@ -247,10 +247,10 @@
...
     t.string "lead_application_source", default: "", null: false
     t.boolean "confidential", default: true, null: false
-    t.boolean "can_access_private_user_data", default: false, null: false
-    t.boolean "can_find_or_create_accounts", default: false, null: false
-    t.boolean "can_message_users", default: false, null: false
-    t.boolean "can_skip_oauth_screen", default: false, null: false
+    t.boolean "can_access_private_user_data", default: false
+    t.boolean "can_find_or_create_accounts", default: false
+    t.boolean "can_message_users", default: false
+    t.boolean "can_skip_oauth_screen", default: false
```
BryanEliDimas: I don't know why the `null: false` went away. I think this is not correct. You might've ran a migration at one point which modified your local database, and then switched back to your branch, ran another migration and the changes from the other branch remained? Maybe? I think you can fix this by simply reverting these lines in your schema.

db/schema.rb
```diff
@@ -333,6 +333,7 @@
     t.citext "support_identifier", null: false
     t.boolean "is_test"
     t.integer "school_type", default: 0, null: false
+    t.boolean "using_openstax", default: false
```
BryanEliDimas: I was going to suggest that you rename this column to `is_usinng_openstax` because that is the Rails convention but... I see that `using_openstax` is already being used all throughout the codebase (I don't know why) so... I'll leave it up to you to decide.

app/models/user.rb
edwoodward: @BryanEliDimas I'm removing it, but just for my understanding, why do other boolean values in that model have the function defined, but this one should not?

BryanEliDimas: None of the methods correspond to a database field by the same name without the `?` so we must define them, except for `is_test`, in which case I don't know the reason why 🤔. It looks like the intent was to cast it into a boolean (because of the double bang `!!`) but it's already a boolean in the database so in that case, defining that method is unnecessary but **not** problematic because it gets the value from the DB instead of a hardcoded `true` or `false`.

BryanEliDimas: We could (and should attempt to) remove `def is_test?`, and the tests should still pass. If they don't, then I must be wrong **or** it doesn't work because of some corner case with a gem.

edwoodward: Removed

app/routines/update_user_salesforce_info.rb
edwoodward: Tests added

db/schema.rb
edwoodward: Fixed

db/schema.rb
edwoodward: Leaving it as is.

PR #746, opened 2020-01-07
populate missing new SSO cookie

jpslav:
Users who last logged in prior to the release of the new SSO cookie code will not have it.  This PR checks whenever a user uses Accounts to see if the SSO cookie is missing and if so adds it.

### DIFF
--- a/lib/user_session_management.rb
+++ b/lib/user_session_management.rb
@@ -35,7 +35,14 @@ module UserSessionManagement
     @current_user ||= begin
       user_hash = sso_cookie_jar.subject || old_sso_cookie_jar[old_sso_cookie_name]&.[]('user')
       current_sso_user = User.find_by(uuid: user_hash['uuid']) if user_hash.present?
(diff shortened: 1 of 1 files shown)


lib/user_session_management.rb
```diff
@@ -35,7 +35,14 @@ def current_user
     @current_user ||= begin
       user_hash = sso_cookie_jar.subject || old_sso_cookie_jar[old_sso_cookie_name]&.[]('user')
       current_sso_user = User.find_by(uuid: user_hash['uuid']) if user_hash.present?
-      current_sso_user.nil? ? AnonymousUser.instance : current_sso_user
+
+      if current_sso_user.nil?
+        AnonymousUser.instance
+      else
+        # Some users may not have the new SSO cookie yet, so set it
+        sso_cookie_jar.subject = user_hash if sso_cookie_jar.subject.blank?
```
jpslav: This is the only new line.  The other “new” lines are just restating the `... ? ... : ...` ternary operator as an `if else`

PR #747, opened 2020-01-13
[WIP] Complete the new student signup flow (PR 2 of 2)

BryanEliDimas:
There are a few things TODO from the last merged PR. Since I needed to have that PR merged, I decided to split up the work for the issue into 2 PRs.

- [ ] Create feature and unit tests for the signup flow


BryanEliDimas:
Things that I'm aware of aren't working yet:

- [ ] Feature flag https://github.com/openstax/business-intel/issues/778
- [ ] All of the security logs in the right places, aptly named.
- [ ] From "My Account" page, adding a social login strategy. (It currently works but redirects to the old flow.
- [ ] The "My Account" page's code is super ugly — meaning, it needs to be cleaned up because I have a lot of inline CSS, I just superseded/overwrote the old CSS instead of coding a new page from scratch, some things need to be done inline per the UX specs (while right now they redirect you to a new page), and maybe more things like that on the My Account page.
- [ ] I am not entirely sure that the redirects (specifically, the `redirect_back_path` is working properly all of the time under all circumstances).

Will update this list as I think of more things. Heads up @Dantemss.

BryanEliDimas:
The couple commits I had in this branch, I already committed in my other branch, so this PR can be closed.

PR #749, opened 2020-01-24
Student accounts report

edwoodward:
Admin command to query and display the number of student accounts created since July 1.


app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
+require 'date'
+
+module Admin
+    class StudentAccountsController < BaseController
+      layout 'admin'
+  
+      def actions
```
BryanEliDimas: Since [some newer version of Rail, I think Rails 5], you don't need to define a function in the controller for GET requests. So unless you're defining an `@instance_variable` or something like that, I'd suggest removing this function to make the controller cleaner ™️ (creating clean controllers, and models, is a big problem in Rails).

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+        #July is a keyword, so using month7
+        month7 = Date.parse year.inspect + "-07-01"
+        return (month7...today).count
+      end
+
+      def query(days_num)
+        role_type = :student
+        user_ids = User.where(role: role_type)
+        return SecurityLog.where(event_type: :sign_up_successful, user_id: user_ids).where('created_at > ?', days_num.days.ago).count
```
BryanEliDimas: Actually, `ActiveRecord` can take a `Date` object and automagically insert it properly into a query, so you don't need to calculate the number of days ahead of time. You can just do `SecurityLog.where('created_at >= ?', Date.parse('2019-07-01')).count`. And therefore, you can get rid of the `get_number_of_days` function.

Another thing, I would use `created_at >= ?` July 1st instead of `created_at > ?` July 1st to include those accounts created that day.

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        today = Date.new
+        year = today.year
+        month = today.month
+        if month < 6
+            year = year - 1
+        end
+        #July is a keyword, so using month7
+        month7 = Date.parse year.inspect + "-07-01"
+        return (month7...today).count
+      end
```
BryanEliDimas: See my other comment below where I suggest removing this function.

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+module Admin
+    class StudentAccountsController < BaseController
+      layout 'admin'
+  
+      def actions
+      end
+  
+      def get_number_of_accounts
+        days_num = get_number_of_days
+        accounts_num = query(days_num)
```
BryanEliDimas: I would rename `accounts_num` to `result`. But also note that after removing `get_number_of_days`, these two lines will just be `result = query` which is okay I suppose (although there might be no need to do that and just call `query` where needed.

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        flash[:notice] = "The number of student acccounts created since July 1 is " + accounts_num.to_s
```
BryanEliDimas: Please use string interpolation like such: `"The number of student acccounts created since July 1st 2019 is #{result}"`

app/views/layouts/admin.html.erb
```diff
@@ -44,6 +44,7 @@
                 <li><%= link_to 'Actions', main_app.routes.url_helpers.actions_admin_users_path %></li>
                 <li><%= link_to 'Pre-Auth States', main_app.routes.url_helpers.admin_pre_auth_states_path %></li>
                 <li><%= link_to 'Reports', main_app.routes.url_helpers.admin_reports_path %></li>
+                <li><%= link_to 'Student Accounts Report', main_app.routes.url_helpers.actions_admin_student_accounts_path %></li>
```
BryanEliDimas: I think that creating a new page and everything for just this one report isn't necessary. To me it seems like the `'Reports'` page was perfect for that. Here's what it currently looks like: 

Screen Shot 2020-01-28 at 10 41 58 AM

This report, then, would be just one extra line :).

config/routes.rb
```diff
@@ -210,6 +210,13 @@
       end
     end
 
+    resource :student_accounts, only: [], controller: :student_accounts do
+      collection do
+        get :actions
```
BryanEliDimas: I would call this `reports` instead of `actions`. `action` in Rails has a specific/particular meaning, and it's not **really** an "action"—IMO more of a query or a report. I suppose unless you think of it as **executing** a query 🤷‍♂ .

app/controllers/admin/student_accounts_controller.rb
BryanEliDimas: Also, interesting way to do this :) (displaying a query's result as an `alert`. I would've displayed it inline like the Reports page does.

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,37 @@
+require 'date'
```
BryanEliDimas: I don't think that you need to do this. I believe that `'date'` is present/required by default in Ruby/Rails.

app/views/layouts/admin.html.erb
edwoodward: @BryanEliDimas I did not add this to the existing user report because the use case is different. The existing report is about user totals. I added a report for student accounts created since July 1. I thought mixing the two could be confusing.

app/controllers/admin/student_accounts_controller.rb
edwoodward: @BryanEliDimas I used the Salesforce update as an example to create my code.

app/controllers/admin/student_accounts_controller.rb
edwoodward: The page will not display without actions being defined. I originally was using show, but Dante had me change it to actions.

app/controllers/admin/student_accounts_controller.rb
edwoodward: Requested change has been made

app/controllers/admin/student_accounts_controller.rb
edwoodward: I still need to calculate which year to use in the query. I renamed the function and removed unneeded code.

app/controllers/admin/student_accounts_controller.rb
edwoodward: I renamed it to be more descriptive.

app/controllers/admin/student_accounts_controller.rb
edwoodward: Requested change has been made

config/routes.rb
edwoodward: I tried this and it did not work. I reverted the code.

app/controllers/admin/student_accounts_controller.rb
edwoodward: This change has been made.

app/views/layouts/admin.html.erb
jpslav: I think it is probably ok to add it to the user reports page, because it is reporting on "student users", just as how the other sections here are about "general users" and "faculty users".  

So a heading of "Students" with a line item of "New since July 1" seems reasonable.

app/controllers/admin/student_accounts_controller.rb
jpslav: @edwoodward I looked at this and thought "every time I try some custom code like this I find that Rails has already done it somewhere", so I looked around, but I didn't find it! So congrats on finding a hole! :-)

We'll want to test this code somewhere, and testing utility methods in a controller is kind of a heavy lift, so I think a better thing to do might be to monkey patch (extend) the `DateTime` class.  

E.g. you could make a file in `lib/date_time.rb` that has:

```ruby
class DateTime        # in Ruby you can reopen a class and add more stuff to it
  def self.last_july_1
    # pasting code here from your `get_year` and `query` methods
    today = DateTime.now
    year = today.year
    month = today.month
    if month < 7
      year = year - 1
    end
    # Should probably use `DateTime` here since rails dates are DateTime objects, and 
    # can also use string interpolation, i.e. `DateTime.parse("#{year}-07-01")`
    Date.parse(year.inspect + '-07-01')    
  end
end
```

Instead of `last_july_1` a name that fits other [`DateTime` methods](https://api.rubyonrails.org/classes/DateTime.html) might be `beginning_of_rice_fiscal_year` or `beginning_of_last_q3` or `beginning_of_last_july`.

Files in the `lib` directory are not automatically read by Rails like those in `app` are.  To make sure the code above is seen by Rails, you'd `require` it from some other file that Rails already requires.  What we typically use is `config/environment.rb` -- if you go here you'll see [a bunch of `require` statements](https://github.com/openstax/accounts/blob/master/config/environment.rb#L4-L22) for files in the `lib` directory.  

With this `DateTime.last_july_1` method in hand, you can write a spec in `spec/lib/date_time_spec.rb` (a file you'd make) to test that it gives the right datetime.  To test it, you'll want to make the method see fake values for `DateTime.now` in your code (so that you can test that "last July 1" is different if "now" is Sept 28 vs April 1.  In rspec, the way to do this is to mock the return value for `DateTime.now` with a line such as:

```ruby
allow(DateTime).to receive(:now).and_return(DateTime.parse("2019-09-28"))
```

You'll see examples of this all over the specs.  You can run this individual spec file with:

```
$> bundle exec rspec ./spec/lib/date_time_spec.rb
```

With the spec working, see below for a comment that talks about using your new method.

app/controllers/admin/student_accounts_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        if month < 7
+            year = year - 1
+        end
+        return year
+      end
+
+      def query(year)
+        role_type = :student
+        user_ids = User.where(role: role_type)
+        return SecurityLog.where(event_type: :sign_up_successful, user_id: user_ids).where('created_at >= ?', Date.parse(year.inspect + '-07-01')).count
```
jpslav: @BryanEliDimas can check me on this, but I _think_ that a created `User` record is a valid, succesfully-signed-up user (i.e. that we don't have users in incomplete states that never fully signed up).

If as discussed in the other comment we switch to putting this query in the user reports page, we should then able to follow the [pattern of queries there](https://github.com/openstax/accounts/blob/master/app/views/admin/reports/show.html.erb#L5-L17) and compute the count you want with:

```ruby
User.student.where(created_at: DateTime.last_july_1..DateTime.now).count
```

Breaking that down...

`.student` is a utility method provided by the [`enum :role` line](https://github.com/openstax/accounts/blob/master/app/models/user.rb#L37) in the `User` model, it is the same as saying `.where(role: :student)`.  See more about ActiveRecord enums [here](https://api.rubyonrails.org/v5.2.4.1/classes/ActiveRecord/Enum.html)

`DateTime.last_july_1..DateTime.now` returns a [Ruby range](https://ruby-doc.org/core-2.5.1/Range.html) with a start time of last July and an end time of right now.

`.where(created_at: some_range)` does a query where `created_at` falls in that range

config/routes.rb
jpslav: For this to work, would have had to change this to `get :reports` and change the controller method from `def actions` to `def reports` and change `actions.html.erb` to `reports.html.erb`.

app/controllers/admin/student_accounts_controller.rb
edwoodward: DateTime function created and spec added

app/views/layouts/admin.html.erb
edwoodward: Report moved to User report

app/controllers/admin/student_accounts_controller.rb
edwoodward: Done

spec/lib/date_time_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'date'
+require 'date_time'
+
+describe DateTime do
+    it 'get correct beginning of rice fiscal year' do
+        allow(DateTime).to receive(:now).and_return(Date.parse("2019-09-28"))
+        year1 = DateTime.beginning_of_rice_fiscal_year
```
jpslav: maybe not `year1` for var name b/c it isn't a year value.  could just say

`expect(DateTime.beginning_of_rice_fiscal_year).to eq DateTime.parse("2019-07-01")` 

on this line and similar expectations after each `allow` line

spec/lib/date_time_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'date'
+require 'date_time'
+
+describe DateTime do
+    it 'get correct beginning of rice fiscal year' do
```
jpslav: I think 2 spaces for indent is the Accounts way

spec/lib/date_time_spec.rb
edwoodward: Done

spec/lib/date_time_spec.rb
edwoodward: Done

PR #751, opened 2020-01-24
Add unverified contact infos to CS report with pin codes

mwharrison:
This adds all the contact infos to the CS report so they can get users their pins if they don't get the email to confirm their account.

### DIFF
--- a/app/views/admin/pre_auth_states/index.html.erb
+++ b/app/views/admin/pre_auth_states/index.html.erb
@@ -12,6 +12,9 @@
   <p class="desc">Signup states record transient information users provide as they are going through the sign up process.  These
     states are deleted once sign up is complete, so the data on this page is only useful for helping a user who is
     stuck in the middle of the sign up process.</p>
+  <p class="desc">There are lists compiled for users that go through the old accounts flow (should only be educators) and the
+    new account flow (should only be students). You can also use these to check that users are going to the proper flows.
+  These lists will be combined when the new flow is rolled out for all users (and this messaging should be removed).</p>
 
   <h4>Notes</h4>
 
@@ -22,8 +25,9 @@
   </ol>
 
   <h4 style="margin-top:20px">Records created within last <%= pluralize((params[:since] || 1).to_i, "day") %>, newest records first</h4>
-
-  <% @pre_auth_states.each do |pre_auth_state| %>
+  <hr>
+  <h5>Old Flow Users</h5>
+  <% if @pre_auth_states.each do |pre_auth_state| %>
     <div class="entry">
       <div class="basics">
         <%= pre_auth_state.contact_info_value || "No email yet" %> |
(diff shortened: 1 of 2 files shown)


app/views/admin/pre_auth_states/index.html.erb
```diff
@@ -41,6 +45,30 @@
         </div>
       <% end %>
     </div>
+  <% end.empty? %>
```
BryanEliDimas: Haven't reviewed the whole thing completely yet but this line does not look correct to me because `each` doesn't return anything like the way `map` does, so `empty?` will probably always evaluate to `true`. The fix _might be_ simply replacing `each` with `map`. Even better, though, would be to do `if`...`else`, more Ruby-like.

app/views/admin/pre_auth_states/index.html.erb
```diff
@@ -41,6 +45,30 @@
...
+    <i>There are no users in an unverified state</i>
+  <% end %>
+
+  <hr>
+  <h5>New Flow Users</h5>
```
BryanEliDimas: They're not really "users" or "contacts"—I'd say they're "emails." Fyi, I think I would've listed @unverified_users = User.where(state: 'unverified')` and then for each of those `user.email_address.first.confirmation_pin `, etc. because `@unverified_contacts = ContactInfo.`... on the other hand, will also list additional emails that users have added to their account after the fact (after signup) but have not gotten around to verifying yet.

app/views/admin/pre_auth_states/index.html.erb
mwharrison: I tested this.. it works

app/views/admin/pre_auth_states/index.html.erb
mwharrison: What if they contact CS about one of those 'other emails' they added? (although they should only be adding one during initial signup.


I did `@unverified_users = User.where(state: 'unverified')` first.. but it ended up with me having to do more loops to get all the unverified emails.

app/views/admin/pre_auth_states/index.html.erb
BryanEliDimas: Ok, I'm surprised (it ain't Ruby-like) but I believe you; might be worth writing a feature test though?

app/views/admin/pre_auth_states/index.html.erb
BryanEliDimas: > What if they contact CS about one of those 'other emails' they added? 

Well, that's a different story, that's what the security logs and User Search page are for. This page is to answer the question "how many people started the signup process but didn't finish?" That's exactly what a PreAuthState is.

mwharrison: I thought this page was to give CS access to the users' pin if they don't get the email.

app/views/admin/pre_auth_states/index.html.erb
```diff
@@ -12,6 +12,9 @@
   <p class="desc">Signup states record transient information users provide as they are going through the sign up process.  These
     states are deleted once sign up is complete, so the data on this page is only useful for helping a user who is
     stuck in the middle of the sign up process.</p>
+  <p class="desc">There are lists compiled for users that go through the old accounts flow (should only be educators) and the
+    new account flow (should only be students). You can also use these to check that users are going to the proper flows.
```
BryanEliDimas: > You can also use these to check that users are going to the proper flows.

How so? 🤔

PR #759, opened 2020-03-03
Add error/validation when password is the same as before

BryanEliDimas:
Part of https://github.com/openstax/business-intel/issues/851

### DIFF
--- a/spec/handlers/newflow/change_password_spec.rb
+++ b/spec/handlers/newflow/change_password_spec.rb
@@ -31,29 +31,33 @@ module Newflow
         FactoryBot.create(:identity, user: user, password: 'password')
       end
 
-      it 'fails if password is too short' do
-        params = {
+      let(:params) do
+        {
           change_password_form: {
-            password: 'pwd',
+            password: password,
           }
         }
-
-        result = described_class.call(caller: user, params: params)
-
-        expect(result.errors.any?).to be(true)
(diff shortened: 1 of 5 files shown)


BryanEliDimas:
Can I please get a code review? @steveburkett

spec/handlers/newflow/change_password_spec.rb
```diff
@@ -27,19 +27,33 @@ module Newflow
...
-      }
-    end
+      before do
+        FactoryBot.create(:identity, user: user, password: 'password')
+      end
 
       it 'fails if password is too short' do
-        # create the password to begin with
-        FactoryBot.create(:identity, user: user, password: 'password')
+        params = {
```
steveburkett: minor...this params construct (all thru the test) could be dryed up a bit thru a let(:params) and a let(:password) fields.

BryanEliDimas: Nice, I'd like to do that but I can't use `let(:password)` from within an example 🤔 .

BryanEliDimas: Done. I used nested `describe` blocks and `example do` blocks.

BryanEliDimas:
@steveburkett Dry-ed up the specs as you suggested, it worked out well! Specs passed. Merging PR.

PR #760, opened 2020-03-03
Hide student role option when feature flag is on

BryanEliDimas:
Part of https://github.com/openstax/business-intel/issues/851

### DIFF
--- a/app/views/signup/start.html.erb
+++ b/app/views/signup/start.html.erb
@@ -15,6 +15,11 @@ signup_roles = {
   other: t('.other')
 }
 
+# hide student role option when feature flag is on.
+if Settings::Db.store.newflow_feature_flag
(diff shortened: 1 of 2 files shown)


app/views/signup/start.html.erb
```diff
@@ -15,6 +15,11 @@ signup_roles = {
   other: t('.other')
 }
 
+# hide student role option when feature flag is on.
+if Settings::Db.store.newflow_feature_flag
```
steveburkett: this is a lot of conditional logic in a view template.  can this be pulled out into helpers or a decorator that can be unit tested?  is this IF statement tested somethere?

BryanEliDimas:
Added tests. Merging. Thanks for reviewing @steveburkett

PR #761, opened 2020-03-03
Add back_arrow to Forgot password form

BryanEliDimas:
Part of openstax/business-intel#851


BryanEliDimas:
Thanks for approving. Merging.

PR #762, opened 2020-03-03
Pre-populate email value in Change your (signup) email form

BryanEliDimas:
Part of openstax/business-intel#851

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -14,7 +14,10 @@ module Newflow
     before_action :maybe_skip_to_sign_up, only: [:login_form]
     before_action :known_signup_role_redirect, only: [:login_form]
     before_action :restart_if_missing_unverified_user,
-      only: [:verify_email, :verify_pin, :change_your_email, :confirmation_form, :confirmation_form_updated_email]
+      only: [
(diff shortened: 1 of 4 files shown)


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -70,6 +73,10 @@ def confirmation_form
       @email = unverified_user.email_addresses.first.value
     end
 
+    def change_your_email
+      @email = unverified_user.email_addresses.first.value
```
steveburkett: can anything in line 77 be nil at any time?  this is assuming that email_addresses is an array and always has something.

steveburkett: also, the name of this function seems a bit off from what the code is doing, which is setting email from an unverified user.  should the def be called:    set_email_from_unverified_user?

BryanEliDimas: Nope. An unverified user is a user that started but hasn't finished signing up. They must always have one email address. A user `has_many :email_addresses`, so it is indeed always an array.

`change_your_email` is not just a function to set the email. It's actually the name of a controller action which has its corresponding view called `change_your_email.html.erb`.

BryanEliDimas:
Thanks for approving the PR. Addressed your comments. Merged.

PR #764, opened 2020-03-04
Remove unverified users over 1 year old

steveburkett:
Cleanup unverified users over 1 year old.

https://app.zenhub.com/workspaces/openstax-bit-5bacfe3f4b5806bc2bea3764/issues/openstax/business-intel/865


spec/models/user_spec.rb
```diff
@@ -378,4 +378,19 @@
...
+      FactoryBot.create(:user, created_at: 10.years.ago, state: User::UNVERIFIED)
+      FactoryBot.create(:user, created_at: 1.month.ago, state: User::ACTIVATED)
+      FactoryBot.create(:user, created_at: 1.month.ago, state: User::UNVERIFIED)
+    end
+
+    it 'removes the unverified users older than a year' do
+      expect(User.count).to eq 4
+      described_class.cleanup_unverified_users
+      expect(User.count).to eq 3
+    end
```
BryanEliDimas: This doesn't really test that an UNVERIFIED, created more than a year ago, user is deleted. It only tests that A USER, some user, gets deleted. If an activated user got deleted instead, this test would still pass. Also if the wrong unverified user gets deleted, test will still pass 😟.

BryanEliDimas: In fact, I think the wrong unverified user is getting deleted.

After the expectations I wrote a `debugger`:
```
      expect(User.count).to eq 4
      described_class.cleanup_unverified_users
      expect(User.count).to eq 3
      debugger
```
And then I checked for the unverified user that survived:
```
(byebug) User.by_unverified.first.created_at
Thu, 04 Mar 2010 22:09:08 UTC +00:00
```
Which seems wrong to me. So I think the user scope is wrong, should be less than: 
```
where("created_at < ?", 1.year.ago)
```

steveburkett: Fixed. thanks!

PR #765, opened 2020-03-05
Track Google Analytics events in new Student flow

edwoodward:


app/views/layouts/_google_analytics.html.erb
```diff
@@ -7,5 +7,71 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
...
+      eventLabel: 'Referrer - ' + document.referrer
+    });
+    
+  } else if(window.location.href.indexOf("profile") > -1) {
+    ga('send', {
+      hitType: 'event',
+      eventCategory: 'Profile',
+      eventAction: 'Referrer',
+      eventLabel: 'Referrer - ' + document.referrer
+    });
```
BryanEliDimas: This `if(window.location.href.indexOf("profile") > -1)` will evaluate to true for:
1. the new flow profile `/i/profile`
2. the old flow profile `/profile`
3. **and** the old flow "signup profile" step `/signup/profile`

Wouldn't we want to differentiate between the three of them?

app/views/layouts/_google_analytics.html.erb
```diff
@@ -7,5 +7,71 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
...
+  ga('send', {
+      hitType: 'event',
+      eventCategory: 'SHOW/HIDE password',
+      eventAction: 'show/hide',
+      eventLabel: buttonText + ' - ' + location
+    });
+}
+
+document.addEventListener("DOMContentLoaded", function() {
+  if(window.location.href.indexOf("login") > -1) {
```
BryanEliDimas: If we only want to track these events for the new flow, you might want to check instead for `indexOf("i/login")` or `indexOf(<%= newflow_login_path %>)`

app/views/layouts/_google_analytics.html.erb
```diff
@@ -7,5 +7,79 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
...
+  ga('send', {
+    hitType: 'event',
+    eventCategory: category,
+    eventAction: action,
+    eventLabel: label
+  });
+});
+
+window.addEventListener('beforeunload', (event) => {
+  if((window.location.href.indexOf("signup") > -1) || (window.location.href.indexOf("profile") > -1)) {
```
BryanEliDimas: Same as my other comment...

`window.location.href.indexOf("profile") > -1` will evaluate to true for:
1. the new flow profile `/i/profile`
2. the old flow profile `/profile`
3. **and** the old flow "signup profile" step `/signup/profile`

Wouldn't we want to differentiate between the three of them?

app/views/newflow/login_signup/student_signup_form.html.erb
```diff
@@ -40,14 +40,16 @@
 
                         <div class="fb-share-button">
                             <%= link_to newflow_auth_path(:facebooknewflow),
-                                        class: 'facebook btn' do %>
+                                        class: 'facebook btn',
+                                        :data => {:ga_category => 'Account Creation', ga_action: 'Click', ga_label: '3B-Facebook'} do %>
```
BryanEliDimas: Don't we only want to track **successful** logins with facebook? Otherwise, this will only track how many people **tried** to log in with facebook. Same for google.

app/views/newflow/login_signup/student_signup_form.html.erb
```diff
@@ -155,7 +157,8 @@
                 <div class="content">
                     <%= f.submit I18n.t(:"login_signup_form.continue_button"),
                                         data: { disable_with: I18n.t(:"login_signup_form.continue_button") },
-                                        class: 'primary', id: 'signup_form_submit_button' %>
+                                        class: 'primary', id: 'signup_form_submit_button',
+                                        :data => {:ga_category => 'Account Creation', ga_action: 'Click', ga_label: '3A-Email'} %>
```
BryanEliDimas: Similar as the facebook and google buttons. This will only track **how many times** someone **tried** to sign up with email and password. I want to make sure, is that really what we want?

app/views/layouts/_google_analytics.html.erb
```diff
@@ -7,5 +7,79 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
...
+      eventAction: 'show/hide',
+      eventLabel: buttonText + ' - ' + location
+    });
+}
+
+document.addEventListener("DOMContentLoaded", function() {
+  if(window.location.href.indexOf("login") > -1) {
+    ga('send', {
+      hitType: 'event',
+      eventCategory: 'Login &amp; Account Creation',
```
BryanEliDimas: How come here is `&amp;` but  in https://github.com/openstax/accounts/pull/765/files?short_path=04c6e90#diff-27b203ccd75c16fc33e2a9a21845ea8fR75 there is `&`? Perhaps it'll make no difference in the GA console but it _could_ which would result in two different event categories.

app/views/newflow/login_signup/login_form.html.erb
```diff
@@ -71,7 +71,8 @@
                     <div class="content">
                         <%= f.submit I18n.t(:"login_signup_form.continue_button"),
                                     disable_with: I18n.t(:"login_signup_form.continue_button"),
-                                    class: 'primary'
+                                    class: 'primary', 
+                                    :data => {:ga_category => 'Login & Account Creation', ga_action: 'Continue', ga_label: 'Continue '}
```
BryanEliDimas: It would be really nice if we used the new Ruby syntax `key: 'value'` vs `:key => 'value'` everywhere.

edwoodward: I tried this, but it threw an error when using key:value on the first item in the hash. `syntax error, unexpected ':', expecting =>`

app/views/newflow/login_signup/student_signup_form.html.erb
edwoodward: How would I detect a failed login? It would not return to one of our pages would it? I could check if the user got to the profile page via facebook or google as the referrer.

app/views/newflow/login_signup/student_signup_form.html.erb
edwoodward: The requirements state to have it on the Continue button.

app/views/newflow/login_signup/student_signup_form.html.erb
edwoodward: Also the requirements state capturing clicking the Facebook and Google buttons.

app/views/layouts/_google_analytics.html.erb
edwoodward: This is capturing the new flow, so I updated the index check.

app/views/layouts/_google_analytics.html.erb
edwoodward: Myeditor did that and I neglected to go back and fix it. Thanks for catching it. I changed it to &.

app/views/layouts/_google_analytics.html.erb
edwoodward: Good catch. I made changes to the index checks.

app/views/layouts/_google_analytics.html.erb
edwoodward: This is capturing the new flow, so I updated the index check.

app/views/newflow/login_signup/login_form.html.erb
BryanEliDimas: Something's not right. Try this has got to work:
`data: { ga_category : 'Login & Account Creation', ga_action: 'Continue', ga_label: 'Continue ' }`

BryanEliDimas: I made the change, pushed a commit.

BryanEliDimas:
@edwoodward I made the Ruby syntax changes and pushed a commit. Then I rebased master because there were conflicts with profile_newflow.hml.erb and had to force-push that.

BryanEliDimas:
Waiting for specs to pass and then I'll merge.

app/views/newflow/login_signup/student_signup_form.html.erb
BryanEliDimas: Ok, that makes sense because we have another way of tracking actually successful signups (the Security Logs) 👍.

app/views/newflow/login_signup/student_signup_form.html.erb
BryanEliDimas: Since we have another way of tracking successful signups (the Security Logs), then just tracking clicks regardless of the outcome is another thing we want to track I suppose, it makes sense, I just wanted to be sure that was the case. All good.

app/views/layouts/_google_analytics.html.erb
BryanEliDimas: Great. Thanks.

PR #769, opened 2020-03-06
Admin reports for daniel

BryanEliDimas:
Issue https://github.com/openstax/business-intel/issues/608

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -244,4 +244,12 @@ module ApplicationHelper
     # by_email_or_username
     !current_user.is_anonymous? && EmailAddress.verified.where(user: current_user).first.try(:value)
   end
+
+  def average_time_for_users_to_become_activated
+    query = "SELECT AVG(EXTRACT('EPOCH' FROM AGE(created_at, activated_at))) FROM users WHERE activated_at IS NOT NULL"
+    diff_in_seconds = ActiveRecord::Base.connection.execute(query)[0]['avg']
+    return 'unknown' unless diff_in_seconds.present?
+    distance_of_time_in_words(diff_in_seconds)
+  end
+
 end
--- a/app/views/admin/reports/show.html.erb
+++ b/app/views/admin/reports/show.html.erb
@@ -21,6 +21,17 @@
 <h3>Students</h3>
 <p>
   Student accounts created since July 1: <%= User.student.where(created_at: DateTime.beginning_of_rice_fiscal_year..DateTime.now).count %> <br>
-  Student accounts created in the past week: <%= User.student.where(created_at: DateTime.now - 1.week..DateTime.now).count %>
-</p>
+  Student accounts created in the past week: <%= User.student.where(created_at: DateTime.now - 1.week..DateTime.now).count %> <br>
+  How many people start the account creation process versus how many finish it? <br>
+  <% since_year = 2018 %>
+  <% years = since_year..Time.now.year %>
+  <% years.each do |year| %>
+    <% time_range = Time.local(year)..Time.local(year + 1) %>
+    <% pas = PreAuthState.where(created_at: time_range).count %>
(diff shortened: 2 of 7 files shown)


BryanEliDimas:
@steveburkett Admittedly, this baby's ugly 🥴. But as long as the test is correct (as long as it works), as you can tell, for admin-only features we don't care as much about performance or code cleanness.

app/helpers/application_helper.rb
```diff
@@ -244,4 +244,12 @@ def current_users_resetting_password_email
     # by_email_or_username
     !current_user.is_anonymous? && EmailAddress.verified.where(user: current_user).first.try(:value)
   end
+
+  def average_time_for_users_to_become_activated
+    query = "SELECT AVG(EXTRACT('EPOCH' FROM AGE(created_at, activated_at))) FROM users WHERE activated_at IS NOT NULL"
```
steveburkett: suggestion-queries like this should live in the model (so that it can be reused and unit tested there)

app/views/admin/reports/show.html.erb
```diff
@@ -21,6 +21,19 @@
...
+  Student accounts created in the past week: <%= User.student.where(created_at: DateTime.now - 1.week..DateTime.now).count %> <br>
+  How many people start the account creation process versus how many finish it? <br>
+  <% since_year = 2018 %>
+  <% years = since_year..Time.now.year %>
+  <% years.each do |year| %>
+    <% time_range = Time.local(year)..Time.local(year + 1) %>
+    <% pas = PreAuthState.where(created_at: time_range).count %>
+    <% users = User.where(state: User::ACTIVATED, created_at: time_range).count %>
+
+    <%= "#{year}:  #{pas} started / #{users} finished = #{ ((users > 0 ? (pas / Float(users)) : 0) * 100).round(2) }%" %>
```
steveburkett: this sort of thing should live in a decorator (or at least a helper) and unit tested there

app/views/admin/reports/show.html.erb
```diff
@@ -21,6 +21,19 @@
...
-  Student accounts created in the past week: <%= User.student.where(created_at: DateTime.now - 1.week..DateTime.now).count %>
-</p>
+  Student accounts created in the past week: <%= User.student.where(created_at: DateTime.now - 1.week..DateTime.now).count %> <br>
+  How many people start the account creation process versus how many finish it? <br>
+  <% since_year = 2018 %>
+  <% years = since_year..Time.now.year %>
+  <% years.each do |year| %>
+    <% time_range = Time.local(year)..Time.local(year + 1) %>
+    <% pas = PreAuthState.where(created_at: time_range).count %>
+    <% users = User.where(state: User::ACTIVATED, created_at: time_range).count %>
```
steveburkett: this should live in a model, as a scope, where it can be unit tested and reused.  the whole years stuff here encapsulated in a helper

spec/features/admin/reports_spec.rb
```diff
@@ -33,7 +34,46 @@
...
+    end
+
+    visit admin_reports_path
+
+    expect(page).to have_content('2018: 2 started / 4 finished = 50.0%')
+    expect(page).to have_content('2019: 3 started / 6 finished = 50.0%')
+    expect(page).to have_content('2020: 4 started / 9 finished = 44.44%') # 8 finished plus the admin
+  end
+
+  context "What's the average time someone spends signing up (started the process but did not finish)?" do
```
steveburkett: view tests are so much slower than isolated unit tests (mentioned above).  way better to isolate unit test and have minimal view testing.  view testing is usually best as golden path.

BryanEliDimas:
These are excellent suggestions and I wish I had more time to make the changes now but I'll have to postpone that for later. Merging this PR now. Thanks @steveburkett! 

> @steveburkett Admittedly, this baby's ugly 🥴. But as long as the test is correct (as long as it works), as you can tell, for admin-only features we don't care as much about performance or code cleanness.

PR #773, opened 2020-03-12
add migration to remove newflow_feature_flag

steveburkett:
The newflow_feature_flag has been renamed (to studentflow_feature_flag).   The older flag (newflow) needed to be removed from the db as well due to use of rails_settings_cache gem.

### DIFF
--- a/db/migrate/20200311202045_update_settings_value_to_student_flow.rb
+++ b/db/migrate/20200311202045_update_settings_value_to_student_flow.rb
@@ -1,0 +1,8 @@
+class UpdateSettingsValueToStudentFlow < ActiveRecord::Migration[5.2]
+  def up
+    ActiveRecord::Base.connection.execute("delete from Settings where var = 'newflow_feature_flag'")
+  end
+
+  def down
+  end
+end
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -268,11 +268,7 @@
     t.text "return_to"
     t.jsonb "signed_data"
     t.boolean "is_partial_info_allowed", default: false, null: false
-    t.string "first_name", default: ""
-    t.string "last_name", default: ""
-    t.bigint "user_id"
     t.index ["contact_info_kind"], name: "index_pre_auth_states_on_contact_info_kind"
-    t.index ["user_id"], name: "index_pre_auth_states_on_user_id"
```
BryanEliDimas: Why were these removed? Maybe you need to rebase?

db/migrate/20200311202045_update_settings_value_to_student_flow.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateSettingsValueToStudentFlow < ActiveRecord::Migration[5.2]
+  def up
+    ActiveRecord::Base.connection.execute("delete from Settings where var = 'newflow_feature_flag'")
```
BryanEliDimas: Great.

db/schema.rb
BryanEliDimas: You were right to remove this.

BryanEliDimas:
Rebased, so I had to force-push.

PR #778, opened 2020-03-18
Logo now points to current environment

edwoodward:
Dependent on https://github.com/openstax/tutor-deployment/pull/553

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,8 +6,8 @@ module ApplicationHelper
     'http://openstax.force.com/support/articles/FAQ/Can-t-log-in-to-your-OpenStax-account'
   end
 
-  def accounts_url
-    main_app.root_path
+  def logo_url
+    Rails.application.secrets.openstax_url
   end
 
   def unless_errors(options={}, &block)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,8 +6,8 @@ module ApplicationHelper
     'http://openstax.force.com/support/articles/FAQ/Can-t-log-in-to-your-OpenStax-account'
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -6,8 +6,8 @@ def knowledge_base_url
     'http://openstax.force.com/support/articles/FAQ/Can-t-log-in-to-your-OpenStax-account'
   end
 
-  def openstax_url
-    'https://openstax.org'
+  def accounts_url
+    main_app.root_path
```
BryanEliDimas: Isn't this the same as doing `'/'`?

BryanEliDimas: Which would point to Accounts (`accounts-dev.openstax.org`, `accounts.openstax.org` in prod, etc.) whereas we want it to point to "osweb" (`cms-dev.openstax.org`, `openstax.org` in prod, etc.)

edwoodward: Possibly. I implemented what was used in the old flow based on a conversation with Brittany.

edwoodward: Updated to use deployment information for openstax URL

app/helpers/application_helper.rb
```diff
@@ -6,8 +6,8 @@ def knowledge_base_url
     'http://openstax.force.com/support/articles/FAQ/Can-t-log-in-to-your-OpenStax-account'
   end
 
-  def openstax_url
-    'https://openstax.org'
+  def logo_url
+    Rails.application.secrets.openstax_url
```
BryanEliDimas: Please add an entry to secrets.yml.example file?

edwoodward: Done

PR #779, opened 2020-03-18
Store referring app on login

BryanEliDimas:
Addresses issue https://github.com/openstax/business-intel/issues/704

### DIFF
--- a/spec/lib/os_web_string_spec.rb
+++ b/spec/lib/os_web_string_spec.rb
@@ -1,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe OsWebString, type: :lib do
+  let(:osweb_url) { 'https://openstax.org' }
+
+  before do
+    Rails.application.secrets.openstax_url = osweb_url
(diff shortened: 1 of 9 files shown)


spec/lib/os_web_string_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.describe OsWebString, type: :lib do
+  let(:osweb_url) { 'https://openstax.org' }
+
+  before do
+    Rails.application.secrets.openstax_url = osweb_url
+  end
+
+  example 'a passed-in string matches what we have stored in the secrets as the openstax_url' do
+    expect(described_class.new(osweb_url).came_from_osweb?).to eq(true)
```
steveburkett: even cooler here would be to declare a subject(so that it's obvious)..which is the super-string, and use the named subject in your expect.

BryanEliDimas:
We discovered that this PR wont fix the underlying problem. Closing it.

PR #782, opened 2020-03-23
Added check for gaShowHide before it is called

edwoodward:


app/assets/javascripts/newflow.js
```diff
@@ -15,10 +15,17 @@ $(document).ready(function () {
         var password_field = $('[name$="[password]"]')[0];
         if ($(password_field).attr("type") == "password") {
             $(password_field).attr("type", "text");
-            gaShowHide('Show');
+            if (typeof gaShowHide === "function")
```
steveburkett: the other idea (that's prob. better) is to have the function always defined, but does nothing in dev mode.  this is a bit more scalable than this (defensive) idea.  but it works.

PR #783, opened 2020-03-24
WIP- Changes to work with New deployment

scb6:


Gemfile
```diff
@@ -11,6 +11,11 @@ gem 'rails-i18n', '~> 5'
 gem 'pattern-library', git: 'https://github.com/openstax/pattern-library.git', branch: 'master'
 gem 'bootsnap', require: false
 
+# New Deployments
+gem 'aws-sdk-ssm'
+gem 'dotenv'
```
BryanEliDimas: I don't think you needed to add this because it's already there as 'dotenv-rails': https://github.com/openstax/accounts/pull/783/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R156

config/application.rb
```diff
@@ -46,21 +46,34 @@ class Application < Rails::Application
...
+        expires_in: 90.minutes,
+        compress: true,
+      }
+    end 
 
     def is_real_production?
       secrets.environment_name == "prodtutor"
     end
 
+    def is_assets_precompile?
```
BryanEliDimas: You defined this function twice

config/secrets.yml.example
```diff
@@ -92,9 +78,6 @@ development:
       old_name: <%= ENV['SSO_COOKIE_OLD_NAME'] || 'ox' %>
       name: <%= ENV['SSO_COOKIE_NAME'] || 'oxa' %>
       options:
-        <% if ENV.key?('SSO_COOKIE_DOMAIN') %>
-        domain: <%= ENV['SSO_COOKIE_DOMAIN'] %>
-        <% end %>
```
BryanEliDimas: I don't think these should've been deleted but I'm not sure.

config/secrets.yml.example
```diff
@@ -190,3 +173,101 @@ test:
 
   # The ip-api.com service gives us location data based on user IP address
   ip_api_key: the_ip_api_key
+
+production:
```
BryanEliDimas: Won't the deployment scripts overwrite everything under this key?

config/secrets.yml.example
```diff
@@ -190,3 +173,101 @@ test:
...
+  # Exception email settings
+  exception:
+    contact_name: OpenStax Accounts
+    sender: '"Sender" <sender@localhost>'
+    recipients: recipients@localhost
+
+  # Salesforce settings
+  salesforce:
+    login_domain: ssm(salesforce_login_domain)
+    username: ssm(salesforce_username)
```
BryanEliDimas: No idea what `ssm` is. Seems like a function but yaml has no functions afaik.

PR #788, opened 2020-03-26
Also push to SalesForce the phone number upon signup with social

BryanEliDimas:
When feature flag is OFF, the Educator button in "welcome page" should redirect to the old flow, and only when it is ON, should it redirect to the new Educator flow.


BryanEliDimas:
I think this is not necessary, actually, because Educators are *not* able to signup with a social provider. Closing it.

PR #790, opened 2020-03-26
Use Educator feat flag for signup path

BryanEliDimas:
Only redirect to the new flow Educator signup path if the feature flag for educators is ON.

### DIFF
--- a/app/views/newflow/login_signup/welcome.html.erb
+++ b/app/views/newflow/login_signup/welcome.html.erb
@@ -1,3 +1,7 @@
+<%
+conditional_educator_signup_path = Settings::Db.store.educator_feature_flag ? educator_signup_path : signup_path(bpff: 9)
+%>
+
 <div id="login-signup-form">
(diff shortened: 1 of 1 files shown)


app/views/newflow/login_signup/welcome.html.erb
```diff
@@ -1,3 +1,7 @@
+<%
+conditional_educator_signup_path = Settings::Db.store.educator_feature_flag ? educator_signup_path : signup_path(bpff: 9)
```
steveburkett: can we name this something better than bpff? :)

steveburkett: not a big deal here...but would be great to put conditional logic in a helper so that it can be easily tested.

BryanEliDimas: Sure! Feel free to create a PR or make the change in one of your PRs.

PR #793, opened 2020-03-27
Removed call to SecurityLog since user_session_management is handling it

edwoodward:
user_session_management is adding more data to the SecurityLog than login_signup_controller, so I decided to keep it and remove the duplicate.


BryanEliDimas:
Actually, I would like to include this PR in this release (code freeze on Fri 3/27). I'll merge it and create another accounts version/release. Thanks Ed.

Btw, both sec log entries contain the same amount of information... I just hadn't noticed that I'd ended up creating duplicate sec log entries 👍 .

PR #794, opened 2020-03-30
Fix error reported by Sentry: undefined method 'email_addresses'

BryanEliDimas:
Fixes error`NoMethodError: undefined method email_addresses' for nil:NilClass` 
in `/i/verify_email_by_pin` 

https://sentry.cnx.org/openstax/accounts/issues/64952/

Also solves a related error: https://sentry.cnx.org/openstax/accounts/issues/129080/

### DIFF
--- a/app/routines/newflow/activate_user.rb
+++ b/app/routines/newflow/activate_user.rb
@@ -1,3 +1,6 @@
+# Sets the passed-in user's `state` to `'activated'`
+# and pushes the user to salesforce as a new lead
+# If the user is already `activated`, then it does nothing.
 module Newflow
   class ActivateUser
     lev_routine
@@ -5,8 +8,9 @@ module Newflow
     protected ###############
 
     def exec(user)
-      push_lead_to_salesforce(user)
+      return if user.state.is_activated?
 
+      push_lead_to_salesforce(user)
       user.update(state: 'activated')
     end
 
--- a/app/routines/newflow/activate_user.rb
+++ b/app/routines/newflow/activate_user.rb
@@ -8,7 +8,7 @@ module Newflow
     protected ###############
 
     def exec(user)
-      return if user.state.is_activated?
+      return if user.is_activated?
 
       push_lead_to_salesforce(user)
       user.update(state: 'activated')
--- a/app/routines/newflow/activate_user.rb
+++ b/app/routines/newflow/activate_user.rb
@@ -8,7 +8,7 @@ module Newflow
     protected ###############
 
     def exec(user)
-      return if user.is_activated?
+      return if user.activated?
 
       push_lead_to_salesforce(user)
       user.update(state: 'activated')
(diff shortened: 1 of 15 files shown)


BryanEliDimas:
@steveburkett Please code review. Thanks in advance.

app/routines/newflow/activate_user.rb
```diff
@@ -1,12 +1,16 @@
...
+# If the user is already `activated`, then it does nothing.
 module Newflow
   class ActivateUser
     lev_routine
 
     protected ###############
 
     def exec(user)
-      push_lead_to_salesforce(user)
+      return if user.state.is_activated?
```
steveburkett: this feels a bit like a violation of the law of demeter.   http://www.virtuouscode.com/2011/07/05/demeter-its-not-just-a-good-idea-its-the-law/

consider perhaps a method on the user object: 
```
class User
   def activated?   (notice this is more idiomatic ruby than is_X(and a ?) 
```

steveburkett: When it does nothing...is there a situation where you might want to know that?  lots of problems become tough problems in rails app because things just silently get decided..not sure what the answer might be, just asking the question.  would you want to know why you called activate and it did nothing? (rails log? an event?)

steveburkett: last ?.  is this being tested somewhere? (that a message is being sent to the user object to activate and a lead pushed to salesforce if a user is in a particular state?

BryanEliDimas: 1. Fixed the apparent law of demeter violation. It was straight-up a bug. The method is defined on user, not on state.
2. **Excellent** question! Silent failures are the devil. "When it does nothing" — I don't know what I should do, if anything 🤔 . I don't think I need to know if the user had already been activated as it isn't a problem **at all**. In fact, the code that calls `ActivateUser` can and should be used to add/verify a new email address for an existing (already activated) user so in those cases it is expected that the user is already activated and calling `ActivateUser` again, even if it does nothing, is not a problem. 🤔 ... That's what I believe. Please, you are more than welcome to bring up 
3. Yes, it is being tested in `VerifyEmailByCode` with `expect_any_instance_of(ActivateUser).to receive(:call).and_call_original` and `ActivateUser` has its own unit tests :)

BryanEliDimas:
Ready for code review @steveburkett . Thanks.

app/routines/newflow/activate_user.rb
```diff
@@ -1,12 +1,16 @@
...
+# If the user is already `activated`, then it does nothing.
 module Newflow
   class ActivateUser
     lev_routine
 
     protected ###############
 
     def exec(user)
-      push_lead_to_salesforce(user)
+      return if user.is_activated?
```
steveburkett: is_X? (is and ?) isnt idiomatic ruby.   that's a java thing.   suggestion...change this to `user.activated?`

BryanEliDimas:
@steveburkett Renamed User#is_activated? to User#activated? :) Merging....

PR #795, opened 2020-03-31
Fix: not saving updated info from social provider

BryanEliDimas:
as I was fixing a bug, I noticed another bug: it turns out that the form where you Confirm your [social] info is not saving the first name + last name IF you change it from what we got from the social provider. This only affects new users signing up with a social provider who wanted to change their first name and/or last name to something different than what is stored in their social network provider. 

## Fixes issue https://github.com/openstax/business-intel/issues/985


BryanEliDimas:
Ready for code review @steveburkett

PR #796, opened 2020-03-31
Prevent users that still need a profile from being sent to OAuth apps

Dantemss:
Also adds a name to users that already escaped profile creation.

### DIFF
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -12,6 +12,7 @@ Doorkeeper.configure do
       newflow_authenticate_user!
     else
       authenticate_user!
+      redirect_to '/signup/profile' if current_user.is_needs_profile?
     end
 
     current_user
--- a/db/migrate/20200331192627_assign_names_to_users_that_bypassed_profile_creation.rb
+++ b/db/migrate/20200331192627_assign_names_to_users_that_bypassed_profile_creation.rb
@@ -1,9 +1,16 @@
 class AssignNamesToUsersThatBypassedProfileCreation < ActiveRecord::Migration[5.2]
   def up
     User.joins(:application_users).where(state: :needs_profile).distinct.find_each do |user|
-      user.first_name ||= 'Unknown'
-      user.last_name ||= user.email_addresses.verified.first&.value || 'Unknown'
-      user.save!
+      next unless user.username.blank?
+
+      email = user.email_addresses.verified.first
+      username = if email.nil?
+        SecureRandom.urlsafe_base64
+      else
+        email.value.split('@').first.gsub(/[^a-zA-Z\d]+/, '_')
+      end
+
+      user.update_attribute :username, username
     end
   end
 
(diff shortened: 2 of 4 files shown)


config/initializers/doorkeeper.rb
```diff
@@ -12,6 +12,7 @@
       newflow_authenticate_user!
     else
       authenticate_user!
+      redirect_to '/signup/profile' if current_user.is_needs_profile?
```
BryanEliDimas: I don't think this is necessary/the problem because there's already this: https://github.com/openstax/accounts/blob/master/app/controllers/application_controller.rb#L8

db/migrate/20200331192627_assign_names_to_users_that_bypassed_profile_creation.rb
```diff
@@ -0,0 +1,12 @@
+class AssignNamesToUsersThatBypassedProfileCreation < ActiveRecord::Migration[5.2]
+  def up
+    User.joins(:application_users).where(state: :needs_profile).distinct.find_each do |user|
+      user.first_name ||= 'Unknown'
+      user.last_name ||= user.email_addresses.verified.first&.value || 'Unknown'
```
BryanEliDimas: I don't think this is what Carly meant. I think users may have either 1) a username or 2) first and last name. So what Carly and I were suggesting is that you just generate a `username` for them based on their email address. So for me, for example, my email address is `bryaneli@rice.edu`—make my username `bryaneli`? Looks better than first name `Unknown`.

Dantemss: I'll let @carlycurran decide what to do here.

carlycurran: Yes I was thinking that we could leave first and last names blank and make the username filed the email address. Does that conflict with the new accounts flow, where we only have email address and no user name?

Dantemss: Accounts and the accounts-rails gem only allow letters, numbers and _ in the username so to put the email addresses in there we'd have to either replace the @ and . with _ or relax these rules.

carlycurran: ugh of course there's a catch. could we use what comes before the @ in the email address?

reedstrm: replace all non_allowed w/ underscore?

config/initializers/doorkeeper.rb
Dantemss: Oh I replied in slack instead of here... the Doorkeeper controller doesn't inherit from the ApplicationController (it inherits from ActionController::Base) so the before_action doesn't work for it.

db/migrate/20200331192627_assign_names_to_users_that_bypassed_profile_creation.rb
Dantemss: Done

config/initializers/doorkeeper.rb
BryanEliDimas: I was gonna ask if we can merge this, and then I started thinking that perhaps we also need to do this when the student (and educator?) feature flag is ON. Because we can expect that it is always ON from now and in the future. So this line should be moved outside and below the if/else.

PR #797, opened 2020-03-31
Fix: undefined method downcase for nil

BryanEliDimas:
Fixes [this error](https://sentry.cnx.org/openstax/accounts/issues/126859/) found by Sentry.

Solution: do what the old flow did — instead of `by_verified_email`, use `LookupUsers.by_email_or_username` which is smarter and handles the cases when the email from social provider is `nil` but also, it looks up users by username 🤷‍♂.

### DIFF
--- a/spec/handlers/newflow/oauth_callback_spec.rb
+++ b/spec/handlers/newflow/oauth_callback_spec.rb
@@ -73,5 +73,18 @@ module Newflow
         end
       end
     end
+
+    context 'when authentication found, but the response from social provider does not include email' do
+      let(:request) {
+        create_newflow_user(email)
+        info = { email: nil, name: Faker::Name.name }
+        MockOmniauthRequest.new 'facebook', Faker::Internet.uuid, info
+      }
+
+      it 'doesn\'t blow up' do
+        result = described_class.call(request: request)
+        expect(result.outputs.user).to eq User.last
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/handlers/newflow/oauth_callback_spec.rb
```diff
@@ -73,5 +73,18 @@ module Newflow
         end
       end
     end
+
+    context 'when authentication found, but the response from social provider does not include email' do
+      let(:request) {
+        create_newflow_user(email)
```
steveburkett: these seem like things that should be split into lets and before do/end blocks?  the info is a let, create_newflow_user is a before, the request is the subject (object being tested)?  separating like this creates clarity, and clarity improves readability and maintainability ... 

http://www.betterspecs.org/#subject

spec/handlers/newflow/oauth_callback_spec.rb
```diff
@@ -73,5 +73,18 @@ module Newflow
...
+        info = { email: nil, name: Faker::Name.name }
+        MockOmniauthRequest.new 'facebook', Faker::Internet.uuid, info
+      }
+
+      it 'doesn\'t blow up' do
```
steveburkett: how about it 'processes the omniauth request without error'?  doesnt blow up doesnt add much information to the it block when you look at the rspec test output.

BryanEliDimas: Thanks for the suggestion. Fixed.

spec/handlers/newflow/oauth_callback_spec.rb
BryanEliDimas: Will do. Great suggestions! 🙏

BryanEliDimas: Done.

PR #799, opened 2020-04-01
Educator complete form (step 4 of 4)

steveburkett:
Final step in the educator onboarding flow.  This PR is all the UI controls MINUS the dynamic behavior (that's a different card). 

https://app.zenhub.com/workspaces/openstax-bit-5bacfe3f4b5806bc2bea3764/issues/openstax/business-intel/920

educator complete


app/assets/stylesheets/newflow.scss
```diff
@@ -4,6 +4,8 @@ $font-path: '/accounts/fonts';
 @import "font-awesome";
 @import "newflow_colors";
 @import "intlTelInput";
+@import "radio_button";
+@import "radio_buttons";
```
BryanEliDimas, line 8: `radio_button` and `radio_buttons`? Looks like a typo, but is it? If not, please add a comment why.

app/helpers/newflow_form_helper.rb
```diff
@@ -42,29 +43,45 @@ def text_field(name:,
         desired_class_name = "#{desired_class_name} #{supplemental_class}"
       end
 
-      input = (
+      if numberonly
+        input = (
+        @f.number_field name,
```
BryanEliDimas, line 48: Instead of a `numberonly` option to the `text_field`, I would try to create a `number_field` function instead. But I'm guessing that you did consider it already, so just a suggestion.

config/locales/en.yml
```diff
@@ -1,4 +1,25 @@
 en:
+  educator_complete_form:
```
BryanEliDimas, line 2: Excellent decision to not nest these under `login_signup_form` 😄

config/routes.rb
```diff
@@ -16,6 +16,8 @@
     post 'i/signup/student', action: :student_signup, as: :newflow_signup_post
     get 'i/signup/educator', action: :educator_signup_form, as: :educator_signup
     post 'i/signup/educator', action: :educator_signup, as: :educator_signup_post
+    get 'i/signup/educator_complete', action: :educator_complete_form, as: :educator_complete
+    post 'i/signup/educator_complete', action: :educator_complete, as: :educator_complete_post
```
BryanEliDimas, line 20: This works, but it'd be better if they were nested under `/educator/` like so: `'i/signup/educator/complete'`

PR #801, opened 2020-04-02
Add Google client secrets to .env.example and add another test

BryanEliDimas:
as the title says. Just improvements I noticed while debugging something.

### DIFF
--- a/spec/handlers/newflow/student_signup_spec.rb
+++ b/spec/handlers/newflow/student_signup_spec.rb
@@ -7,7 +7,7 @@ module Newflow
         load('db/seeds.rb')
       end
 
-      let(:result) do
+      let(:subject) do
         described_class.call(params: params)
       end
 
@@ -22,7 +22,7 @@ module Newflow
             newsletter: true,
             contract_1_id: FinePrint::Contract.first.id,
             contract_2_id: FinePrint::Contract.second.id,
-            role: 'student'
+            role: :student
           }
         }
       end
@@ -32,15 +32,15 @@ module Newflow
       end
 
       it 'creates an (unverified) user with role = student' do
-        expect { result }.to change { User.where(state: 'unverified', role: 'student').count }
+        expect { subject }.to change { User.where(state: 'unverified', role: 'student').count }
       end
 
       it 'creates an identity' do
-        expect { result }.to change { Identity.count }
+        expect { subject }.to change { Identity.count }
       end
 
       it 'creates an authentication with provider = identity' do
-        expect { result }.to change { Authentication.where(provider: 'identity').count }
+        expect { subject }.to change { Authentication.where(provider: 'identity').count }
       end
 
       it 'agrees to terms of use and privacy policy when contracts_required' do
@@ -60,7 +60,7 @@ module Newflow
       end
 
       it 'creates an email address' do
-        expect { result }.to change { EmailAddress.count }
+        expect { subject }.to change { EmailAddress.count }
       end
 
       it 'sends a confirmation email' do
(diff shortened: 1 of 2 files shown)


spec/handlers/newflow/student_signup_spec.rb
```diff
@@ -69,14 +69,76 @@ module Newflow
...
         expect(User.last.receive_newsletter).to be(true)
       end
+
+      it 'outputs a user' do
+        expect(subject.outputs.user).to be_present
+      end
+    end
+
+    context 'success -- when passed-in a user_from_signed_params' do
+      # a user that already had an account, then tries to sign up using signed params (so via an LMS),
```
steveburkett, line 87: imo, context should be all u need here for comments...it should say what you're saying in the comment.  comments are rarely needed ever in code, code (esp ruby dsls like rspec, should read like prose.

spec/handlers/newflow/student_signup_spec.rb
```diff
@@ -7,7 +7,7 @@ module Newflow
         load('db/seeds.rb')
       end
 
-      let(:result) do
+      let(:subject) do
```
steveburkett, line 10: consider a named subject, to add context.  what is this?   :student_signup? a special one? what kind of one?

spec/handlers/newflow/student_signup_spec.rb
```diff
@@ -69,14 +69,76 @@ module Newflow
...
+        described_class.call(params: params, user_from_signed_params: user_from_signed_params)
+      end
+
+      it 'outputs a user' do
+        expect(subject.outputs.user).to be_present
+      end
+
+      it 'only creates one user' do
+        subject
```
steveburkett, line 139: subject is the object under test, it should do something.  e.g.,  signed_in_user.call (where signed_in_user is your subject).

spec/handlers/newflow/student_signup_spec.rb
BryanEliDimas: True. The comment is/was more descriptive than the context. Moved the comment into the context.

spec/handlers/newflow/student_signup_spec.rb
BryanEliDimas: Considered. Simply `subject` is perfectly fine IMO in this case.

spec/handlers/newflow/student_signup_spec.rb
BryanEliDimas: Very true. Fixed.

spec/handlers/newflow/student_signup_spec.rb
BryanEliDimas: It **does** something :)
```
        described_class.call(params: params, user_from_signed_params: user_from_signed_params)
```

steveburkett: @BryanEliDimas sorry, that wasnt my point.  my point was that the actual subject is an object (not an object and a verb).  see "RSpec's subject is a special variable that refers to the object being tested"https://stackoverflow.com/questions/38437162/whats-the-difference-between-rspecs-subject-and-let-when-should-they-be-used

BryanEliDimas: Isn't the `subject` the `described_class`, then?

steveburkett: @BryanEliDimas yeah, if it's that simple. sure.   the overall point is to use a named subject when it adds "documentation" and clarity to the test.  but of course, not to overengineer it, def didnt mean that :)

PR #786, opened 2020-03-25
fix the phone number to use the newform helper

steveburkett:
Fix problems found by developer examination. 

image

CodePen for the intlTelInput control https://codepen.io/jackocnr/pen/yehvE

PR #802, opened 2020-04-02
Submit role in educator signup form

BryanEliDimas:
Bug fix: the educator signup form is not submitting. It's missing the role. This PR fixes it.


app/views/newflow/login_signup/educator_signup_form.html.erb
```diff
@@ -137,6 +137,7 @@
             </label>
             <%= f.hidden_field :contract_1_id, value: contracts.first.id %>
             <%= f.hidden_field :contract_2_id, value: contracts.second.id %>
+            <%= f.hidden_field :role, value: :instructor %>
```
steveburkett, line 140: FYI, this was fixed in https://github.com/openstax/accounts/pull/786 (which is currently blocked).  Sounds good to get it in now of course. Thanks!

PR #803, opened 2020-04-03
Fix: undefined method is_activated? for nil

BryanEliDimas:
Fixes error reported by Sentry https://sentry.cnx.org/openstax/accounts/issues/129080/activity/

The problem was that `save_unverified_user` saves any user regardless of their actual `state` while `unverified_user` does `where(state: 'unverified` therefore returns only actual unverified users. In production, an error gets raised every time a user wants to login or signup but didn't finish the signup process in the old flow – which means that their their `state` is `needs_profile`. 

This PR fixes that by converting `needs_profile` users to `unverified` users. Also added an extra controller test to login_signup_controller to test `'when no email address is returned'` in oauth_callback.

## Issue https://github.com/openstax/business-intel/issues/991

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -176,7 +176,7 @@ module Newflow
 
             if !user.activated?
               # not activated means signup
-              unverified_user = EnsureUnverifiedUser.call(user).outputs.user
+              unverified_user = ensure_unverified_user(user)
 
               save_unverified_user(unverified_user)
               @first_name = user.first_name
@@ -356,6 +356,10 @@ module Newflow
         # TODO: when we create the Educator flow, redirect to there.
         redirect_to newflow_signup_student_path(request.query_parameters)
       end
+    end
+
+    def ensure_unverified_user(user)
+      EnsureUnverifiedUser.call(user).outputs.user
     end
 
     def cache_client_app
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -176,7 +176,13 @@ module Newflow
 
             if !user.activated?
               # not activated means signup
-              # NOTE: not necessarily "unverified" user, just not activated 😣. # TODO: refactor
+              # Users could have started signing up in the old flow in which case their state is 'needs_profile'
+              # ... in the new flow, that translates to 'unverified'. So update the user state:
+              if user.state == 'needs_profile'
+                user.update_attributes(state: 'unverified', is_newflow: true)
+                security_log(:user_updated, user: user, state_was: 'needs_profile', state_changed_to: 'unverified')
+              end
+
               save_unverified_user(user)
               @first_name = user.first_name
               @last_name = user.last_name
@@ -369,7 +375,7 @@ module Newflow
       id = session[:unverified_user_id]&.to_i
       return unless id.present?
 
-      @unverified_user ||= User.find_by(id: id)
+      @unverified_user ||= User.find_by(id: id, state: 'unverified')
     end
 
     def exit_newflow_signup_if_logged_in
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -176,6 +176,7 @@ module Newflow
 
             if !user.activated?
               # not activated means signup
+              # NOTE: not necessarily "unverified" user, just not activated 😣. # TODO: refactor
               save_unverified_user(user)
               @first_name = user.first_name
               @last_name = user.last_name
@@ -368,7 +369,7 @@ module Newflow
       id = session[:unverified_user_id]&.to_i
       return unless id.present?
 
-      @unverified_user ||= User.find_by(id: id, state: 'unverified')
+      @unverified_user ||= User.find_by(id: id)
     end
 
     def exit_newflow_signup_if_logged_in
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -176,14 +176,9 @@ module Newflow
 
             if !user.activated?
               # not activated means signup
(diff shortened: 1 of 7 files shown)


spec/rails_helper.rb
```diff
@@ -48,10 +48,10 @@
   Capybara::Selenium::Driver.new app, browser: :chrome, options: options
 end
 
-Capybara.javascript_driver = :selenium_chrome_headless
-
 if EnvUtilities.load_boolean(name: 'HEADLESS', default: false)
   # Run the feature specs in a full browser (note, this takes over your computer's focus)
+  Capybara.javascript_driver = :selenium_chrome_headless
```
BryanEliDimas, line 53: I had an off by one error 😬

app/controllers/newflow/login_signup_controller.rb
```diff
@@ -176,7 +176,9 @@ def oauth_callback
 
             if !user.activated?
               # not activated means signup
-              save_unverified_user(user)
+              unverified_user = EnsureUnverifiedUser.call(user).outputs.user
```
steveburkett, line 179: law of demeter violation.  spanning multiple types with periods. https://medium.com/@gioch/design-patterns-law-of-demeter-with-rails-49a44a9689fe

Violations like this dont need to be "the law".  just common sense.  when one call spans multiple types, it's creating a coupling that is brittle.  it makes changing other code down the road way more difficult.

app/routines/newflow/ensure_unverified_user.rb
```diff
@@ -0,0 +1,25 @@
+# Users who started signing up in the old flow but never filled out their profile information
+# are in the state 'needs_profile'. In the new flow, that translates to 'unverified'.
+# So we update their state to match the new flow.
+module Newflow
+  class EnsureUnverifiedUser
+    lev_routine
+
+    def exec(user)
+      if user.state == 'needs_profile'
```
steveburkett, line 9: I would def recommend putting strings like this in one place, as constants.  scattering them around like this is error prone and makes the code difficult to change.

spec/controllers/newflow/login_signup_controller_spec.rb
```diff
@@ -484,23 +484,47 @@ module Newflow
...
+            { email: nil, name: Faker::Name.name }
+          end
+
+          let(:params) do
+            { provider: 'facebook', uid: Faker::Internet.uuid }
+          end
+
+          before do
+            allow_any_instance_of(OauthCallback).to receive(:oauth_response)  do
+              # TODO: refactor this?
```
steveburkett, line 519: prob. best to refactor now instead of punting if at all possible.

spec/routines/newflow/ensure_unverified_user_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it 'outputs the user' do
+      expect(described_class.call(subject).outputs.user).to be_a(User)
+    end
+
+    it 'creates a security log' do
+      expect {
+        described_class.call(subject)
+      }.to change {
```
steveburkett, line 21: consider using begin/end for multi-line blocks (lines in an editor) 

> Prefer {...} over do...end for single-line blocks. Avoid using {...} for multi-line blocks. Always use do...end for “control flow” and “method definitions” (e.g. in Rakefiles and certain DSLs). Avoid do...end when chaining.

from https://shopify.github.io/ruby-style-guide/ and most other ruby style guides.

spec/routines/newflow/ensure_unverified_user_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        SecurityLog.where(event_type: :user_updated).count
+      }
+    end
+
+    it 'doesnt change the state if user is activated' do
+      subject = FactoryBot.create(:user, state: 'activated')
```
steveburkett, line 27: best to create a subject in a subject clause.  rspec allows and recommends this. http://www.betterspecs.org/#subject

spec/support/newflow_feature_helpers.rb
```diff
@@ -118,6 +118,7 @@ def submit_signup_form
   find('#login-signup-form').click # to hide the password tooltip
   wait_for_animations
   check('signup_terms_accepted')
+  wait_for_animations
```
steveburkett, line 121: does this guarantee to work?  does it slow the tests down?  stuff like this is the stuff that experienced rails devs shudder over (because as a project grows and gains thousands of tests, and when they get to taking 1 hour or so..they're often tasked to go in and remove javascript waits).  Not sure if there's a better answer or if this causes much of a delay, just bringing it up for consideration.

spec/routines/newflow/ensure_unverified_user_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      described_class.call(subject)
+      subject.reload
+      expect(subject.state).to eq('activated')
```
steveburkett, line 30: state = activated should be referred to as a constant.

spec/controllers/newflow/login_signup_controller_spec.rb
BryanEliDimas: I had a question mark in it because I didn't know if or how I could refactor it. After looking into it, I really don't know how to refactor it 😬 . Removed the comment 🙃 .

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Yes, I'm aware of the law of Demeter and that I'm breaking that rule there when I do that. The caveat is that you should know when to break the rule(s). In this case, and with all usage of the Lev gem in Accounts, we trust that Lev is pretty stable, relatively stable and so its interface will not change.

I did go ahead and encapsulated that piece of code in a function so that the controller action is more easily testable though. 👍 . Still not perfect but still an improvement, I believe.

BryanEliDimas:
Addressed all of your Code Review suggestions. Nice job and thanks for the through code review! 

Please re-review?

PR #804, opened 2020-04-03
Educator | Step 3 of 4 | SheerID iframe

BryanEliDimas:
Implements Educators' Step 3 of 4 (a SheerID iframe). 
Issue: https://github.com/openstax/business-intel/issues/868

Refactored the "old flow" code into a **legacy** module. Had to change the translations as well because of that—ie. `I18n.t(:"users.signup.start")` -> `I18n.t(:"legacy.users.signup.start")`


BryanEliDimas:
Just trying to merge commits and rebase against latest master but otherwise, this PR is ready for code review.

app/controllers/faculty_access_controller.rb
```diff
@@ -36,6 +37,12 @@ def pending
 
   protected
 
+  # def redirect_to_sheerid_if_newflow
+  #   if request.get? && Settings::Db.store.educator_feature_flag && !params[:bpff].present?
+  #     redirect_to(educator_sheerid_form)
+  #   end
+  # end
+
```
BryanEliDimas, line 45: Totally forgot about these comments in this controller. I'll delete them when I address any other suggestions you may have.

app/controllers/legacy/authentications_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+          security_log :authentication_deleted,
+                      authentication_id: authentication.id,
+                      authentication_provider: authentication.provider,
+                      authentication_uid: authentication.uid
+          render status: :ok,
+                plain: (I18n.t :"controllers.authentications.authentication_removed",
+                              authentication: params[:provider].titleize)
+        end,
+        failure: lambda do
+          render status: 422, plain: @handler_result.errors.map(&:message).to_sentence
```
steveburkett, line 20: consider moving this demeteried train of dots into one method:    @handler_result.human_friendly(), so that it can be used anywhere and unit tested inside the handler_result code.

app/controllers/legacy/sessions_controller.rb
```diff
@@ -0,0 +1,305 @@
...
+    # Handle OAuth callback (actual login)
+    # May add authentication method (OAuth provider) to account
+    def create
+      handle_with(
+        SessionsCreate,
+        user_state: self,
+        pre_auth_state: pre_auth_state,
+        login_providers: get_login_state[:providers],
+        success: -> do
+          authentication = @handler_result.outputs[:authentication]
```
steveburkett, line 90: consider using a polymorphic object (SessionsCreateUser) to handle this instead of this giant case statement

app/controllers/legacy/sessions_controller.rb
```diff
@@ -0,0 +1,305 @@
...
+        failure: -> do
+          errors = @handler_result.errors
+          lost_user = errors.any? do |error|
+            [:unknown_callback_state, :invalid_omniauth_data].include? error.code
+          end
+
+          if lost_user
+            # Something weird happened to the user's session or omniauth data and we lost their info
+            Rails.logger.warn do
+              oauth = request.env['omniauth.auth']
```
steveburkett, line 162: this is far far too large a method.  difficult to understand, test, extend.

app/views/newflow/password_management/forgot_password_form.html.erb
```diff
@@ -0,0 +1,73 @@
...
+    <div class="content back-arrow-container">
+        <%= render partial: 'newflow/tab_group', locals: { active_class: 'login--active'} %>
+
+        <%= newflow_login_signup_card(
+                    classes: "login-page",
+                    header: I18n.t(:"login_signup_form.reset_my_password")) do %>
+            <% lev_form_for :forgot_password_form, url: send_reset_password_email_path do |f| %>
+                <%
+                    form_helper = NewflowFormHelper::Newflow.new(
+                        f: f,
```
steveburkett, line 11: would really love to see variables fully spelled out and not "f".   yeah..we know what "f" means, but in general, always good to never use "f", or "x", or "z" sorts of things.

app/views/newflow/signup/_change_signup_email_form_template.html.erb
```diff
@@ -36,9 +40,13 @@
...
 
                 <div class="content">
-                    <%= f.submit I18n.t(:'login_signup_form.send_my_pin'),
-                                        disable_with: I18n.t(:'login_signup_form.continue_button'),
-                                        class: 'primary' %>
+                    <%=
+                        f.submit(
+                            I18n.t(:"login_signup_form.send_my_pin"),
+                            disable_with: I18n.t(:"login_signup_form.continue_button"),
+                            class: 'primary'
```
steveburkett, line 47: would like to see consistent use of " vs '.   i like the idea of ' for non-interpolated strings, " for interpolated ones.  this allows a quick glance to know which is which.

app/views/newflow/student_signup/student_email_verification_form.html.erb
```diff
@@ -0,0 +1,40 @@
+<%
+    page_header = I18n.t(
```
steveburkett, line 2: consider:  building up a data structure in the controller that has all this stuff and exposing a single PORO to this form for all its data needs (instead of building up this giant set of data inside the view.   it's usually best to keep views thin (any rails-y layer)

app/controllers/faculty_access_controller.rb
```diff
@@ -1,6 +1,7 @@
 class FacultyAccessController < ApplicationController
 
   prepend_before_action :disallow_signup, only: :apply
+  # before_action :redirect_to_sheerid_if_newflow, only: [:apply]
```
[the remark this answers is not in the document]
BryanEliDimas: Fixed and created controller specs.

app/controllers/legacy/authentications_controller.rb
BryanEliDimas: This is legacy code that is expected to be deleted soon. So honestly I don't think it's worth refactoring legacy code ATM. I admit that I did reuse that legacy code though, in SocialAuthController#remove_auth_strategy and I was tempted to change that line to `@handler_result.errors.first&.message` but it still suffers from the same problem anyway, so I'm just going to leave it untouched for now—again, because it's legacy code that we intend to get rid of replace soon.

app/controllers/legacy/sessions_controller.rb
BryanEliDimas: Very good suggestion, actually. I really like that idea, although, there _is_ already a `SessionsCreate` handler that does most of the logic related to this stuff.

app/controllers/legacy/sessions_controller.rb
BryanEliDimas: Totally agree! 😅  It caused me a significant amount of headaches in the beginning when trying to understand Accounts. But then again, this is now legacy code that is already getting replaced entirely by new code, so not worth refactoring this legacy code IMO.

app/views/newflow/password_management/forgot_password_form.html.erb
BryanEliDimas: I couldn't agree more! I did change some of them already, where `f` and `fh` are used instead of `form` and `form_helper` but I missed this one and I admit that there are still a few more left to spell out, it's just too many to fix at once. I could do it though, if you insist.

app/views/newflow/signup/_change_signup_email_form_template.html.erb
BryanEliDimas: You're referring to the translations' strings? I intentionally made the exception to that rule for ALL translations. At least now, though, that rule is consistent across ALL translations whereas before very very few of them used single quotes for translations. My personal preference when it comes to translations, so it could be changed. It would just be a TON of changes.

app/views/newflow/student_signup/student_email_verification_form.html.erb
BryanEliDimas: I love that idea! That would be a great next step for the newflow views 👍 . I do personally adopt the rule "make it work, then make it better" though. This makes it work _and_ it's already a huge improvement over how it was in the old flow, IMO. So, my hope is that already improved code like this will reveal patterns that can **later** then be implemented even more uniformly at a higher level.

PR #807, opened 2020-04-08
Added font files and set font-path

edwoodward:


BryanEliDimas:
@edwoodward Code reviewed. Restarted Travis because there are intermittent tests.

PR #809, opened 2020-04-08
Prevent 403 error when email blank on forgot password form

BryanEliDimas:
## Part of https://github.com/openstax/business-intel/issues/951

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -271,9 +271,10 @@ module Newflow
         },
         failure: lambda {
           user = @handler_result.outputs.user
+          email = @handler_result.outputs.email
           code = @handler_result.errors.first.code
-          security_log :reset_password_failed, user: user, reason: code
-          redirect_to newflow_login_path
+          security_log :reset_password_failed, user: user, email: email, reason: code
+          render :forgot_password_form
         }
       )
     end
--- a/app/handlers/newflow/send_reset_password_email.rb
+++ b/app/handlers/newflow/send_reset_password_email.rb
@@ -20,7 +20,7 @@ module Newflow
     def handle
       outputs.email = forgot_password_form_params.email
 
-      fatal_error(code: :blank,
+      fatal_error(code: :email_is_blank,
         offending_inputs: :email,
         message: I18n.t(:"login_signup_form.email_is_blank")
       ) unless outputs.email.present? || logged_in_user
--- a/app/handlers/newflow/send_reset_password_email.rb
+++ b/app/handlers/newflow/send_reset_password_email.rb
@@ -14,11 +14,17 @@ module Newflow
     protected #################
 
     def authorized?
-      forgot_password_form_params.email.present? || logged_in_user
+      true
     end
 
     def handle
       outputs.email = forgot_password_form_params.email
+
+      fatal_error(code: :blank,
+        offending_inputs: :email,
+        message: I18n.t(:"login_signup_form.email_is_blank")
+      ) unless outputs.email.present? || logged_in_user
+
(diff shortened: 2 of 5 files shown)


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -271,9 +271,10 @@ def send_reset_password_email
         },
         failure: lambda {
           user = @handler_result.outputs.user
+          email = @handler_result.outputs.email
```
steveburkett, line 274: optional -- this is a law of demeter violation...not a biggie (just 2 types), if it can be @handler_result.email (even as a delegate)?

app/handlers/newflow/send_reset_password_email.rb
```diff
@@ -14,11 +14,17 @@ class SendResetPasswordEmail
     protected #################
 
     def authorized?
-      forgot_password_form_params.email.present? || logged_in_user
+      true
     end
 
     def handle
       outputs.email = forgot_password_form_params.email
+
```
steveburkett, line 22: does handle have a spec to make sure it gets a fatal_error for these conditions?

BryanEliDimas: Now it does :) Pushed a commit. Thanks for the suggestion!

spec/handlers/newflow/send_reset_password_email_spec.rb
```diff
@@ -54,5 +54,23 @@ module Newflow
...
+    context 'when no email present and no logged in user' do
+      let(:params) do
+        {
+          forgot_password_form: {
+            email: ''
+          }
+        }
+      end
+
+      let(:subject) do
```
steveburkett, line 67: please use the rspec explicit subject http://www.betterspecs.org/#subject

spec/handlers/newflow/send_reset_password_email_spec.rb
```diff
@@ -54,5 +54,23 @@ module Newflow
         expect(result).to have_routine_error(:cannot_find_user)
       end
     end
+
+    context 'when no email present and no logged in user' do
+      let(:params) do
+        {
+          forgot_password_form: {
+            email: ''
```
steveburkett, line 62: is no email present empty string AND nil?  email: nil?  having '' here is unclear to this.

BryanEliDimas: **I believe** Rails `params` never post as `nil` for empty fields. Instead, they get submitted or interpreted as empty strings. Am I wrong? 🤔

steveburkett: i ask because it's prob. important to understand the possible test cases here:  no email field, nil email field, empty string email field, for thoroughness of the test.  just asking if you've done that here.

BryanEliDimas: I am not going to test all three cases (no email field, nil email field, empty string email field), I only whatever would get submitted by a form with nothing in the email field (in other words, when an end-user leaves the email field empty and submits the form that way... which I believe becomes an empty string in Rails in the `params` object.

BryanEliDimas: Either way, both empty string and nil evaluate to false by the `.present?` method which is what's important here. 

```
...
        unless outputs.email.present? || logged_in_user
```

spec/handlers/newflow/send_reset_password_email_spec.rb
BryanEliDimas: Fixed.

steveburkett:
yes, that was indeed the point (instead of a check like email == '').
that's good and the review comment prompted a check for this.  this is the
purpose of a review. thanks.

PR #810, opened 2020-04-10
Store return_to url in newflow

BryanEliDimas:
Might have forgotten to store the `return_to` url in the new flow like the old flow does. The action_interceptor gem and Tutor uses it I believe.


BryanEliDimas:
On second thought, this is really not a problem in the new flow, actually. The new flow just works differently. So I am not going to create an issue for it, just going to close this PR. 
cc: @BruceP99166

PR #811, opened 2020-04-10
Refactor

BryanEliDimas:
The goal is to **not** have the new flow JS affect the old flow JS or the other way around.


BryanEliDimas:
Ping: @steveburkett ; test email notifications

PR #813, opened 2020-04-13
Fix: X icon only reloads the page when arriving from an OAuth app

BryanEliDimas:
Fix https://github.com/openstax/business-intel/issues/968

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -1,6 +1,7 @@
 module Newflow
   # Contains every action for login and signup
   class LoginSignupController < ApplicationController
+    include LoginSignupHelper
     layout 'newflow_layout'
 
     skip_before_action :authenticate_user!
@@ -326,16 +327,16 @@ module Newflow
     end
 
     def exit_accounts
-      referrer_params = Addressable::URI.parse(request.referrer).query_values.to_h
+      referrer_params = extract_params(request.referrer)
 
-      if (return_to = referrer_params['r'])
-        if Host.trusted?(return_to)
-          redirect_to(return_to)
+      if (r = referrer_params[:r])
+        if Host.trusted?(r)
+          redirect_to(r)
         else
           raise Lev::SecurityTransgression
         end
-      elsif (return_to = referrer_params['redirect_uri'])
-        redirect_to(return_to)
+      elsif (redirect_uri = referrer_params[:redirect_uri])
+        redirect_to(redirect_uri)
       else
         redirect_back # defined in action_interceptor gem
       end
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -10,7 +10,7 @@ module Newflow
 
     before_action :newflow_authenticate_user!, only: [:profile_newflow]
     before_action :save_new_params_in_session, only: [:login_form]
-    before_action :store_authorization_url_as_fallback, only: [:login_form, :login, :student_signup_form, :student_signup]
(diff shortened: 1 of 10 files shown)


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -326,6 +325,22 @@ def change_password
       end
     end
 
+    def exit_accounts
+      referrer_params = Addressable::URI.parse(request.referrer).query_values.to_h
+
+      if (return_to = referrer_params['r'])
```
steveburkett, line 331: I would consider refactoring this to improve.    (1) not a great idea to assign variables inside of if statements.  (2) this method has high cyclometric complexity.  i encourage you to install and run flog on it https://blog.feabhas.com/2018/07/code-quality-cyclomatic-complexity/.   (3) does this method have unit tests covering all the if branches (one of the dangers of high cyclometric complexity, and bugs.

spec/features/newflow/student_login_flow_spec.rb
```diff
@@ -139,6 +139,41 @@
         expect(page).to have_text('This is another banner.')
         screenshot!
       end
+
+      context 'when user clicks X icon' do
+        context "when user arrived from oauth app" do
+          let(:app) do
+            app = FactoryBot.create(:doorkeeper_application, skip_terms: true,
```
steveburkett, line 146: would be nice to see this broken up a bit into lets and befores.

BryanEliDimas: Fixed.

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Re: 1. I agree that it's uncommon and should be avoided if necessary but I insist that doing so in this case is alright — especially after the refactoring I just did/pushed. Requesting re-review. :) 
2. Just glanced at the rather long article but I get the point, I know what you mean. 
3. I just improved all of my code immensely, IMO. I'm hoping that'll suffice. Requesting re-review. Thanks!

app/controllers/newflow/login_signup_controller.rb
```diff
@@ -326,16 +327,16 @@ def change_password
     end
 
     def exit_accounts
-      referrer_params = Addressable::URI.parse(request.referrer).query_values.to_h
+      referrer_params = extract_params(request.referrer)
 
-      if (return_to = referrer_params['r'])
-        if Host.trusted?(return_to)
-          redirect_to(return_to)
+      if (r = referrer_params[:r])
```
steveburkett, line 332: I would suggest a better name than r.  what is r?  it's good to write code thinking about how would a new dev look at this and instantly know what r is and what's going on

spec/controllers/newflow/login_signup_controller_spec.rb
```diff
@@ -756,5 +756,63 @@ module Newflow
...
+        let(:target_url) { Faker::Internet.url }
+
+        context 'when Referer includes `r` param' do
+          subject do
+            "#{host}?r=#{target_url}"
+          end
+
+          context 'when `r` is trusted' do
+            before do
+              allow(Host).to receive(:trusted?).with(target_url).once.and_return(true)
```
steveburkett, line 776: so much better! :)

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: 👍  I'm thinking of documenting this param in the https://github.com/openstax/accounts#special-parameters section of the README and/or adding a comment to the controller action `exit_accounts`.

PR #819, opened 2020-04-20
Document `r` and `redirect_uri` param usage

BryanEliDimas:
As it was suggested in code review for PR https://github.com/openstax/accounts/pull/813#discussion_r411689897

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -31,7 +31,7 @@ module Newflow
         success: lambda {
           clear_newflow_state
           sign_in!(@handler_result.outputs.user)
-          redirect_back # back to `r`eturn parameter. See `before_action :save_redirect`.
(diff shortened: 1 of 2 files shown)


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -329,7 +329,7 @@ def change_password
     def exit_accounts
       referrer_params = extract_params(request.referrer)
 
-      if (r = referrer_params[:r])
+      if (r = referrer_params[:r]) # `r`edirect parameter. See `before_action :save_redirect`.
```
steveburkett, line 332: why cant this just be `if (redirect_param = referrer_params[:r])` and not need the comment?

BryanEliDimas: Good question; sure, I can do that.

BryanEliDimas: Done.

PR #822, opened 2020-04-21
Fix exit icon when user goes to the signup page, then back to the login page, and then tries to exit

BryanEliDimas:
When Dev Verify-ing https://github.com/openstax/accounts/pull/813, I found that only checking the referrer for the `redirect_uri` would not work. This PR fixes that.

Before:
- From the login page, user goes away (ie. to the signup page), goes back to the login page, clicks on the `X` exit icon... user is redirected back to the `oauth/authorize` path (although from the user perspective it looks like the page just gets refreshed).

User had to click the `X` exit icon one more time in order to be taken back to the OAuth app (ie. Tutor). This is because `redirect_uri` gets cleared, so `redirect_back` takes user back to the `stored_url` which at that point was the `oauth/authorize` path. **Then**, the second time user click on X icon, `redirect_uri` is present, so user is successfully sent back to the app.

After:
- Always look for the `redirect_uri` in the `stored_url` because if the user arrived from an OAuth app from the beginning, then it was stored on arrival.

Created a feature spec for this scenario :).

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -333,7 +333,7 @@ module Newflow
         else
           raise Lev::SecurityTransgression
         end
-      elsif (redirect_uri = extract_params(stored_url)[:redirect_uri])
+      elsif !signed_in? && (redirect_uri = extract_params(stored_url)[:redirect_uri])
         redirect_to(redirect_uri)
       else
         redirect_back # defined in action_interceptor gem
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -327,15 +327,13 @@ module Newflow
     end
 
     def exit_accounts
-      referrer_params = extract_params(request.referrer)
-
-      if (redirect_param = referrer_params[:r])
+      if (redirect_param = extract_params(request.referrer)[:r])
         if Host.trusted?(redirect_param)
           redirect_to(redirect_param)
         else
           raise Lev::SecurityTransgression
         end
-      elsif (redirect_uri = referrer_params[:redirect_uri])
+      elsif (redirect_uri = extract_params(stored_url)[:redirect_uri])
         redirect_to(redirect_uri)
       else
         redirect_back # defined in action_interceptor gem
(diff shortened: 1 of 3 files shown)


BryanEliDimas:
Travis is correctly failing so moving this PR back to WIP.

BryanEliDimas:
@steveburkett Please code review?

app/controllers/newflow/login_signup_controller.rb
```diff
@@ -327,15 +327,13 @@ def change_password
...
-
-      if (redirect_param = referrer_params[:r])
+      if (redirect_param = extract_params(request.referrer)[:r])
         if Host.trusted?(redirect_param)
           redirect_to(redirect_param)
         else
           raise Lev::SecurityTransgression
         end
-      elsif (redirect_uri = referrer_params[:redirect_uri])
+      elsif !signed_in? && (redirect_uri = extract_params(stored_url)[:redirect_uri])
```
steveburkett, line 336: consider use of safe assignment inside a condition https://github.com/rubocop-hq/ruby-style-guide#safe-assignment-in-condition

spec/controllers/newflow/login_signup_controller_spec.rb
```diff
@@ -813,6 +802,21 @@ module Newflow
...
         end
       end
+
+      context 'when the stored url includes `redirect_uri` param' do
+        before do
+          allow_any_instance_of(described_class).to receive(:stored_url).and_return(subject)
+        end
+
+        subject do
+          "#{host}?redirect_uri=#{target_url}"
```
steveburkett, line 812: this looks more like a let to me? (the subject is the actual object under test, the controller in this case)

let(:redirect_uri) { ... }

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Yeap, I'm doing that. It says "[...] unless the assignment is wrapped in parentheses"

spec/controllers/newflow/login_signup_controller_spec.rb
BryanEliDimas: Oh, true. I am still not always sure which is the subject under test.

app/controllers/newflow/login_signup_controller.rb
steveburkett: yeah, ok.  i'm of the opinion it's best to avoid all assignments in IFs, because i know for a fact that they are a source of bugs down the road.  esp deeper nested IFs and scoping.  just my 2 cents....

PR #823, opened 2020-04-22
Changed font path, moved fonts into accounts directory

edwoodward:


BryanEliDimas:
Argh, just thought of something: this will work when Accounts is being accessed at openstax.org/accounts URL, but will it work when being accessed like accounts.openstax.org? @edwoodward

PR #824, opened 2020-04-22
Remove ses and update concourse

scb6:


scb6:
https://travis-ci.org/github/openstax/accounts/builds/678275723 is passing. Forcing merge.

PR #827, opened 2020-04-28
Service object to fetch book data from the CMS

BryanEliDimas:
Creates a service object to fetch book data from the CMS for use in the new educator flow — step 4 of 4.

This PR is the first part. The second part will be to cache the response and sync with the CMS every 1 day.


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -67,6 +67,12 @@ def educator_signup
       )
     end
 
+    def educator_complete_form
+      book_data = FetchBookData.new
```
steveburkett, line 71: suggestion:  create a "factory" (private) method that creates this thru memoization.

```
private
def book_data
  @book_data ||= FetchBookData.new
end

```
 then use it as book_data.subjects, book_data.titles

app/views/newflow/login_signup/educator_complete_form.html.erb
```diff
@@ -135,7 +135,7 @@
                 <%= I18n.t(:"educator_complete_form.select_option") %>
               </div>
               <%= fh.select name: :books_used,
-                            options: Settings::Subjects.map {|(short_name, info)| [  info["title"], short_name] },
+                            options: @book_titles,
```
steveburkett, line 138: be sure to ask @RoyEJohnson if his mult-select control is going to be redesigned to group titles by subject. that would have implications here.

lib/fetch_book_data.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def titles
+    @titles ||= fetch_titles
+  end
+
+  def fetch_subjects
+    results = cms_fetch(SUBJECTS_URL)
+    return [] if results.blank?
+
+    results.collect { |subject| subject.fetch('name', nil) }
```
steveburkett, line 22: "prefer map over collect" https://github.com/rubocop-hq/ruby-style-guide

lib/fetch_book_data.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    begin
+      Net::HTTP.start(uri.host, uri.port, use_ssl: true, read_timeout: TIMEOUT) do |http|
+        response = Net::HTTP.get_response(uri)
+        body = JSON.parse(response.body)
+        return body
+      end
+    rescue Net::ReadTimeout => ee
+      Raven.capture_message("Fetching book data from the CMS timed out")
+      return nil
```
steveburkett, line 46: suggestion.  this does work (of course, only in that you're capturing the exception.  i would still consider sending the ee.message and stack trace to Raven tho for ease of debugging).   A better pattern to follow, over returning nil is to return an error object that has contextual information.  This is a much better "code reuse" pattern to follow and more OO.  might be over-kill just for this specific situation, but mentioned it because it's a good practice to follow with API calls like this. 

more info... "nil values are the worst" by Avdi Grimm https://www.rubytapas.com/2017/01/31/two-screencasts-two-ways-eradicate-ruby-nil-values/

spec/lib/fetch_book_data_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'rails_helper'
+require 'vcr_helper'
+
+describe FetchBookData, type: :lib, vcr: VCR_OPTS do
+  describe '#subjects' do
+    subject { described_class.new.subjects }
```
steveburkett, line 6: suggestion:  consider using named subjects 

http://www.betterspecs.org/#subject

It's not mandatory, but why not take advantage of rspec (or ruby, or any dsl) to add even more self-documentation to code one writes, esp if it's free, which it is here.

lib/fetch_book_data.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# Service object for pulling in book data from the CMS
+class FetchBookData
+  CMS_API_URL = Rails.application.secrets.cms_api_url
+  SUBJECTS_URL = "#{CMS_API_URL}snippets/subjects/?format=json"
+  TITLES_URL = "#{CMS_API_URL}v2/pages/?type=books.Book&format=json&limit=250&fields=title,book_subjects,book_state"
+  TIMEOUT = 1
+
+  def subjects
+    @subjects ||= fetch_subjects
```
steveburkett, line 11: suggestion.   memoize the actual data from the call, but allow subjects and titles to be re-fetched, allow a flag that's defaulted to override this behavior to fetch live (this is just a consideration for multiple use of this object).  

what about caching the data (redis perhaps) ?  does each refresh of step4 need to fetch from the CMS for thousands of students a day?  how often do subjects and titles really change?   

have you you done performance testing on this call?  are you certain that thousands of educators at once signing up will be handled by this with quick response (hence the question about consideration of caching).  

something to think about.

spec/lib/fetch_book_data_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+        # cannot really test read timeout b/c of how webmock inserts itself
+        # https://github.com/bblimke/webmock/issues/286#issuecomment-19457387
+        allow(Net::HTTP).to receive(:start).and_raise(Net::ReadTimeout)
+      end
+
+      it 'returns empty array' do
+        expect(subject).to match([])
+      end
+
+      it 'calls Raven' do
```
steveburkett, line 29: this could be a tiny bit better if it said it 'calls Raven with a timed out message'.  that is even clearer.

spec/lib/fetch_book_data_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  describe '#subjects' do
+    subject { described_class.new.subjects }
+
+    context 'success' do
+      let(:book_subjects) do
+        ["Math", "Science", "Humanities", "Social Sciences", "Business", "Essentials", "College Success", "High School"]
+      end
+
+      example do
+        expect(subject).to match(book_subjects)
```
steveburkett, line 14: when the CMS changes, and adds a subject. this test would start failing.  is that intentional?

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Ah, yes. Thank you for the suggestion!

app/views/newflow/login_signup/educator_complete_form.html.erb
BryanEliDimas: Future feature :). Gonna make that improvement in the future. For now, a simple array works. 

In the future because we only have you in the team till today 😢 .

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Great idea. Fixed.

lib/fetch_book_data.rb
BryanEliDimas: Right; I am going to cache the API response but for now just creating the FetchBookData service that calls the API. 

> memoize the actual data from the call, but allow subjects and titles to be re-fetched, allow a flag that's defaulted to override this behavior to fetch live (this is just a consideration for multiple use of this object).

Not entirely sure what you mean by this, or what the advantage is of "memoize the actual data from the call, but allow subjects and titles to be re-fetched" 🤔 .

spec/lib/fetch_book_data_spec.rb
BryanEliDimas: Agree; fixed.

spec/lib/fetch_book_data_spec.rb
BryanEliDimas: Would it? I think it won't because we use the VCR gem, so the responses are cached (until you update them when you choose to by re-recording the request/response).

steveburkett: i mean this when VCR is cleared.

lib/fetch_book_data.rb
steveburkett: just making the suggestion to build in a little bit of flexibility into this class (optionally) to allow one to use it to fetch "live" or to fetch cached (perhaps by using a defaulted parameter).  this is not that big of a deal, esp since you're caching it anyway.

lib/fetch_book_data.rb
BryanEliDimas: > sending the ee.message and stack trace to Raven

Totally agree, that's a great idea when the exception is something other than a timeout (gonna send the `extra` option to Raven) but do we really need it when the exception is just a timeout?

BryanEliDimas: Oh, we already send to Raven the whole exception when the exception is something other than a timeout. Ok, still gonna send it when it's a timeout 😄 .

spec/lib/fetch_book_data_spec.rb
BryanEliDimas: In that case then, yes. I think so. Or what could/should I do instead?

lib/fetch_book_data.rb
BryanEliDimas: I see. Yeah, for now I'm going to handle the caching outside of this object.

BryanEliDimas:
@steveburkett  Pushed a commit and addressed all your comments. Please re-review?

PR #828, opened 2020-04-29
Oauth admins can edit redirect uri for oauth applications

edwoodward:


app/controllers/oauth/applications_controller.rb
```diff
@@ -36,51 +38,61 @@ def create
...
-        end
-      else
-        respond_to do |format|
-          format.html { render :edit }
-          format.json do
-            render json: { errors: @application.errors.full_messages },
-                   status: :unprocessable_entity
+      
+      Doorkeeper::Application.transaction do
+        if handle_member_ids(params["member_ids"]) && @application.update_attributes(app_params)
```
jpslav, line 75: I think you should be using `app_params[:member_ids]` (or similar) because `params` hasn't been permitted... when you call `.permit` on `params` down in `def app_params`, it is [returning a new object, not modifying the `params` object](https://apidock.com/rails/ActionController/Parameters/permit).

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
-        :email_from_address, :confidential,
-        :can_access_private_user_data, :can_find_or_create_accounts, :can_message_users,
-        :can_skip_oauth_screen,
-      )
+      if @user.is_administrator?
+        params.require(:doorkeeper_application).permit(
+          :name, :redirect_uri, :scopes, :email_subject_prefix, :lead_application_source,
+          :email_from_address, :confidential,
+          :can_access_private_user_data, :can_find_or_create_accounts, :can_message_users,
+          :can_skip_oauth_screen, :member_ids
```
jpslav, line 129: IIRC, `:member_ids` was outside of the nested `:doorkeeper_application` parameters?  If so I don't think using it here is doing what is desired since `params.require(:doorkeeper_application)` returns just the parameters nested under `:doorkeeper_application`.

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+          :redirect_uri
+        )
+      end
+    end
+
+    def handle_member_ids(member_ids)
+      if valid_member_ids(member_ids)
+        if !member_ids.nil?
+          str_array = member_ids.split(" ")
+          mi_array = str_array.map(&:to_i)
```
jpslav, line 142: I know above I recommended a different way to approach this, but if this were to stay the same I'd recommend some different naming conventions:

1. Make clear that the input to the method is a string and not an array of member IDs: `if valid_member_ids(member_ids)` --> `if valid_member_ids(member_ids_string)`
2. Avoid naming things "_array", just make them plural which implies an array: `mi_array` -> `member_ids` 

And as shown above, chaining short commands is a typical Ruby thing (Ruby values conciseness); you can eliminate the `str_array` here with `member_ids_string.split.map(&:to_i)`

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+          #add the owner back to the member_ids array
+          mi_array.unshift(@application.owner.owner_ids[0])
```
jpslav, line 144: What is going on in these two lines?

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+          @application.owner.member_ids = mi_array if @application.owner.member_ids != mi_array
```
jpslav, line 145: Why check that `mi_array` is changing the status quo?  At least here it doesn't really buy us an efficiency (may be less efficient due to the array comparison)

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+        end
+        return true
+      else
+        @application.errors.add(:owner, 'Member Ids must be a space separated list of integers')
+        return false
+      end
+    end
+
+    def valid_member_ids(member_ids)
+      if member_ids.nil? || member_ids.empty?
```
jpslav, line 155: FYI, in Rails, the ActiveSupport gem (which is always included) gives us the `blank?` method on basic classes: `nil.blank?` and `[].blank?` are both `true` so this line could be `if member_ids.blank?`

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+        return true
+      end
+      !!member_ids.match(NUMBERS_ONLY_REGEX)
```
jpslav, line 158: A more concise Ruby way to write this method is:

```ruby
def valid_member_ids(member_ids)
  member_ids.blank? || !!member_ids.match(NUMBERS_ONLY_REGEX)
end
```

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+    end
+
+    def set_member_ids
+      #remove the owner from the array since it cannot be removed by users
+      ids = @application.owner.member_ids
+      owner_id = @application.owner.owner_ids[0]
+      ids.delete(owner_id)
```
jpslav, line 165: Are you intending to modify the state of `@application.owner.member_ids` here?  `ids` is a reference to that array so I think that's what is happening.  E.g. see this test on the Ruby console:

```ruby
irb(main):002:0> hash = {a: [1,2]}
=> {:a=>[1, 2]}
irb(main):003:0> a_ref = hash[:a]
=> [1, 2]
irb(main):004:0> a_ref.delete(2)
=> 2
irb(main):005:0> hash
=> {:a=>[1]}
```

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+    end
+
+    def valid_member_ids(member_ids)
+      if member_ids.nil? || member_ids.empty?
+        return true
+      end
+      !!member_ids.match(NUMBERS_ONLY_REGEX)
+    end
+
+    def set_member_ids
```
jpslav, line 161: When I read this method name I read it as saying "I'm setting some instance variable called `@member_ids`".  But the method as written isn't setting anything, it is returning an ID. (while also deleting an entry from an application's `owner.member_ids` array which I don't quite follow?).  So two options:

1. Use a more descriptive name like `get_current_member_ids`, or
2. Since the only place this method is used you are actually setting the value of `@member_ids`, i.e. the calls are currently `@member_ids = set_member_ids`, you could actually set the `@member_ids` variable in this method.  Instance variables are available across methods and in the view, so this is legit.  Probably would still be a good idea to rename the method to something more descriptive.

app/controllers/oauth/applications_controller.rb
jpslav: Also what is going on with the owner ID and member ID -- that is not clear to me.  Whatever it is, can you exercise that code in a spec?

app/views/admin/users/_form.html.erb
```diff
@@ -244,6 +244,8 @@
   </div>
   <% end %>
 
+  <% if !@user.is_administrator? %>
```
jpslav, line 247: Why is this change in the PR and what is the reason for it?  If not an admin show the save button??  Seems odd.

lib/authenticate_methods.rb
```diff
@@ -63,7 +63,7 @@ def authenticate_user!
   end
 
   def authenticate_admin!
-    return if current_user.is_administrator?
+    return if current_user.is_administrator? || !current_user.oauth_applications.empty?
```
jpslav, line 66: This is a pretty massive bug.  Makes any user with oauth applications an administrator across all of Accounts.

I know you added this because of our conversation where I said "check that they are an admin or have oauth applications" but we need to do that check in a _separate_ method that is only used for the specific controller we are modifying.  `authenticate_admin!` is used all over the place to authenticate that someone is an admin.

I think something like the following would do it.

```ruby
class ApplicationsController < Doorkeeper::ApplicationsController
  before_action :authenticate_admin_or_oauth_application_owner!

  ...

  protected

  def authenticate_admin_or_oauth_application_owner!
    return if current_user.oauth_applications.any?
    authenticate_admin! # calling this existing method lets us take advantage of
                        # how it sends users to the login page and back here if not logged in
  end
end
```

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -160,13 +162,15 @@ module Oauth
     it "should not let a user get the list of his applications" do
```
jpslav, line 162: Since the HTTP status is now 200 should the "not" be in this spec title?

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -160,13 +162,15 @@ module Oauth
...
       controller.sign_in! user
       get(:index)
-      expect(response).to have_http_status :forbidden
+      expect(response).to have_http_status :success
+      expect(response.body).to eq('')
```
jpslav, line 166: Why doesn't the trusted application show up (since `trusted_application_admin.owner.add_member(user)` is run)

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -160,13 +162,15 @@ module Oauth
...
     end
 
     it "should not let a user get his own application" do
       controller.sign_in! user
       get(:show, params: { id: untrusted_application_user.id })
-      expect(response).to have_http_status :forbidden
+      expect(response).to have_http_status :success
+      expect(response.body).to eq('')
```
jpslav, line 173: ditto the 200 and "not"

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -6,6 +6,7 @@ module Oauth
     let!(:admin) { FactoryBot.create :user, :terms_agreed, :admin }
     let!(:user)  { FactoryBot.create :user, :terms_agreed }
     let!(:user2) { FactoryBot.create :user }
+    let!(:user3) { FactoryBot.create :user, :terms_agreed }
```
jpslav, line 9: I see that you are following the user variable naming pattern, but here it might be nice to give this user a meaningful name (for whatever its purpose is in this spec)

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -182,7 +186,7 @@ module Oauth
     end
 
     it "should not let a user create an application" do
-      controller.sign_in! user
+      controller.sign_in! user3
```
jpslav, line 189: only admins can create right? so can you reuse `user` here? Kind of nice that `user` has an oauth application and this would show that this user does not have create permissions.

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -197,7 +201,8 @@ module Oauth
     it "should not let a user edit his own application" do
       controller.sign_in! user
       get(:edit, params: { id: untrusted_application_user.id })
-      expect(response).to have_http_status :forbidden
+      expect(response).to have_http_status :success
```
jpslav, line 204: again 200 success responses do not equate to disallowed; this is probably something that should trigger your spidey senses -- like something is funky here.  Possible that when you fix the `authenticate_admin!` bug these will become forbidden again.

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -227,6 +232,73 @@ module Oauth
...
+    it "should let an admin create an application with Oauth Admins" do
+      controller.sign_in! admin
+      post(:create,
+        params: {
+            doorkeeper_application: {
+              name: 'Some app',
+              redirect_uri: 'https://www.example.com',
+              can_message_users: true,
+            },
+            member_ids: user2.id
```
jpslav, line 244: Is there a spec somewhere that exercises the "number space number space" member IDs input?  If not, we need one.

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -227,6 +232,73 @@ module Oauth
...
+        params: {
+          id: trusted_application_admin.id,
+          doorkeeper_application: {
+            redirect_uri: 'https://www.example.org'
+          }
+        }
+      )
+      expect(response).to redirect_to(oauth_application_path(trusted_application_admin.id))
+      expect(assigns(:application).redirect_uri).to eq('https://www.example.org')
+    end
```
jpslav, line 277: Can you add a spec showing that oauth admins cannot do certain updates? (all the things that weren't permitted in the controller)

app/controllers/oauth/applications_controller.rb
```diff
@@ -36,51 +38,61 @@ def create
...
+                                            application_params: app_params
+          flash[:notice] = I18n.t(
+            :notice, scope: %i[doorkeeper flash applications update]
+          )
+          respond_to do |format|
+            format.html { redirect_to oauth_application_url(@application) }
+            format.json { render json: @application }
+          end
+        else
+          @member_ids = set_member_ids
```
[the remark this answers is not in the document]
edwoodward: if I don't reset it, the Ids do not show up when there is an error. This was Dante's suggestion. The member_ids field is outside of the form values so I have to handle it manually.

app/controllers/oauth/applications_controller.rb
edwoodward: The owner cannot be removed, so I am removing the owner before editing and putting the owner back in the array before updating. The owner is always the first item in the member_ids, so I'm putting it back where it was.

app/controllers/oauth/applications_controller.rb
edwoodward: I tried using .blank and .present and did not get the correct results which is why I used this line. It is removed now.

app/controllers/oauth/applications_controller.rb
edwoodward: I knew my code did not have the correct "Rubyness". I'm learning to think that way.

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: The user does not get a list of applications, just the heading which is why I didn't change the name.

edwoodward: On further review, I no longer think this is a valid test. The user in the test is part of the member_ids so it will not fail as it previously did.

app/views/admin/users/_form.html.erb
edwoodward: I think I was originally thinking the oauth admin permission would be on the user and this was the start of that, but I found out that would not work and forgot I added the line. It is removed.

lib/authenticate_methods.rb
edwoodward: I searched for admin_authentication which was the alias for authenticate_admin and it is only used in the controller I was editing. I neglected to search for authenticate_admin. I assumed since it was a doorkeeper method it was only a part of the Oauth Applications. I missed that is was used in BaseController. Thanks for catching this.

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: This was caused by the improper implementation of admin_authentication. It is now returning a 403 as expected if the user is neither an admin nor an oauth admin.

app/controllers/oauth/applications_controller.rb
```diff
@@ -3,6 +3,8 @@ class ApplicationsController < Doorkeeper::ApplicationsController
     before_action :set_user
     before_action :admin_authentication!
 
+    NUMBERS_ONLY_REGEX = '^(?=.*\d)[\s\d]+$'.freeze
```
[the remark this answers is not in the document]
edwoodward: Done

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
-        :can_skip_oauth_screen,
-      )
+      if @user.is_administrator?
+        params.require(:doorkeeper_application).permit(
+          :name, :redirect_uri, :scopes, :email_subject_prefix, :lead_application_source,
+          :email_from_address, :confidential,
+          :can_access_private_user_data, :can_find_or_create_accounts, :can_message_users,
+          :can_skip_oauth_screen, :member_ids
+        )
+      elsif @application.owner.member_ids.include?(@user.id)
```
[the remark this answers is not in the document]
edwoodward: Done

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,49 @@ def set_user
...
+        params.require(:doorkeeper_application).permit(
+          :redirect_uri
+        )
+      end
+    end
+
+    def handle_member_ids(member_ids)
```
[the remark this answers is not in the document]
edwoodward: Done

app/controllers/oauth/applications_controller.rb
edwoodward: This is fixed

app/controllers/oauth/applications_controller.rb
edwoodward: Removed member_ids

app/controllers/oauth/applications_controller.rb
edwoodward: Got it

app/controllers/oauth/applications_controller.rb
edwoodward: Based on other coding in other frameworks, it is more efficient to do nothing with the database then to update it needlessly. Maybe the database does not get updated if there is not a change and I am not aware or Rails is super efficient with the database. The code has been changed.

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: renamed

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: Updated

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: I removed this test since it is no longer the application works

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: Added test

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: Test added

app/controllers/oauth/applications_controller.rb
edwoodward: I changed the code to make a duplicate of the member_id array instead of changing it as you illustrated. I don't know how I can write a test for this. I'm open to suggestions.

app/controllers/oauth/applications_controller.rb
edwoodward: Method updated

edwoodward:
Admins can now add oauth admins to an Application
admin-oauth-admin-editing

OAuth Admins can see a list of applications with less options than an admin
Admin options
admin-application-list
OAuth Admin options
oauth-admin-app-list

OAuth Admins can only edit the Redirect URI field

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,57 @@ def set_user
...
+      end
+    
+      member_ids = params[:member_ids].split.map(&:to_i)
+      member_ids.each do |member|
+        if !User.where(id: member).exists?
+          @application.errors.add(:owner, "#{member} is not a valid user id")
+          return false
+        end
+      end
+      return member_ids.unshift(@application.owner.owner_ids[0])
```
jpslav, line 161: Should this be code that is in the group logic?  Seems not the best for outsider users to have to remember to do this.

If the code has to live here, can you add a comment?  Future readers will not know what is happening.

app/controllers/oauth/applications_controller.rb
jpslav: Can you say why the owner ID / member ID stuff is happening?  @Dantemss and I think something here is not right (e.g. he said "there's no reason the first owner has to be a member too as far as I remember")

config/initializers/doorkeeper_models.rb
```diff
@@ -16,6 +16,8 @@
 
   has_many :security_logs, inverse_of: :application
 
+  delegate :member_ids, to: :owner
```
jpslav, line 19: is this being used? if not, use it or else delete it

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -158,17 +161,11 @@ module Oauth
...
+      controller.sign_in! not_oauth_admin_user
       get(:index)
       expect(response).to have_http_status :forbidden
     end
 
-    it "should not let a user get his own application" do
-      controller.sign_in! user
-      get(:show, params: { id: untrusted_application_user.id })
-      expect(response).to have_http_status :forbidden
-    end
```
jpslav, line 170: Put this spec back, change it to succeeding and remove "not" from title?  (because a user can get his app)

spec/controllers/oauth/applications_controller_spec.rb
```diff
@@ -194,12 +191,6 @@ module Oauth
       expect(response).to have_http_status :forbidden
     end
 
-    it "should not let a user edit his own application" do
-      controller.sign_in! user
-      get(:edit, params: { id: untrusted_application_user.id })
-      expect(response).to have_http_status :forbidden
-    end
```
jpslav, line 201: Put this spec back, change to success, and change title? (because a user can edit his app)

app/controllers/oauth/applications_controller.rb
```diff
@@ -36,51 +38,61 @@ def create
...
+            end
           end
+          raise ActiveRecord::Rollback
         end
       end
     end
 
     def edit
       OSU::AccessPolicy.require_action_allowed!(:update, @user, @application)
+      get_current_member_ids
```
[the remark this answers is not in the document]
edwoodward: It should be setting @member_ids which it is now. My local server didn't pick up the my code changes and everything still worked using the old code.

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,57 @@ def set_user
...
+        params.require(:doorkeeper_application).permit(
+          :redirect_uri
+        )
+      end
+    end
+
+    def add_application_owners
+      member_ids = validated_member_ids
+      return false if @application.errors.any?
+      return true if !current_user.is_administrator?
```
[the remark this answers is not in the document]
edwoodward: The method needs to return at this point becuase member_ids are not returned as part of params when an oauth_admin is updating. I added an additional check for the user having access to the application.

app/controllers/oauth/applications_controller.rb
edwoodward: I had a comment, but it got lost in the other code changes. It is added back. I don;t think this belongs in the group since it is specific to editing the member_ids.

app/controllers/oauth/applications_controller.rb
edwoodward: I was mainly doing that because of how the data looked on my local in the rails console. The owner was always first. I changed it so it is just added back.

config/initializers/doorkeeper_models.rb
edwoodward: Removed

app/controllers/oauth/applications_controller.rb
```diff
@@ -36,51 +38,61 @@ def create
...
-        end
-      else
-        respond_to do |format|
-          format.html { render :new }
-          format.json do
-            render json: { errors: @application.errors.full_messages },
-                   status: :unprocessable_entity
+      Doorkeeper::Application.transaction do
+        if add_application_owners && @application.save
+          @application.owner.save
```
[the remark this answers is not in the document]
edwoodward: Removed

app/controllers/oauth/applications_controller.rb
```diff
@@ -36,51 +38,61 @@ def create
...
-      else
-        respond_to do |format|
-          format.html { render :edit }
-          format.json do
-            render json: { errors: @application.errors.full_messages },
-                   status: :unprocessable_entity
+      
+      Doorkeeper::Application.transaction do
+        if add_application_owners && @application.update_attributes(app_params)
+          @application.owner.save
```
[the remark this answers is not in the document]
edwoodward: removed

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +121,57 @@ def set_user
...
+    def validated_member_ids
+      return [] if !params[:member_ids].present?
+    
+      if !params[:member_ids].match(SPACE_SEPARATED_NUMBERS_REGEX)
+        @application.errors.add(:owner, 'Member Ids must be a space separated list of integers')
+        return false
+      end
+    
+      member_ids = params[:member_ids].split.map(&:to_i)
+      member_ids.each do |member|
```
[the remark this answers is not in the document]
edwoodward: Done

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: Done

spec/controllers/oauth/applications_controller_spec.rb
edwoodward: Done

app/controllers/oauth/applications_controller.rb
```diff
@@ -109,12 +119,53 @@ def set_user
...
+      member_ids.each do |member_id|
+        if !User.where(id: member_id).exists?
+          @application.errors.add(:owner, "#{member_id} is not a valid user id")
+          return false
+        end
+      end
+      return member_ids
+    end
+
+    def get_current_member_ids
```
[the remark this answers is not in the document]
edwoodward: Fixed

PR #834, opened 2020-05-06
Fix: Student Sign Up redirects to Tutor

BryanEliDimas:
Fix https://github.com/openstax/business-intel/issues/1110

We actually do want to pass-in that param to the function because when it's not present, it gets cleared.


BryanEliDimas:
Not 100% sure if I should create a feature test for this one 🤔 . The tradeoff is that feature tests are "expensive"

PR #844, opened 2020-05-22
Allow submitting other educator role name

BryanEliDimas:
QA [found a problem](https://openstax.slack.com/archives/C6NRD0CCQ/p1590175423228100) with submitting Educator step 4 form. This PR fixes the immediate problem.

The long term problem will be solved by https://github.com/openstax/business-intel/issues/1144

### DIFF
--- a/app/controllers/newflow/login_signup_controller.rb
+++ b/app/controllers/newflow/login_signup_controller.rb
@@ -80,6 +80,9 @@ module Newflow
           redirect_to signup_done_path
         },
         failure: lambda {
+          @book_subjects = book_data.subjects
+          @book_titles = book_data.titles
+
           render :educator_profile_form
         }
       )
--- a/app/handlers/newflow/educator_complete_profile.rb
+++ b/app/handlers/newflow/educator_complete_profile.rb
@@ -17,13 +17,6 @@ module Newflow
           in: %w(instructor administrator other),
         }
       )
-
-      validates(
-        :num_students_per_semester_taught,
-        numericality: {
-          only_integer: true,
-        }
-      )
     end
 
     protected ###############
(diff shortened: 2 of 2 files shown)


app/controllers/newflow/login_signup_controller.rb
```diff
@@ -80,6 +80,9 @@ def educator_complete
           redirect_to signup_done_path
         },
         failure: lambda {
+          @book_subjects = book_data.subjects
```
steveburkett, line 83: is book_data guaranteed to not be nil?

app/handlers/newflow/educator_complete_profile.rb
```diff
@@ -67,6 +60,10 @@ def push_lead_to_salesforce
       end
     end
 
+    def kind_of_educator
+      signup_params.educator_specific_role == 'other' ? signup_params.other_role_name : signup_params.educator_specific_role
```
steveburkett, line 64: consider putting 'other' as a constant.

consider a method name of `educator_type`? (subtle..just wondering if this sounds better than "kind_of". either way is fine :)

steveburkett: does the role need to be stripped? (any chance this role could come in w/ leading or trailing spaces?  it's good to nip this in the bud if so and trim the inputs (always)

app/controllers/newflow/login_signup_controller.rb
BryanEliDimas: Not really but that is something I am working on improving in a separate PR: https://github.com/openstax/accounts/pull/840

app/handlers/newflow/educator_complete_profile.rb
BryanEliDimas: These are good ideas I would have implemented before hitting the merge button. I can definitely implement them in another PR which I'm currently working on unless you think I should reverse this PR

PR #846, opened 2020-05-28
Read more Salesforce fields

Dantemss:
- Read school_location and child_of_kip from SF School/Account
- Added home school school_type
- Refreshed cassettes
- Specs for the new stuff

Needed for https://github.com/openstax/tutor-server/pull/2074

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -121,6 +121,7 @@ module Admin
       @user.role = params[:user][:role] if params[:user][:role]
       @user.faculty_status = params[:user][:faculty_status] if params[:user][:faculty_status]
       @user.school_type = params[:user][:school_type] if params[:user][:school_type]
+      @user.school_location = params[:user][:school_location] if params[:user][:school_location]
       @user.is_kip = params[:user][:is_kip]
       if @user.external_uuids.any? && params[:user][:keep_external_uuids] == '0'
         @user.external_uuids.destroy_all
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -122,4 +122,21 @@ class AnonymousUser
     raise AnonymousUserIsImmutableError, "Cannot set school type on the AnonymousUser."
   end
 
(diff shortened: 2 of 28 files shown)


Dantemss:
I can't seem to get Travis to run for some reason.

app/models/anonymous_user.rb
```diff
@@ -122,4 +122,21 @@ def school_type=(type)
...
+  end
+
+  def school_location
+    User::DEFAULT_SCHOOL_LOCATION.to_s
+  end
+
+  def school_location=(type)
+    raise AnonymousUserIsImmutableError, "Cannot set school location on the AnonymousUser."
+  end
 end
```
nathanstitt, line 142: will school locations always be valid ruby method names?  For instance this would fail if they had a `'` or something in the name

Dantemss: They are an enum so they have to be valid... The 3 we have currently are:
```
unknown_school_location
domestic_school
foreign_school
```

Dantemss:
1. school_location is separate from school_type in SF
school_type can be set to home_school
school_location is domestic_school or foreign_school
any school type can have a school_location, it's just whatever is set in SF

2. Oddly enough, there's already a different field for this which has a location as coordinates but there are only like 2 records that have values there. So it would be a different field. Hopefully the accounts-rails gem handles unknown enums, but that is untested.

We should not rename because the field is not specific to home schools.

app/controllers/admin/users_controller.rb
```diff
@@ -121,6 +121,7 @@ def update_user
       @user.role = params[:user][:role] if params[:user][:role]
       @user.faculty_status = params[:user][:faculty_status] if params[:user][:faculty_status]
       @user.school_type = params[:user][:school_type] if params[:user][:school_type]
+      @user.school_location = params[:user][:school_location] if params[:user][:school_location]
```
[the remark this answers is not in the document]
Dantemss: My secret is I just find all instances of school_type and go to all of them and add school_location too...

PR #848, opened 2020-06-01
SheerID webhook

BryanEliDimas:
Create webhook for SheerID
https://github.com/openstax/business-intel/issues/871

### DIFF
--- a/app/controllers/newflow/base_controller.rb
+++ b/app/controllers/newflow/base_controller.rb
@@ -2,44 +2,13 @@ module Newflow
   class BaseController < ApplicationController
     include ApplicationHelper
 
-    skip_forgery_protection(only: :sheerid_webhook)
-    skip_before_action :authenticate_user!, :check_if_password_expired
+    layout 'newflow_layout'
 
-    before_action :newflow_authenticate_user!, only: :profile_newflow
+    skip_before_action :authenticate_user!
+
     before_action :set_active_banners
 
-    def profile_newflow
-      render layout: 'application'
-    end
-
-    # SheerID makes a POST request to this endpoint when it verifies an educator
-    # http://developer.sheerid.com/program-settings#webhooks
-    def sheerid_webhook
-      handle_with(
-        SheeridWebhook,
-        success: lambda {
-          render(status: :ok)
-        }, failure: lambda {
-          render(status: :unprocessable_entity)
-        }
-      )
-    end
-
-    def exit_accounts
-      if (redirect_param = extract_params(request.referrer)[:r])
-        if Host.trusted?(redirect_param)
-          redirect_to(redirect_param)
-        else
-          raise Lev::SecurityTransgression
-        end
-      elsif !signed_in? && (redirect_uri = extract_params(stored_url)[:redirect_uri])
-        redirect_to(redirect_uri)
-      else
-        redirect_back # defined in the `action_interceptor` gem
-      end
-    end
-
-    private #################
+    protected #################
 
     def set_active_banners
       return unless request.get?
--- a/app/controllers/newflow/base_controller.rb
+++ b/app/controllers/newflow/base_controller.rb
@@ -2,6 +2,7 @@ module Newflow
   class BaseController < ApplicationController
     include ApplicationHelper
 
+    skip_forgery_protection(only: :sheerid_webhook)
     skip_before_action :authenticate_user!, :check_if_password_expired
 
     before_action :newflow_authenticate_user!, only: :profile_newflow
@@ -9,6 +10,19 @@ module Newflow
 
     def profile_newflow
       render layout: 'application'
+    end
+
+    # SheerID makes a POST request to this endpoint when it verifies an educator
+    # http://developer.sheerid.com/program-settings#webhooks
+    def sheerid_webhook
+      handle_with(
+        SheeridWebhook,
+        success: lambda {
+          render(status: :ok)
+        }, failure: lambda {
+          render(status: :unprocessable_entity)
+        }
+      )
     end
 
     def exit_accounts
--- a/app/routines/newflow/update_user_from_sheerid_webhook.rb
+++ b/app/routines/newflow/update_user_from_sheerid_webhook.rb
@@ -12,46 +12,48 @@ module Newflow
 
       last_response = details.fetch('lastResponse')
       sheer_id_status = last_response.fetch('currentStep')
-      user_sheerid_info = details.fetch('personInfo')
+      sheerid_person_info = details.fetch('personInfo')
 
-      email = user_sheerid_info.fetch('email')
-      outputs[:user] = user = User.find_by(sheerid_verification_id: verification_id)
+      outputs.sheerid_email = sheerid_person_info.fetch('email')
+      outputs.existing_email = EmailAddress.find_by(value: outputs.sheerid_email)
+      outputs.user = user = User.find_by(sheerid_verification_id: verification_id)
 
-      fatal_error(code: :email_mismatch) if email_mismatch?(email, user)
+      capture_mismatch_error! if email_mismatch?
 
-      user.update!(
-        first_name: user_sheerid_info.fetch('firstName'),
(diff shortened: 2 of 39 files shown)


app/routines/newflow/update_user_from_sheerid_webhook.rb
```diff
@@ -0,0 +1,58 @@
...
+      create_security_log(user, details)
+    end
+
+    private #################
+
+    # true if no user found with such email or if that email email address belongs to someone else
+    def email_mismatch?(email, user)
+      existing_email = EmailAddress.find_by(value: email)
+
+      if !user.present? || !existing_email.present? || existing_email.user_id != user.id
```
steveburkett, line 37: !user.present? might be a tiny bit clearer as user.blank?

app/controllers/newflow/base_controller.rb
```diff
@@ -11,6 +12,19 @@ def profile_newflow
       render layout: 'application'
     end
 
+    # SheerID makes a POST request to this endpoint when it verifies an educator
+    # http://developer.sheerid.com/program-settings#webhooks
+    def sheerid_webhook
```
steveburkett, line 17: you want this on the basecontroller, available to all controllers?  should it be off on its own?

app/routines/newflow/update_user_from_sheerid_webhook.rb
steveburkett: a code comment wouldnt have been necessary if this big IF statement was written as a method and then self-documenting

config/initializers/sheerid_ip_whitelist.rb
```diff
@@ -0,0 +1,11 @@
+SHEERID_IP_WHITELIST = [
+  # production environment
+  '54.236.234.189',
```
steveburkett, line 3: you really want a whitelist committed to git?

lib/sheerid_api.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  PROGRAM_ID = ENV['SHEERID_PROGRAM_ID']
+  BASE_URL = 'https://services.sheerid.com/rest/v2'
+
+  WEBHOOKS_URL = "#{BASE_URL}/program/#{PROGRAM_ID}/webhook"
+  INFO_URL = "#{BASE_URL}/info"
+  PROGRAM_THEME_URL = "#{BASE_URL}/#{PROGRAM_ID}/theme"
+
+  class << self
+    def get_verification_details(verification_id)
```
steveburkett, line 20: imo, for a longer list of methods, it's better to do the def self.x way of class methods instead of this.  a long train of these makes is harder to follow

lib/sheerid_api.rb
```diff
@@ -0,0 +1,65 @@
...
+    private ###################
+
+    def make_request(http_method, url, body = nil)
+      begin
+        response = Faraday.send(http_method, url, body, HEADERS)
+        body = JSON.parse(response.body)
+        return body
+      rescue Net::ReadTimeout => ee
+        Raven.capture_message("SheeridAPI: timeout")
+        return nil
```
steveburkett, line 34: returning nil is a bit of an anti-practice. better to return a object that has meaning (e.g, SheerApiResult).

steveburkett: also, returns are needed here

spec/controllers/newflow/base_controller_spec.rb
```diff
@@ -27,6 +27,36 @@ module Newflow
       end
     end
 
+    describe 'POST #sheerid_webhook' do
+      let(:handler) { SheeridWebhook }
+
+      let(:params) do
+        {
```
steveburkett, line 34: imo, this should be one line.  stuff like this creates lots and lots of vertical space and over time, gets harder to read

spec/routines/newflow/update_user_from_sheerid_webhook_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'rails_helper'
+require 'vcr_helper'
+
+module Newflow
+  RSpec.describe UpdateUserFromSheeridWebhook, type: :routine, vcr: VCR_OPTS  do
+    context 'when success' do
+      subject(:run) { described_class.call(verification_id: verification_id) }
```
steveburkett, line 7: suggestion to name the subject better here than run.   what is the object under test?  a subject would the sheer_id_webhook?  then, you'd do an action on the subject in the test.   

for example, take this test for Person.  the subject is the object under test (the Person).  sometimes you might not use one since it's implicit, but often an explicit subject is helpful.  

the subject is the (noun) object under test. 

```
describe Person do
  context "born 19 years ago" do
    subject { Person.new(:birthdate => 19.years.ago }
    it { should be_eligible_to_vote }
    its(:age) { should == 19 }
    it "should be younger than a 20 year old" do
      twenty_year_old = Person.new(:birthday => 20.years.ago)
      subject.should be_younger_than(twenty_year_old)
    end
  end
end
```

or 

```
describe "named subject" do
  subject(:a) { A.new }
  it "has been instantiated" do
    expect(a).to be_an(A)
  end
end
```

spec/routines/newflow/update_user_from_sheerid_webhook_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      it 'confirms their faculty status' do
+        expect(educator.confirmed_faculty?).to be_falsey
+        run
+        educator.reload
+        expect(educator.confirmed_faculty?).to be_truthy
+      end
+
+      it 'creates a security log' do
+        expect {
+          run
```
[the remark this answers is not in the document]
BryanEliDimas: It does look much better. Fixed: `expect { run }.to change(SecurityLog, :count).by(1)`

app/routines/newflow/update_user_from_sheerid_webhook.rb
BryanEliDimas: True. Good point. I rewrote this part. Although, I'm not sure that I totally like the one-liner I turned it into 😅 . Will ask for re-review.

app/controllers/newflow/base_controller.rb
BryanEliDimas: Sure, sounds like a good idea. Created an `OtherController`.

config/initializers/sheerid_ip_whitelist.rb
BryanEliDimas: I don't think it's the worst thing in the world. I believe this is public information anyway.

lib/sheerid_api.rb
BryanEliDimas: I'm not sure I understand 🤔

spec/controllers/newflow/base_controller_spec.rb
BryanEliDimas: improved.

spec/routines/newflow/update_user_from_sheerid_webhook_spec.rb
BryanEliDimas: Thank you for explaining that very well 😃 . I _did_ mean "run" as a noun, in this case. Like, "go for a run." I can't think of a better name for this SUT. I'm open to specific suggestions 😄 .

lib/sheerid_api.rb
BryanEliDimas: I gave it a shot last week and I reverted it because I didn't like it too much 🤷‍♂️ . One thing I wasn't sure about, for example, is where to define this "null object". I'm familiar with the Null Object pattern :).

lib/sheerid_api.rb
```diff
@@ -0,0 +1,65 @@
+# Service object for interacting with SheerID's API
+# docs found at: http://developer.sheerid.com/rest-api
+class SheeridAPI
+  DEFAULT_WEBHOOK_URL = 'https://dev.openstax.org/accounts/i/sheerid/webhook'
```
mwharrison, line 4: lets remove this if it's not used

PR #850, opened 2020-06-04
Fix couple of UX items

BryanEliDimas:
https://github.com/openstax/business-intel/issues/918

### DIFF
--- a/app/controllers/newflow/educator_signup_controller.rb
+++ b/app/controllers/newflow/educator_signup_controller.rb
@@ -33,6 +33,7 @@ module Newflow
 
     def educator_change_signup_email_form
       @email = unverified_user.email_addresses.first.value
+      @total_steps = 4
     end
 
     def educator_change_signup_email
--- a/app/views/newflow/educator_signup/educator_change_signup_email_form.html.erb
+++ b/app/views/newflow/educator_signup/educator_change_signup_email_form.html.erb
@@ -1,8 +1,17 @@
+<%
+    current_step = I18n.t(
+        :"login_signup_form.step_counter",
+        current_step: 2,
+        total_steps: @total_steps
+    )
+%>
+
 <%=
     render(
         partial: 'newflow/signup/change_signup_email_form_template',
         locals: {
             submit_to_path: educator_change_signup_email_path,
+            current_step: current_step,
             email_address_value: @email
         }
     )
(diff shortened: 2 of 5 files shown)


app/views/newflow/educator_signup/educator_change_signup_email_form.html.erb
```diff
@@ -1,8 +1,17 @@
+<%
+    current_step = I18n.t(
+        :"login_signup_form.step_counter",
+        current_step: 2,
+        total_steps: @total_steps
+    )
```
steveburkett, line 6: the current step is another one of those pieces of state data that would be good to expose to the views thru a data object as a PORO.   fine i guess for here...but just sayin

app/controllers/newflow/educator_signup_controller.rb
```diff
@@ -33,6 +33,7 @@ def educator_signup
 
     def educator_change_signup_email_form
       @email = unverified_user.email_addresses.first.value
+      @total_steps = 4
```
steveburkett, line 36: personally, i really dislike "magic numbers" (better to put all this state stuff into a PORO that's passed as a component of a view-controller interface adapter object.  but that would be a lot more work.

app/views/newflow/educator_signup/educator_change_signup_email_form.html.erb
BryanEliDimas: Right; every single variable in the view(s) between `<%` and `%>` CAN be eventually put into a ViewModel — as part of the "Model–view–viewmodel" concept.

At this time, the whole Accounts app is an MVC though. Future improvement for sure 👍 .

app/controllers/newflow/educator_signup_controller.rb
BryanEliDimas: Would it be worse or better if there was no instance variable for the number 4. I feel like I went one (small) step beyond just putting it in the view—by putting it as an instance variable. We're still (as a team) not totally decided if we are ever going to skip the entire step 3 or... not, that's still in the works but yeah, for now I think that it doesn't hurt to keep this instance variable—for now.

BryanEliDimas: Again, for the record, I also dislike "magic numbers" and I agree about creating a PORO (a ModelView object), but those are small improvements I hope/plan to do soon in a future PR.

PR #855, opened 2020-06-11
Added opt_out_of_cookies field to admin user editing and API

edwoodward:


app/models/user.rb
```diff
@@ -171,6 +171,10 @@ def is_test?
     !!is_test
   end
 
+  def opt_out_of_cookies?
+    !!opt_out_of_cookies
+  end
+
```
BryanEliDimas, line 177: I think defining this is not necessary, actually, because Rails essentially detects that the DB column is a boolean and then it automagically responds to the method with the same but with a question mark at the end. In this case, `opt_out_of_cookies?`. Yay, Rails magic! 😅

db/migrate/20200610181353_add_opt_out_of_cookies_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddOptOutOfCookiesToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :opt_out_of_cookies, :boolean, default: false
```
BryanEliDimas, line 3: I have come to learn that it can — sometimes — be useful to _not_ add a default value to boolean flags unless the default is `true`. That's because, for example, **possibly** with this flag, we may want to query for users whose flag has not been updated and whose flag has not been explicitly been set to to `false`. 

`nil` is falsey anyway, so that's why not setting a default to `false` would work :).

Just something to consider.

db/schema.rb
```diff
@@ -247,10 +248,10 @@
...
     t.string "lead_application_source", default: "", null: false
     t.boolean "confidential", default: true, null: false
-    t.boolean "can_access_private_user_data", default: false, null: false
-    t.boolean "can_find_or_create_accounts", default: false, null: false
-    t.boolean "can_message_users", default: false, null: false
-    t.boolean "can_skip_oauth_screen", default: false, null: false
+    t.boolean "can_access_private_user_data", default: false
+    t.boolean "can_find_or_create_accounts", default: false
+    t.boolean "can_message_users", default: false
+    t.boolean "can_skip_oauth_screen", default: false
```
BryanEliDimas, line 254: I don't think these should've changed. Not in this PR at least. 

🤔 ... could it be my fault from a previous PR? I would suggest recreating your local DB.

edwoodward: Not sure where this came from. It was not part of the conflict with master, but is fixed now.

db/migrate/20200610181353_add_opt_out_of_cookies_to_users.rb
edwoodward: We want this to explicitly be false unless someone sets it to true in the admin which is why I added the default value.

app/models/user.rb
edwoodward: I originally did not define it and the admin update did not work. However, I retested it without the method and the field updated. It is removed.

db/schema.rb
```diff
@@ -361,4 +362,4 @@
   add_foreign_key "oauth_access_grants", "oauth_applications", column: "application_id"
   add_foreign_key "oauth_access_tokens", "oauth_applications", column: "application_id"
   add_foreign_key "users", "oauth_applications", column: "source_application_id"
-end
+end
```
BryanEliDimas, line 365: Please add a newline at the end of the line. It might be a setting in your editor btw; I know there is one for VSCode and — if I remember correctly — for Sublime text as well, to do automatically.

edwoodward: Why should there be a new line?

BryanEliDimas: It's a best practice in Ruby per the Ruby style guide. https://rubystyle.guide/#newline-eof

edwoodward: New line added and I configured Rubymine to add it in the future.

edwoodward:
@BryanEliDimas New test added to user_representer_spec.

PR #859, opened 2020-06-18
Verify instructors using SheerID, sync with Salesforce

BryanEliDimas:
https://github.com/openstax/business-intel/issues/871


app/routines/newflow/create_salesforce_lead.rb
```diff
@@ -0,0 +1,101 @@
...
+    end
+
+    def referring_app_for(user)
+      user&.source_application&.lead_application_source || 'Accounts'
+    end
+
+    def store_salesforce_lead_id(user, lead_id)
+      fatal_error(code: :lead_id_is_blank, message: :lead_id_is_blank.to_s.titleize) if lead_id.blank?
+
+      return true if user.salesforce_lead_id.present? # nothing to do
```
steveburkett, line 64: suggestion.  returning boolean here means the method is a question.  if so, the method name should end in a ?.

if it's not, i suggest returning a success/error object that contains better description of what success or failure means.  perhaps failure means that the object might log the errors when done. or something like that.

app/routines/newflow/create_salesforce_lead.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      user.salesforce_lead_id = lead_id
+
+      if user.save
```
steveburkett, line 68: passing in user here is kinda dangerous eh?  a huge object. consider
```
create_sales_force_lead do
   user.save
end # the end of this yield auto creates a security log.  the return of the yield could create the failed signup
```

app/routines/newflow/educator_signup/activate_account.rb
```diff
@@ -0,0 +1,33 @@
...
+          event_data: {
+            user_became_activated: 'true'
+          }
+        )
+      end
+
+      private ###############
+
+      def create_salesforce_lead_for(user)
+        CreateSalesforceLead.perform_later(user: user)
```
steveburkett, line 29: what happens if the creation of the salesforce lead fails in the activation of an account?  how would this be detected & reconciled?

app/routines/newflow/process_sheerid_webhook_request.rb
```diff
@@ -0,0 +1,30 @@
...
+    uses_routine UpsertSheeridVerification
+
+    protected ###############
+
+    def exec(verification_id:)
+      status.set_job_name(self.class.name)
+      status.set_job_args(verification_id: verification_id)
+
+      verification = upsert_verification(verification_id)
+      if verification.errors.none? && verification.verified? && (user = EmailAddress.verified.find_by(value: verification.email)&.user)
```
steveburkett, line 15: suggestion...just put the assignment of user = on a different line, and do a `&& user`

app/routines/newflow/update_salesforce_lead.rb
```diff
@@ -0,0 +1,59 @@
...
+    protected #################
+
+    def exec(user:)
+      status.set_job_name(self.class.name)
+      status.set_job_args(user: user.to_global_id.to_s)
+
+      lead_id = user.salesforce_lead_id
+      lead = fetch_lead(lead_id)
+
+      if not lead.present?
```
steveburkett, line 14: it's standard to use the ! for not.  or even better, turn to an affirmative `if lead.blank?`

app/routines/newflow/update_salesforce_lead.rb
```diff
@@ -0,0 +1,59 @@
...
+      OpenStax::Salesforce::Remote::Lead.find(lead_id)
+    end
+
+    def update_salesforce_lead!(lead, user)
+      lead.update(
+        # sheerid_verification_status: 'verified' # TODO: future feature that must be enabled by SF
+        first_name: user.first_name,
+        last_name: user.last_name,
+        school: user.sheerid_reported_school || user.self_reported_school || 'unknown',
+        # role: user.role,
```
steveburkett, line 38: intentional to leave these commented lines in?

app/routines/newflow/update_salesforce_lead.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def log_error(user, lead)
+      message = "#{self.class.name} ERROR; Lead (#{lead&.id}) for user (#{user.id}); Error: #{lead&.errors&.full_messages}"
+      Rails.logger.warn(message)
+      send_to_raven(message)
+    end
+
+    def send_to_raven(message)
+      Raven.capture_message(message)
```
steveburkett, line 56: does this one-liner do much else? used anywhere else? seems a bit much to just create a method for this.

app/routines/newflow/update_salesforce_lead.rb
```diff
@@ -0,0 +1,59 @@
...
+      )
+    end
+
+    def log_success(user, lead)
+      message = "#{self.class.name} SUCCESS (#{lead.id}) for user (#{user.id})"
+      Rails.logger.info(message)
+    end
+
+    def log_error(user, lead)
+      message = "#{self.class.name} ERROR; Lead (#{lead&.id}) for user (#{user.id}); Error: #{lead&.errors&.full_messages}"
```
steveburkett, line 50: consider: creating a rails log object that encapsulates the prefixing like this.. otherwize...a new dev might forget to follow this standard.

app/routines/newflow/verify_educator.rb
```diff
@@ -0,0 +1,88 @@
...
+    lev_routine active_job_enqueue_options: { queue: :educator_signup_queue }, use_jobba: true
+    uses_routine UpsertSheeridVerification
+    uses_routine UpdateSalesforceLead
+
+    protected ###############
+
+    def exec(verification_id:, user:)
+      status.set_job_name(self.class.name)
+      status.set_job_args(verification_id: verification_id.to_s, user: user.to_global_id.to_s)
+
```
steveburkett, line 14: is all this unit tested?

app/routines/newflow/verify_educator.rb
```diff
@@ -0,0 +1,88 @@
...
+    def fetch_verification(verification_id)
+      @fetch_verification ||= run(UpsertSheeridVerification, verification_id: verification_id).outputs.verification
+    end
+
+    def update_salesforce_lead_for(user)
+      run(UpdateSalesforceLead, user: user).outputs.lead.errors.none?
+    end
+
+    def email_mismatch?(user, email)
+      user.blank? || email.blank? || email.user_id != user.id
```
steveburkett, line 54: this could all be simplified to 
```
email&.user_id != user&.id
```

spec/handlers/newflow/student_signup/signup_form_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+
+module Newflow
+  module StudentSignup
+    describe SignupForm, type: :handler do
+      context 'when success' do
+        before(:all) do
+          load('db/seeds.rb')
+        end
+
+        let(:subject) do
```
steveburkett, line 11: this isnt really a subject, and can be too easily confused with rspec's subject DSL method.

maybe rename to:

```
       let(:handler_call) do
          described_class.call(params: params)
        end
```

app/handlers/newflow/educator_signup/complete_profile.rb
```diff
@@ -0,0 +1,125 @@
...
+      def kind_of_educator
+        signup_params.educator_specific_role == OTHER ? signup_params.other_role_name : signup_params.educator_specific_role
+      end
+
+      def books_or_subjects
+        subjects = signup_params.subjects_of_interest&.reject(&:empty?)
+        titles = signup_params.books_used&.reject(&:empty?)
+
+        if subjects&.any?
+          return subjects.join(';')
```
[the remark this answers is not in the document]
BryanEliDimas: Fixed. Removed it.

app/handlers/newflow/educator_signup/signup_form.rb
```diff
@@ -0,0 +1,113 @@
...
+        message = I18n.t(:"login_signup_form.#{code}")
+        nonfatal_error(
+          code: code,
+          message: message,
+          offending_inputs: field
+        )
+      end
+
+      def create_user
+        user = User.create(
```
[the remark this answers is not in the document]
BryanEliDimas: Considered it but I don't think that's necessary. Everywhere a User record is created in Accounts, the key/values are different, that why.

app/handlers/newflow/student_signup/signup_form.rb
```diff
@@ -0,0 +1,111 @@
...
+        message = I18n.t(:"login_signup_form.#{code}")
+        nonfatal_error(
+          code: code,
+          message: message,
+          offending_inputs: field
+        )
+      end
+
+      def create_user
+        user = User.create(
```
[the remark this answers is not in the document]
BryanEliDimas: Right. Yes. This is for **student** signup though.

app/routines/newflow/verify_educator.rb
BryanEliDimas: This is part of the Lev gem which we own and that gem is unit tested. This part doesn't anything besides making more informative log entries, I believe, also the reason why it wasn't unit tested in the legacy flow which has been running for years.

app/routines/newflow/update_salesforce_lead.rb
BryanEliDimas: It's not a standard but if it was, I would make some kind of object.

app/routines/newflow/verify_educator.rb
BryanEliDimas: If both are `nil` then it would still evaluate to true, so that's why I decided to make it this way. But if the rest of the logic is correct, then user should never be nil. In that case, you might be right. Fixed.

app/handlers/newflow/verify_email_by_pin.rb
```diff
@@ -26,9 +24,13 @@ def handle
...
 
       claiming_user = options[:email_address].user
-      run(ActivateUser, claiming_user)
-
+      activate_user(claiming_user)
       outputs.user = claiming_user
     end
+
+    def activate_user(claiming_user)
+      raise('Must implement')
```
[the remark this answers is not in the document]
BryanEliDimas: Considered. That's a good idea if we create a gem out of this lib, or a rails Engine.

spec/handlers/newflow/student_signup/signup_form_spec.rb
BryanEliDimas: Fixed.

app/handlers/newflow/student_signup/signup_form.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        if LookupUsers.by_verified_email(signup_params.email).first
+          fatal_error(
+            code: :email_taken,
+            message: I18n.t(:"login_signup_form.email_address_taken"),
+            offending_inputs: :email
+          )
+        end
+
+        if options[:user_from_signed_params].present?
```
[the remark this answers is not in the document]
BryanEliDimas: Yes.

app/routines/newflow/create_salesforce_lead.rb
BryanEliDimas: My intention is to use it just like ActiveRecord's `save` method. But that is a good idea I thought about.

app/routines/newflow/create_salesforce_lead.rb
BryanEliDimas: Indeed, if we didn't own the code that we pass the user record to. But we do. So it's fine and we have been doing that in the legacy flow for years as well; battle-tested.

app/routines/newflow/educator_signup/activate_account.rb
BryanEliDimas: The Lev gem (as magical as it is) takes care of those matters/questions. Most importantly, it wraps everything in a transaction and nothing gets committed unless all of it is able to be committed. Also the job is retried automatically and fails after a few attempts.

app/routines/newflow/process_sheerid_webhook_request.rb
BryanEliDimas: Considered.

app/routines/newflow/update_salesforce_lead.rb
BryanEliDimas: replaced with `if lead.blank?` 👍

app/routines/newflow/update_salesforce_lead.rb
BryanEliDimas: Fixed.

app/routines/newflow/update_salesforce_lead.rb
BryanEliDimas: Agreed. Fixed.

app/controllers/legacy/users_controller.rb
```diff
@@ -31,12 +31,6 @@ def update
 
     private
 
-    def prevent_caching
```
mwharrison, line 34: why was only this removed from the legacy user controller? (I realize none of it should be getting used anymore)

app/controllers/newflow/login_controller.rb
```diff
@@ -24,13 +24,14 @@ def login
...
-          save_login_failed_email(@handler_result.outputs.email)
+          email = @handler_result.outputs.email
+          save_login_failed_email(email)
 
           code = @handler_result.errors.first.code
           case code
           when :cannot_find_user, :multiple_users, :incorrect_password, :too_many_login_attempts
             user = @handler_result.outputs.user
-            security_log(:sign_in_failed, { reason: code, user: user }) # also store email?
+            security_log(:sign_in_failed, { reason: code, email: email })
```
mwharrison, line 34: what's the reason behind changing this to email? it concerns me for GDPR reasons, if we get a delete request, we are not supposed to remove all user PII data from our systems, user would be more abstract.

app/handlers/newflow/confirm_oauth_info.rb
```diff
@@ -17,6 +17,7 @@ class ConfirmOauthInfo
       validates :first_name, presence: true
       validates :last_name, presence: true
       validates :email, presence: true
+      validates :terms_accepted, presence: true
```
mwharrison, line 20: 🎉

app/handlers/newflow/student_signup/signup_form.rb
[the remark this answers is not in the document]
mwharrison: couldn't you just pass the role in? not suggesting for this PR but could be a future enhancement

spec/handlers/newflow/educator_signup/complete_profile_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+            expect(result.errors.count).to eq 1
+            expect(result.errors.first.message).to eq 'Please enter number of students taught'
+          end
+        end
+      end
+
+      context 'when success' do
+        context 'salesforce' do
+          it 'calls UpdateSalesforceLead' do
+            skip 'TODO'
```
mwharrison, line 80: why skip?

spec/handlers/newflow/educator_signup/sheerid_webhook_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+module Newflow
+  module EducatorSignup
+    RSpec.describe SheeridWebhook, type: :handler do
```
mwharrison, line 5: is testing the response handling somewhere too?

spec/handlers/newflow/educator_signup/update_salesforce_lead_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+module Newflow
+  module EducatorSignup
+    describe UpdateSalesforceLead do
+      it "sends the subject properly formatted" do
+        skip 'TODO'
```
mwharrison, line 7: why skip?

app/handlers/newflow/verify_user_email_by_pin.rb
```diff
@@ -26,9 +14,13 @@ def handle
...
 
       claiming_user = options[:email_address].user
-      run(ActivateUser, claiming_user)
-
+      activate_user(claiming_user)
       outputs.user = claiming_user
     end
+
+    def activate_user(claiming_user)
+      raise('Must implement')
```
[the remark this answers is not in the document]
BryanEliDimas: Essentially, `Newflow::VerifyUserEmailByPin` is an _interface_, or base class from which `Newflow::StudentSignup::VerifyEmailByPin` and `Newflow::EducatorSignup::VerifyEmailByPin` inherit from. This is how it's done in Ruby land.

spec/handlers/newflow/educator_signup/complete_profile_spec.rb
BryanEliDimas: No reason other than it was to-do. Implemented it here: https://github.com/openstax/accounts/pull/866/files#diff-819e29219488277a6cd25c171a52c838

spec/handlers/newflow/educator_signup/update_salesforce_lead_spec.rb
BryanEliDimas: No reason other than it was to-do. Implemented it here: https://github.com/openstax/accounts/pull/866/files#diff-819e29219488277a6cd25c171a52c838

spec/handlers/newflow/educator_signup/complete_profile_spec.rb
BryanEliDimas: Implemented here: https://github.com/openstax/accounts/pull/866/files#diff-5c242bf9eee796184d23ed60b3ddb4d0L79-L82

spec/handlers/newflow/educator_signup/sheerid_webhook_spec.rb
BryanEliDimas: Yeah. This? https://github.com/openstax/accounts/pull/859/files#diff-82ae3922ca1f3ca175760ea679b5e3f9R170

app/handlers/newflow/verify_user_email_by_pin.rb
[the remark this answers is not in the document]
mwharrison: Not sure I completely understand 😅 but I'll trust you

app/helpers/application_helper.rb
```diff
@@ -34,6 +34,12 @@ def extract_params(url)
     Addressable::URI.parse(url).query_values.to_h.with_indifferent_access
   end
 
+  def prevent_caching
```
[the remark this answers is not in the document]
BryanEliDimas: Yes. It was only being used in the legacy user controller but now I also use it in the `Newflow::OtherController#profile_newflow`

app/controllers/legacy/users_controller.rb
BryanEliDimas: 👍 . Addressed here: https://github.com/openstax/accounts/pull/859#discussion_r450941766

app/controllers/newflow/login_controller.rb
BryanEliDimas: 1) `user` was actually `nil` there but if it wasn't, then 2) the function `security_log` already captures whatever user is logged in as `current_user`. So that part of that line wasn't doing anything. With `email` now we actually track the attempt to log in with a specific email.

PR #863, opened 2020-07-01
Added js code for tag manager and field for tag manager code in admin

edwoodward:


app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,8 @@
       OX.url_prefix = "<%= request.script_name %>";
     </script>
 
+    <%= render partial: "layouts/google_tag_manager" %>
```
BryanEliDimas, line 11: Did you mean to add this the new flow? https://github.com/openstax/accounts/blob/master/app/views/layouts/newflow_layout.html.erb

Only adding it to app/views/layouts/application.html.erb will make it only available in the legacy flow and users' profile.

app/views/layouts/application.html.erb
```diff
@@ -33,7 +35,7 @@
   </head>
 
   <body class="<%= controller_name %> <%= action_name %>">
-
+    <%= render partial: "layouts/tag_manager_snippet" %>
```
BryanEliDimas, line 38: Same as above: Did you mean to add this the new flow? https://github.com/openstax/accounts/blob/master/app/views/layouts/newflow_layout.html.erb

Only adding it to app/views/layouts/application.html.erb will make it only available in the legacy flow and users' profile.

app/views/layouts/application.html.erb
edwoodward: Good point. Fixed

app/views/layouts/application.html.erb
edwoodward: Fixed

PR #873, opened 2020-07-23
Wrap up educator flow

BryanEliDimas:
.


app/models/anonymous_user.rb
```diff
@@ -9,6 +9,10 @@ def is_administrator?
     false
   end
 
+  def student?
```
mwharrison, line 12: should this be `is_student` like the other methods? (just for consistency sake)

spec/features/newflow/educators/educator_signup_flow_spec.rb
```diff
@@ -0,0 +1,245 @@
+require 'rails_helper'
+
+module Newflow
+
+  feature 'Educator signup flow', js: true do
+
+    background { load 'db/seeds.rb' }
+    before(:each) { turn_on_educator_feature_flag }
+
+    let(:first_name) { "Elon#{Faker::Alphanumeric.alphanumeric(number: 20)}" }
```
mwharrison, line 10: why not just use faker to generate the name instead of appending numbers?

spec/features/newflow/educators/educator_signup_flow_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+    let(:last_name) { "Musk#{Faker::Alphanumeric.alphanumeric(number: 20)}" }
+    let(:phone_number) { Faker::PhoneNumber.phone_number }
+    let(:email_value) { "#{Faker::Alphanumeric.alphanumeric(number: 20)}@rice.edu" }
```
mwharrison, line 13: same here, faker can generate emails

app/models/anonymous_user.rb
BryanEliDimas: No because note that this model is just for "anonymous" users — meaning that they are not logged in. And the `student?` method is not explicitly defined in the actual `User` model (logged in users)... Rails defines it for you with Ruby metaprogramming, so this `student?` method in the anon user just follows that convention, unlike the rest of the methods which Ruby does not define with metaprogramming `is_.*?` — those are different.

spec/features/newflow/educators/educator_signup_flow_spec.rb
BryanEliDimas: Good idea. Done.

spec/features/newflow/educators/educator_signup_flow_spec.rb
BryanEliDimas: Faker generates emails, yes, but they are short words which could be a problem because SheerID requires user information to be unique, otherwise it gives a Verification Limit Exceeded error.

mwharrison: There is a unique attribute you can add to emails (https://github.com/faker-ruby/faker/blob/master/doc/default/internet.md)
See: https://github.com/faker-ruby/faker#ensuring-unique-values

mwharrison: Like I said, I'm okay with this - just sharing the realm of possibilities. :)

BryanEliDimas: Yeah, but that's only for every time that the tests run. And I did actually just add `unique` to `Faker::Alphanumeric.unique.alphanumeric(number: 30)` as an extra precaution.

BryanEliDimas: You know what? I changed my mind. I am not really submitting the info to the iframe in the tests (I just simulate it with `simulate_step_3_instant_verification`... so I am going to use the NAME generator. Good suggestion :).

PR #877, opened 2020-07-30
Daily student accounts update to Domo

edwoodward:


app/routines/daily_student_count_to_domo.rb
```diff
@@ -0,0 +1,64 @@
...
+    json = JSON.parse(response.body)
+    json['access_token']
+  end
+
+  def info(message)
+    Rails.logger.info("DailyStudentCountToDomo: " + message)
+  end
+
+  def warn(message)
+    Rails.logger.warn("DailyStudentCountToDomo: " + message)
```
BryanEliDimas, line 62: You could also send it to Sentry. Would be very useful imo.

edwoodward: @BryanEliDimas Sentry added to capture the exception

PR #883, opened 2020-08-07
Fetch salesforce abbreviation for book titles from CMS

BryanEliDimas:
The values for the books that users select in Step 4 of the educator sign up flow is not what Salesforce expects. Book titles are different in Salesforce than from those we present to users. This PR fetch those Salesforce book titles from the CMS and posts that as the values for all books. Sometimes both values/book titles are the same but other times they're not. That's why we were seeing the problem of books not submitting only some times. This PR fixes that.


lib/fetch_book_data.rb
```diff
@@ -3,28 +3,14 @@
...
 
-  def subjects
-    @subjects ||= Rails.cache.fetch('BookData.subjects', expires_in: CACHE_DURATION) do
-      fetch_subjects
-    end
-  end
-
   def titles
-    @titles ||= Rails.cache.fetch('BookData.titles', expires_in: CACHE_DURATION) do
+    # @titles ||= Rails.cache.fetch('BookData.titles', expires_in: CACHE_DURATION) do
```
mwharrison, line 11: why is this commented out?

spec/lib/fetch_book_data_spec.rb
```diff
@@ -2,118 +2,12 @@
...
...", [["U.S. History", "US History"]]], ["Business", [["Introduction to Business", "Introduction to Business"]]], ["Business", [["Business Ethics", "Business Ethics"]]], ["Business", [["Principles of Accounting, Volume 1: Financial Accounting", "Financial Accounting"]]], ["Business", [["Principles of Accounting, Volume 2: Managerial Accounting", "Managerial Accounting"]]], ["Business", [["Principles of Management", "Management"]]], ["Business", [["Entrepreneurship", "Entrepreneurship"]]], ["Business", [["Organizational Behavior", "Organizational Behavior"]]], ["Business", [["Business Law I Essentials", "Business Law I Essentials"]]], ["Essentials", [["Business Law I Essentials", "Business Law I Essentials"]]], ["College Success", [["College Success", "College Success"]]], ["Math", [["Prealgebra 2e", "Prealgebra"]]], ["Math", [["Elementary Algebra 2e", "Elementary Algebra"]]], ["Math", [["Statistics(tc)", "HS Statistics"]]], ["High School", [["Statistics(tc)", "HS Statistics"]]], ["High School", [["Physics", "HS Physics"]]], ["Science", [["Physics", "HS Physics"]]], ["Social Sciences", [["Psychology 2e", "Psychology"]]], ["Math", [["Intermediate Algebra 2e", "Intermediate Algebra"]]]]
```
mwharrison, line 10: this should be interesting when we add new books 🤔

lib/fetch_book_data.rb
BryanEliDimas: Oh, great catch! Gotta undo that.

mwharrison:
@BryanEliDimas where is the code that sends this to SF? Is that also updated?

BryanEliDimas:
@mwharrison The result from this API fetch becomes part of the object which is used to build the grouped options of the `select` tag. See https://github.com/openstax/accounts/blob/master/app/controllers/newflow/educator_signup_controller.rb#L130 and https://github.com/openstax/accounts/blob/master/app/views/newflow/educator_signup/educator_profile_form.html.erb#L211

lib/fetch_book_data.rb
BryanEliDimas: Fixed

PR #887, opened 2020-08-12
Handle the case when SheerID doesn't post userInfo as part of a webhook

BryanEliDimas:
SheerID seems to not always post the `userInfo` as part of a webhook request.


app/handlers/newflow/educator_signup/sheerid_webhook.rb
```diff
@@ -16,7 +16,7 @@ def authorized?
       def handle
         verification_id = params.fetch(VERIFICATION_ID_PARAM_NAME, nil) || params.fetch(REQUEST_ID_PARAM)
         outputs.verification_id = verification_id
-        run(ProcessSheeridWebhookRequest, verification_id: verification_id)
+        ProcessSheeridWebhookRequest.perform_later(verification_id: verification_id)
```
mwharrison, line 19: does perform later send this to a background job?

BryanEliDimas: @mwharrison Yes.

BryanEliDimas: One advantage is that jobs retry if/when they fail

PR #888, opened 2020-08-12
Improved error handling for Domo update

edwoodward:


app/controllers/admin/users_controller.rb
```diff
@@ -63,8 +63,12 @@ def mark_users_updated
 
     def update_domo
       dscd = DailyStudentCountToDomo.new
-      dscd.call
-      flash[:notice] = "The update completed."
+      successful = dscd.call
```
BryanEliDimas, line 66: I didn't notice this in my previous code review, but this naming style is non-ruby-esque by the way. I'd recommend something like send_daily_count.

edwoodward: @BryanEliDimas Are you talking about the function name or the variable name?

BryanEliDimas: Variable

BryanEliDimas: No big deal

edwoodward: @BryanEliDimas Variable updated

PR #895, opened 2020-08-21
Validate all fields for CS verification form as in step 4 form

BryanEliDimas:
Initially, from looking at the UX designs, I thought that the CS Verification form would only include 3 fields, but upon Dev Verification later I found out that it was supposed to include all of the fields that the step 4 includes plus school and school-issued email. So I added them to the frontend but didn't remember to add this validations and updates to the backend. Without this PR, a 500 error gets raised in QA.


BryanEliDimas:
Part of 1320 and v7.2.1 ... I want to merge it onto master. Although, do note that there's another PR coming as part of issue 1320 which will solve the reported bugs. Thus, btw, 1320 will end up with 3 PRs attached.

app/controllers/newflow/educator_signup_controller.rb
```diff
@@ -161,7 +161,7 @@ def educator_cs_verification_form
 
     def educator_pending_cs_verification
       security_log(:user_viewed_signup_form, form_name: action_name)
-      @email_address = current_user.email_addresses.first&.value
+      @email_address = current_user.email_addresses.last&.value
```
mwharrison, line 164: what does this accomplish vs `first`?

mwharrison (approved):

app/controllers/newflow/educator_signup_controller.rb
BryanEliDimas: Previously, users during signup could only ever have one email address, so the `first` or `last` email address was the same. However, now, users may have an email address that they signed up with (which could have come from Gsuite or facebook, if not typed in) and then a "school-provided" email address which they want us to use to verify their identity as educators. Therefore, I want to use *that* email address — which is the **last** added to their account.

PR #896, opened 2020-08-26
accounts deployment fixes

scb6:


config/secrets.yml.example
```diff
@@ -208,7 +208,7 @@ production:
   secret_key_base: random(hex,80)
 
   # Used to create urls in emails
-  email_host: "{{ domain }}"
```
mwharrison, line 211: @scb6 I assume this is using the `domain` from the sso options. Where is the new `hosted_zone_name` coming from?

scb6: it is coming from the aws-ruby gem. Domain is also built from aws-ruby gem

PR #897, opened 2020-09-09
Fix 1320

BryanEliDimas:
First issue: after requesting manual CS Verification, users were still able to view and submit the SheerID form (step 3) and step 4. Instead, they should be sent to the screen that says they are pending CS verification.

Second issue seems related to Salesforce, reported by QA in staging. This, I'll be testing manually in the server. My guess, as it has become common recently in the past, is that there's an incosistency between what Accounts expects in Salesforce and what is actually there is the corresponding instance/server.

> so the first thing i noticed is that i never got access. i looked at my lead in sf and it didnt have a subject, as well as the student number was incorrect. when i selected other on the form, the role came back as other in sf, not what was input. even after changing it manually in sf, i never got the access to resources.

### DIFF
--- a/app/controllers/newflow/educator_signup_controller.rb
+++ b/app/controllers/newflow/educator_signup_controller.rb
@@ -26,7 +26,12 @@ module Newflow
     )
     before_action(:store_if_sheerid_is_unviable_for_user, only: :educator_profile_form)
     before_action(:store_sheerid_verification_for_user, only: :educator_profile_form)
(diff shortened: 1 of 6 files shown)


BryanEliDimas:
I just need to get Travis to pass but otherwise, this PR is ready for code review.

BryanEliDimas:
May I please get a code review on this? @mwharrison

mwharrison (approved):

PR #899, opened 2020-09-11
Fix: Faculty Lead missing for users who sign up with social and/or as a student, then request faculty access

BryanEliDimas:
### The fix is to upsert salesforce leads instead of trying to update one when we don't have a lead id for them stored.

The changes in this PR are the same changes introduced in PR Upsert salesforce leads #894. I made a mistake when merging that PR (I merged it on branch v7.0.3 because it was intended to be a patch to that version) instead of master. So this PR is just to add the changes to master.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -124,7 +124,7 @@ gem 'will_paginate'
 gem 'chronic'
 
 # Salesforce
-gem 'openstax_salesforce', github: 'openstax/openstax_salesforce', ref: '58ee73640cbd33c1a8d00277566a2d30074d3657'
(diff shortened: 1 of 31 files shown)


BryanEliDimas:
I just need to get Travis to pass but otherwise, this PR is ready for code review.

mwharrison (approved):
That's a lot of test updating compared to the code changes! 😄

BryanEliDimas:
Exactly 😅

BryanEliDimas:
Just rebased because there was a conflict in a file. Select4d the changes from master and force-pushed.

PR #917, opened 2020-10-13
Replace use of baby_squeel gem with Arel queries

BryanEliDimas:
https://github.com/openstax/business-intel/issues/1529


app/routines/admin/search_security_log.rb
```diff
@@ -64,26 +64,32 @@ def self.sanitize_times(time_strings, now)
 
     def exec(params = {}, options = {})
 
+      sec = SecurityLog.arel_table
+      usr = User.arel_table
+      ap = Doorkeeper::Application.arel_table
+
```
edwoodward, line 70: I'm not a fan of really short variable names unless they are used nearby. Not a showstopper, but it is always clearer to be more descriptive. Just a suggestion.

app/routines/update_user_salesforce_info.rb
```diff
@@ -33,6 +33,8 @@ def self.call(allow_error_email: false)
 
   def call
     info("Starting")
+    ci_table = ContactInfo.arel_table
+    t = User.arel_table
```
edwoodward, line 38: Another super short variable name that could be more descriptive.

edwoodward (approved):

PR #920, opened 2020-10-26
Update will_paginate to 3.3.0

BryanEliDimas:
Attempt to fix https://sentry.cnx.org/openstax/accounts/issues/778390/ which works locally but not in the dev server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,6 @@ services:
 before_script:
   - which google-chrome
   - google-chrome --version
-  - bundle exec rake webdrivers:chromedriver:version
-  - bundle exec rake webdrivers:chromedriver:update
-  - google-chrome --version
   - bundle exec rake assets:precompile
   - bundle exec rake --trace parallel:create parallel:load_schema parallel:seed
 script: bundle exec parallel_rspec ./spec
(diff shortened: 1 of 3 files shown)


edwoodward (approved):

mwharrison:
Had to force merge this because a Travis test was failing repeatedly on spec student_signup_flow_spec.
Creating a card to refactor these tests so they are more reliable.

BryanEliDimas:
Actually, had to force merge because Travis would not run except for [this one time](https://travis-ci.org/github/openstax/accounts/builds/738979125#L1535) where it failed on student_signup_flow_spec. When I ran it locally though, it passed every time. So that tells me that it's one of those intermittent tests which we've been dealing with for a while now.

After investigating why Travis would not run all of a sudden, the best documentation I found is [this](https://docs.travis-ci.com/user/common-build-problems/#my-tests-broke-but-were-working-yesterday). 

> A very common cause when a test is suddenly breaking without any major code changes involved is a change in upstream dependencies.
> 
> This can be an Ubuntu package or any of your project’s language dependencies, like RubyGems, NPM packages, Pip, Composer, etc.

And one more thing that we can try in the future if this happens. [link](https://docs.travis-ci.com/user/common-build-problems/#segmentation-faults-from-the-language-interpreter-ruby-python-php-nodejs-etc):

> Fix the problem by
> 
> clearing the cache or
> removing the cache key from your .travis.yml (you can add it back in a subsequent commit).

PR #922, opened 2020-10-28
Allow unverified users to log in

BryanEliDimas:
https://github.com/openstax/business-intel/issues/1497

New feature: users can log in after completing step 1 (before having to verify their email address).

Also updates the codecov gem because that seems to have been the problem with Travis not building the last time: https://travis-ci.org/github/openstax/accounts/builds/739127303


mwharrison:
Screen Shot 2020-10-28 at 12 46 50 PM

Force merging due to issues with Travis. Tests ran fine locally with no errors.

lib/user_session_management.rb
```diff
@@ -174,8 +174,8 @@ def get_alternate_signup_url
 
   # New flow below
 
-    def save_unverified_user(user)
-      session[:unverified_user_id] = user.id
+    def save_unverified_user(user_id)
+      session[:unverified_user_id] = user_id
```
mwharrison, line 178: why the change from passing the user object around to the id?

mwharrison (approved):

lib/user_session_management.rb
BryanEliDimas: It's an improvement. Passing the whole (giant and super important) User object is unnecessary and dangerous because it could be modified by the function.

mwharrison: cool, just curious.. you should be able to merge now

BryanEliDimas:
Travis finally ran! 🎉 😅 Merging.

PR #924, opened 2020-11-03
Track device ID

jpslav:
Track a device ID for anonymous users so that we can save some temporary state for them.

* Sets a device ID cookie ("oxdid")
* The cookie is not encrypted because if a user messes with it, that is like us giving them a new one (we expect they may change over time).
* The cookie is accessible to Javascript (not sensitive)

Also fixed some other things:
* Added the `listen` gem so that changes in the source are reflected without restarting the rails server (in development).
* Fixed up the docker compose file and added Docker instructions to the readme.  
* Added a way to choose to which kind of debugging gems to install.  byebug is not compatible with the VS Code options, and trying to use it when both options are installed leads to segmentation faults.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,5 @@
 FROM ruby:2.6.3
 
-# Options: vscode, byebug
-ARG DEBUGGER=vscode
-
 WORKDIR /code
 
 COPY Gemfile Gemfile.lock ./
@@ -12,8 +9,7 @@ ENV BUNDLE_PATH="/bundle_cache"\
         BUNDLE_APP_CONFIG="/bundle_cache"\
         GEM_HOME="/bundle_cache"\
         PATH=/bundle_cache/bin:/bundle_cache/gems/bin:$PATH\
-        PORT=2999\
-        DEBUGGER=${DEBUGGER}
+        PORT=2999
 
 RUN gem install bundler:2.1.4 && \
     bundle config set path '/bundle_cache' && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM ruby:2.6.3
 
+# Options: vscode, byebug
+ARG DEBUGGER=vscode
+
 WORKDIR /code
 
 COPY Gemfile Gemfile.lock ./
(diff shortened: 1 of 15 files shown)


Dockerfile
```diff
@@ -9,10 +12,13 @@ ENV BUNDLE_PATH="/bundle_cache"\
         BUNDLE_APP_CONFIG="/bundle_cache"\
         GEM_HOME="/bundle_cache"\
         PATH=/bundle_cache/bin:/bundle_cache/gems/bin:$PATH\
-        PORT=2999
+        PORT=2999\
+        DEBUGGER=${DEBUGGER}
 
-RUN gem install bundler:2.0.2 \
-        && bundle _2.0.2_ install --without production --path /bundle_cache
```
jpslav, line 15: `--without` and `--path` are deprecated

Dockerfile
```diff
@@ -9,10 +12,13 @@ ENV BUNDLE_PATH="/bundle_cache"\
         BUNDLE_APP_CONFIG="/bundle_cache"\
         GEM_HOME="/bundle_cache"\
         PATH=/bundle_cache/bin:/bundle_cache/gems/bin:$PATH\
-        PORT=2999
+        PORT=2999\
+        DEBUGGER=${DEBUGGER}
```
jpslav, line 16: This environment variable used in the Gemfile to determine which debugger gems to install.

TomWoodward:
do we not have some central place where we define generic data that is added to every event such as user uuid and device id? 

currently this doesn't look like its used anywhere, but we have events defined already right? at what point will this be integrated into something?

config/initializers/controllers.rb
```diff
@@ -69,4 +72,21 @@ def disable_fine_print
   def set_locale
     I18n.locale = http_accept_language.compatible_language_from(I18n.available_locales)
   end
+
+  def set_device_id
+    cookies.delete(:oxdid) if cookies[:oxdid] && !(cookies[:oxdid] =~ UUID_REGEX)
```
steveburkett, line 77: non-trivial if expressions are often better written as multiline.  this is close, but, imo, it'd be a bit easier to grok if multi line. 

https://github.com/airbnb/ruby#only-simple-if-unless

steveburkett: would be nice if this piece were encapsulated in a class w/ a SRP on its own (and unit tested in isolation from the controller)

steveburkett: and then a simpler golden path test for the controller

jpslav: made the `if` better via a new helper method

mwharrison:
Ran tests locally and they are passing - good to merge. @jpslav

mwharrison:
Ran tests locally, in lieu of Travis due to issues we're looking into. Good to merge.

PR #925, opened 2020-11-09
LLPH/BRI book flow

BryanEliDimas:
https://github.com/openstax/business-intel/issues/1589


app/helpers/newflow/login_signup_helper.rb
```diff
@@ -1,11 +1,29 @@
...
 
-    # save (in the seession) or clear the client_app that sent the user here
+    BRI_BOOK_PARAM_NAME = :bri_book
+
+    # save (in the session) or clear the client_app that sent the user here
     def cache_client_app
       set_client_app(params[:client_id])
     end
 
+    def cache_BRI_marketing_if_present
```
mwharrison, line 11: Why is this the only Salesforce field name that has all this cache clearing settings?

app/controllers/newflow/student_signup_controller.rb
```diff
@@ -16,9 +16,11 @@ def student_signup
         contracts_required: !contracts_not_required,
         client_app: get_client_app,
         user_from_signed_params: session[:user_from_signed_params],
+        is_BRI_book: is_BRI_book_adopter?,
         success: lambda {
           save_unverified_user(@handler_result.outputs.user.id)
-          security_log :student_signed_up, { user: @handler_result.outputs.user }
+          security_log(:student_signed_up, { user: @handler_result.outputs.user })
+          clear_cache_BRI_marketing
```
edwoodward, line 23: On student and educator flow, clear_cache_BRI_marketing is called on success. On social_outh flow, it is called on success and failure. Is this intentional?

app/views/newflow/student_signup/student_signup_form.html.erb
```diff
@@ -134,6 +134,17 @@
                 <% end  %>
 
                 <div class="content control-group checkboxes-section">
+
+                    <% if is_BRI_book_adopter? %>
```
mwharrison, line 138: Students don't need to see the Title I question - only educators go through this flow.
Sorry, this wasn't clear on the card.

app/helpers/newflow/login_signup_helper.rb
BryanEliDimas: Think of it as a [pre_auth_state](https://github.com/openstax/accounts/blob/master/app/controllers/legacy/signup_controller.rb#L27-L28). [Here are the methods](https://github.com/openstax/accounts/blob/a65b6c0c5cb73a0b7e86cc02948c935b4c7408ed/lib/user_session_management.rb#L107-L117) for saving and clearing the pre_auth_state cache.

app/controllers/newflow/student_signup_controller.rb
BryanEliDimas: I don't believe it is called on failure.

app/views/newflow/student_signup/student_signup_form.html.erb
BryanEliDimas: Oh! 😣 I'm on it.

app/helpers/newflow/login_signup_helper.rb
mwharrison: I thought there were no more preauth states?

Regardless, I'm still unclear why only this field needs to be specifically cache cleared vs other fields on the Salesforce model (which I guess you're saying is taken care of in the pre_auth_state?). For example - "Other" role name doesn't seem to do this: https://github.com/openstax/accounts/pull/923

Just trying to understand the difference.

app/handlers/newflow/student_signup/signup_form.rb
```diff
@@ -61,11 +64,23 @@ def handle
         end
 
         agree_to_terms
+
+        if options[:is_BRI_book]
```
edwoodward, line 68: This same block is in the educator signup flow. Can it be in one place and not duplicated?

app/routines/newflow/educator_signup/activate_educator.rb
```diff
@@ -4,6 +4,9 @@
 module Newflow
   module EducatorSignup
     class ActivateEducator
+
+      SECURITY_LOG_EVENT_TYPE = :user_became_activated
```
edwoodward, line 8: Can this constant be in one place? It is in the educator and student signup flows.

edwoodward: There is very little code that is different in this class and the Activate Student class. Can there be a base class with the shared code?

app/views/newflow/educator_signup/educator_signup_form.html.erb
```diff
@@ -121,6 +121,17 @@
         </div>
 
         <div class="content control-group checkboxes-section">
+
+          <% if is_BRI_book_adopter? %>
```
edwoodward, line 125: This block is in 3 different forms. Should it be a fragment so it only exists in one place?

app/helpers/newflow/login_signup_helper.rb
BryanEliDimas: There is no need to cache the "Other" role name because that is submitted in a form that the user submits. In this PR, for the BRI marketing/`bri_book` field, there is a need to save in the cache (and then clear), because we get that info from a parameter in the URL and then I need I need to use it as if the user had submitted it in the form.

app/routines/newflow/educator_signup/activate_educator.rb
BryanEliDimas: Fixed.

app/views/newflow/educator_signup/educator_signup_form.html.erb
BryanEliDimas: Great suggestion, thanks! Done.

app/handlers/newflow/oauth_callback.rb
```diff
@@ -68,6 +73,14 @@ def handle # rubocop:disable Metrics/AbcSize
 
     private ###########################
 
+    def BRI_marketing(user)
```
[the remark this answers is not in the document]
BryanEliDimas: Great suggestion, thanks! Done.

BryanEliDimas:
@edwoodward @mwharrison Addressed all suggestions. Included all in one PR.

app/controllers/newflow/student_signup_controller.rb
edwoodward: You're right. I didn't pay attention to the line numbers.

edwoodward (approved):

PR #927, opened 2020-11-11
Change to users representer

edwoodward:
Removed private data restriction from salesforce_contact_id


mwharrison (approved):
@edwoodward I ran the tests locally and they pass - so I am okay merging this.

PR #929, opened 2020-11-12
Dockerized testing

jpslav:


Gemfile
```diff
@@ -173,14 +173,13 @@ group :development, :test do
...
-    gem 'ruby-debug-ide'
-    gem 'debase'
-  end
+  # See config/initializers/04-debugger.rb
+  #
+  # Call 'debugger' anywhere in the code to stop execution and get a debugger console
+  gem 'byebug', require: false
+  # Debug in VS Code
+  gem 'ruby-debug-ide', require: false
+  gem 'debase', require: false
```
jpslav, line 182: Switched to installing the both but requiring (activating) them in the running code depending on the environment variable.

Gemfile
```diff
@@ -241,10 +240,11 @@ group :test do
...
   gem 'db-query-matchers'
 
-# Run feature tests with Capybara + Selenium
+  # Run feature tests with Capybara + Selenium; choose which driver gems to use
+  # based on test environment.
   gem 'capybara'
-  gem 'selenium-webdriver', '>= 3.141.0'
-  gem 'webdrivers', '~> 4.0'
+  gem 'selenium-webdriver', '>= 3.141.0', require: false
+  gem 'webdrivers', '~> 4.0', require: false
```
jpslav, line 247: Same approach here, add both, turn them on selectively in running code

lib/host.rb
```diff
@@ -1,13 +1,17 @@
 module Host
-  TRUSTED_HOST_REGEXES = Rails.application.secrets.trusted_hosts.map do |host|
-    /\A(.*\.)?#{host.gsub('.', '\.')}\z/
+  def self.trusted_hosts
+    Rails.application.secrets.trusted_hosts
```
jpslav, line 3: Making this a method here makes it easier to stub the trusted hosts in specs.

spec/features/admin/users_spec.rb
```diff
@@ -38,6 +38,7 @@
         it 'searches users and does not explode' do
           visit '/'
           click_link 'Popup Console'
+          wait_for_ajax(10) # for some reason this is slow
```
jpslav, line 41: Doesn't mean it will wait 10 seconds, just that it will wait up to 10 seconds

spec/rails_helper.rb
```diff
@@ -48,20 +47,69 @@
...
+  Capybara.asset_host = "http://#{capybara_host}:#{capybara_port}"
+end
 
 Capybara.server = :puma, { Silent: true } # To clean up your test output
 
 # Normalize whitespaces
 Capybara.default_normalize_ws = true
 
+Capybara.configure do |config|
+  config.default_max_wait_time = 15
```
jpslav, line 101: Fixed a number of intermittent "didn't find such and such field" errors; tho note a value of 10 here still had the problem.

spec/rails_helper.rb
```diff
@@ -48,20 +47,69 @@
...
+  Capybara.server_host = capybara_host
+  Capybara.server_port = capybara_port
 else
-  Capybara.javascript_driver = :selenium_chrome
-end
+  require 'webdrivers/chromedriver'
 
-Capybara.asset_host = 'http://localhost:2999'
+  if EnvUtilities.load_boolean(name: 'HEADLESS', default: false)
+    # Run the feature specs in a full browser (note, this takes over your computer's focus)
```
mwharrison, line 83: what's the reason to not run local tests headless too?

spec/rails_helper.rb
mwharrison: what was causing this? the page being checked before it was fully loaded?

README.md
```diff
@@ -112,6 +109,23 @@ and then start the `delayed_job` daemon:
...
+
+```sh
+$> docker-compose run --rm app /bin/bash
+```
+
+To drop into a container with everything installed, then you can
+
+```sh
+$ /code> rspec
+$ /code> rspec ./spec/features
```
mwharrison, line 122: When trying to run tests in docker, I am getting:

```
ActiveRecord::NoDatabaseError:
  FATAL:  database "ox_accounts_test" does not exist
```

is this something that should happen automatically?

mwharrison (approved):

README.md
jpslav: No, I forgot some stuff in readme, adding now

spec/rails_helper.rb
jpslav: to actually see how things render, makes your chrome open and close a million times, but we can use the `screenshot!` to take pictures of every screen or we can use the debugger to stop in one screen and visually inspect it.

spec/rails_helper.rb
jpslav: yeah, the page being slow for who knows what, and Capybara hitting a timeout and being like "I don't see anything!"

PR #932, opened 2020-12-05
Block creation or update of emails that create any duplicates

Dantemss:
Completely blocks adding emails that have been used to an account, even if they have not been verified.

After releasing this, we can probably clean up the data and add a unique index in a follow up release.

### DIFF
--- a/app/models/contact_info.rb
+++ b/app/models/contact_info.rb
@@ -8,9 +8,7 @@ class ContactInfo < ActiveRecord::Base
   validates :type, presence: true
   validates :value,
             presence: true,
-            uniqueness: {scope: [:user_id, :type], case_sensitive: false}
-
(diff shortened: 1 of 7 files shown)


mwharrison (changes requested):
@Dantemss there are some failing tests around emails. Can you take a look?

See here for more deets: https://openstax.slack.com/archives/C6NRD0CCQ/p1607350015141700?thread_ts=1607131336.139800&cid=C6NRD0CCQ

mwharrison:
@Dantemss I am out next week and we're aiming to get this in for the 12/18 code freeze, so just wanted to make sure this was on your radar.

mwharrison (approved):
Looks great - thank you @Dantemss and @nathanstitt!

PR #934, opened 2020-12-09
Rake task to remove dup emails

steveburkett:
Remove duplicate emails

https://app.zenhub.com/workspaces/openstax-bit-5bacfe3f4b5806bc2bea3764/issues/openstax/business-intel/1641

### DIFF
--- a/lib/remove_duplicate_emails.rb
+++ b/lib/remove_duplicate_emails.rb
@@ -1,8 +1,8 @@
 class RemoveDuplicateEmails
-  def run(do_it: false, older_than: '1/10/2020' )
+  def run(do_it: false, older_than: '2020-10-01' )
     # find all duplicate contact infos, older than X
     all_dup_contact_infos = ContactInfo.
-      where("created_at < ?", Date.parse(older_than)).
+      where("created_at < ?", Date.strptime(older_than, "%Y-%m-%d")).
       select(:type, :value).group(:type, :value).
       having('count(*) > 1')
     users_to_remove = [], contactinfos_to_remove = []
--- a/lib/remove_duplicate_emails.rb
+++ b/lib/remove_duplicate_emails.rb
@@ -1,0 +1,39 @@
+class RemoveDuplicateEmails
+  def run(do_it: false, older_than: '1/10/2020' )
+    # find all duplicate contact infos, older than X
+    all_dup_contact_infos = ContactInfo.
+      where("created_at < ?", Date.parse(older_than)).
+      select(:type, :value).group(:type, :value).
+      having('count(*) > 1')
+    users_to_remove = [], contactinfos_to_remove = []
+
+    all_dup_contact_infos.each do |dup_contact_info|
+      # of the duplicate contact infos value ... find the unverified duplicates
+      unverified_contactinfos = ContactInfo.where(type: dup_contact_info.type, value: dup_contact_info.value, verified: false)
+
+      unverified_contactinfos.each do |unverified_contactinfo|
+        if unverified_contactinfo.user.contact_infos.count == 1 && unverified_contactinfo.user.contact_infos.first.value == unverified_contactinfo.value
+          # this is a dup user with just this one unverified email
+          users_to_remove.append(unverified_contactinfo.user.id)
+        else
+          # this is a dup contact info that's unverified w/ no 1:1 user
+          contactinfos_to_remove.append(unverified_contactinfo.id)
+        end
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


lib/remove_duplicate_emails.rb
```diff
@@ -0,0 +1,36 @@
+class RemoveDuplicateEmails
+  def run(do_it: false)
+    # find all duplicate contact infos, older than X
+    all_dup_contact_infos = ContactInfo.where("created_at < ?", Date.parse("1/10/2020")).select(:type, :value).group(:type, :value).having('count(*) > 1')
```
Dantemss, line 4: Maybe put the date as a parameter with default '1/10/2020' so we can run it again later without having to make another release.

lib/remove_duplicate_emails.rb
```diff
@@ -0,0 +1,36 @@
...
+    users_to_remove = [], contactinfos_to_remove = []
+
+    all_dup_contact_infos.each do |dup_contact_info|
+      # of the duplicate contact infos value ... find the unverified duplicates
+      unverified_contactinfos = ContactInfo.where(value: dup_contact_info.value, verified: false)
```
Dantemss, line 9: Need to match the type too

lib/remove_duplicate_emails.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      unverified_contactinfos.each do |unverified_contactinfo|
+        if unverified_contactinfo.user.contact_infos.count == 1 && unverified_contactinfo.user.contact_infos.first.value == unverified_contactinfo.value
+          # this is a dup user with just this one unverified email
+          users_to_remove.append(unverified_contactinfo.user.id)
```
Dantemss, line 14: It seems to me the contact_infos table is missing a foreign key constraint to auto-delete when you delete the user, so you have to add it to contactinfos_to_remove here too.

lib/remove_duplicate_emails.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    users_to_remove = users_to_remove.flatten.uniq.sort
+    contactinfos_to_remove = contactinfos_to_remove.flatten.uniq.sort
+
+    if do_it
+      ActiveRecord::Base.transaction do
+        ContactInfo.where(id: contactinfos_to_remove).destroy_all
+        User.where(id: users_to_remove).destroy_all
+      end
```
Dantemss, line 29: Since the records were loaded outside of the transaction there's a race condition here where they get modified after being loaded... but uh I guess it's fine as long as the `created_at` are far in the past so you can leave it.

Dantemss (changes requested):
Looks good overall, only first 3 comments are important.

lib/remove_duplicate_emails.rb
steveburkett: @Dantemss isnt this that? 

```
  has_many :contact_infos, dependent: :destroy, inverse_of: :user
```

Dantemss: Ah you're right. I thought you had used `delete_all` which wouldn't use the dependent, but it's in fact `destroy_all` which does.

steveburkett:
@Dantemss is this one +1?  i do want to work w/ @mwharrison to test this thoroughly on dev first before merging

lib/remove_duplicate_emails.rb
```diff
@@ -0,0 +1,39 @@
+class RemoveDuplicateEmails
+  def run(do_it: false, older_than: '1/10/2020' )
+    # find all duplicate contact infos, older than X
+    all_dup_contact_infos = ContactInfo.
+      where("created_at < ?", Date.parse(older_than)).
```
mwharrison, line 5: Might consider using `Date.strptime(older_than, "%d/%m/%Y")` to make it a bit more specific.

Dantemss: I would say make it YYYY-mm-dd so there's absolutely no chance for confusion.

mwharrison: That works - I ask because I was initially confused 😄

Dantemss:
@steveburkett Michael found an important issue (date format). Other than that, it looks OK to me.

lib/remove_duplicate_emails.rb
steveburkett: @mwharrison only in USA do we get confused on stuff like dates and measuring distance and weight :)))

Dantemss (dismissed):

Dantemss (approved):

PR #943, opened 2021-03-12
Handle schools that have been deleted from Salesforce

Dantemss:
We check only schools with no users so we wait until users have been updated and moved to another school and don't leave the users without a school in-between.

### DIFF
--- a/app/routines/update_school_salesforce_info.rb
+++ b/app/routines/update_school_salesforce_info.rb
@@ -24,8 +24,22 @@ class UpdateSchoolSalesforceInfo
   def call
     log 'Starting'
 
-    # Go through all SF Schools and cache their information, if it changed
+    # Check if any Schools that have 0 users have been deleted from Salesforce and remove them
+    School.where(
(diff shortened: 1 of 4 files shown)


mwharrison (dismissed):
Is there something that updates the school on a contact if it's changed in Salesforce?
I think that happens during the Salesforce sync, right?

Dantemss:
Yeah the schools are updated here: https://github.com/openstax/accounts/blob/master/app/routines/update_user_salesforce_info.rb#L307

Based on the school found here: https://github.com/openstax/accounts/blob/master/app/routines/update_user_salesforce_info.rb#L51

So if the school id in the Contact gets updated the user will be moved eventually.

Dantemss:
Actually it might be a good idea to not clear the school when the warning triggers, since that could happen if a new school was created in Salesforce and people immediately moved over, before Accounts could cache the new school. I'll make that small change.

mwharrison (approved):
Good catch, approved

PR #947, opened 2021-03-25
Updated mimemagic gem and fixed branch name on GH actions

Dantemss:
The creator or maintainer of mimemagic yanked all old versions from rubygems.org.
This PR is necessary or else deployments will fail.


Dantemss:
The migration flow will keep failing because it tries to install the gem that no longer exists. There are no migrations in this PR anyway.

edwoodward (approved):

mwharrison (approved):

PR #965, opened 2021-05-24
Sending exceptions from user Salesforce update to Sentry

edwoodward:
The timeout issue is most likely caused by the web server timing out. There are other issues in Sentry related to UpdateUserSalesforceInfo that are causing the process to stop on an error. The issues are [here](https://sentry.cnx.org/openstax/accounts/issues/1074738/events/4256673/) and [here](https://sentry.cnx.org/openstax/accounts/issues/1074736/events/4257355/). Wrapping the code in rescue will allow the process to continue and we might get more info in Sentry than from a crash. We also have the option of adding more info to what is being sent to Sentry if we need it.

Errors are collected during the process and sent to devs in an email. Most likely, the email is not being sent when these errors occur. This change will allow the email to be sent and hopefully capture any other errors we are missing.

### DIFF
--- a/app/routines/update_user_salesforce_info.rb
+++ b/app/routines/update_user_salesforce_info.rb
@@ -54,6 +54,7 @@ class UpdateUserSalesforceInfo
         cache_contact_and_school_data_in_user!(contact, school, user)
       rescue StandardError => ee
         error!(exception: ee, user: user)
+        Sentry.capture_exception ee
       end
     end
 
@@ -89,6 +90,7 @@ class UpdateUserSalesforceInfo
           end
         rescue StandardError => ee
           error!(exception: ee, user: user)
+          Sentry.capture_exception ee
         end
       end
     end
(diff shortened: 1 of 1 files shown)


app/routines/update_user_salesforce_info.rb
```diff
@@ -54,6 +54,7 @@ def call
         cache_contact_and_school_data_in_user!(contact, school, user)
       rescue StandardError => ee
         error!(exception: ee, user: user)
+        Sentry.capture_exception ee
```
mwharrison, line 57: It looks like you are adding this everywhere that `error` is called. You could probably just call this one time in this function: https://github.com/openstax/accounts/blob/prevent-salesforce-errors/app/routines/update_user_salesforce_info.rb#L324-L336

mwharrison (changes requested):

app/routines/update_user_salesforce_info.rb
edwoodward: @mwharrison I was selective about where I added it because not all of the errors collected are exceptions and should not be sent to Sentry

mwharrison: Ok - approving

mwharrison (approved):

PR #968, opened 2021-06-17
Added email constraint to database

edwoodward:

PR #975, opened 2021-07-21
Revert "Added email constraint to database"

mwharrison:
Reverts openstax/accounts#968

Issues deploying this, complaining about duplicate email addresses. Reverting to deploy QA.


mwharrison:
@edwoodward this is failing deployment on QA, so I think we should revert it. Certainly don't want issues on production.

I tried running the rake task to clean up duplicate accounts and it didn't help. Let me know if you think it'd be best to move forward another way.

edwoodward (approved):

PR #977, opened 2021-08-02
integrate accounts with salesforce streaming api for quick contact up…

mwharrison:
…dates


app/services/salesforce_subscriber.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def subscribe
+    authorization_hash = @client.authenticate!
+    @client.faye.set_header 'Authorization', "OAuth #{authorization_hash.access_token}"
+    EM.run do
+      @client.subscription "/topic/#{CONTACT_PUSH_TOPIC_NAME}", replay: -1 do |message|
+        Logger.new(LOG_PATH).debug('Contact Received')
+        ContactParser.new(message).save_contact
+      end
+    end
```
jpslav, line 46: maybe we can move some of this into `openstax_salesforce` and pass in application-specific stuff in a block... 🤔

app/services/salesforce_subscriber.rb
```diff
@@ -0,0 +1,59 @@
...
+                            password: ENV['SALESFORCE_PASSWORD'],
+                            security_token: ENV['SALESFORCE_SECURITY_TOKEN'],
+                            client_id: ENV['SALESFORCE_CONSUMER_KEY'],
+                            client_secret: ENV['SALESFORCE_CONSUMER_SECRET'],
+                            host: ENV['SALESFORCE_LOGIN_DOMAIN'])
+  end
+
+  def create_contact_push_topic
+    # Create a PushTopic for subscribing to Contact changes.
+    delete_contact_topics
```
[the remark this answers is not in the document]
mwharrison: I think this is the safest way to do this - a topic doesn't need to be reusable but you need to make sure it exists before trying to use it.

app/services/salesforce_subscriber.rb
mwharrison: I assume you mean this part?

```
authorization_hash = @client.authenticate!
@client.faye.set_header 'Authorization', "OAuth #{authorization_hash.access_token}"
```

I'll look into it, `openstax_salesforce` doesn't use faye right now.

Gemfile
```diff
@@ -164,6 +164,11 @@ gem 'oj_mimic_json'
 # CORS for local testing/dev
 gem 'rack-cors'
 
+# Salesforce streaming API
+gem 'cookiejar', git: 'https://github.com/MissionCapital/cookiejar.git'
+gem 'faye', '0.8.9'
+gem 'restforce', '~> 5.0.6'
```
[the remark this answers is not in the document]
mwharrison: removed

Gemfile
```diff
@@ -164,6 +164,10 @@ gem 'oj_mimic_json'
 # CORS for local testing/dev
 gem 'rack-cors'
 
+# Salesforce streaming API
+gem 'cookiejar', git: 'https://github.com/MissionCapital/cookiejar.git'
+gem 'faye', '0.8.9'
```
edwoodward, line 169: The version of cookiejar (latest 0.3.3 2016) and faye (latest 1.4.0 2020) are not the latest. Cookiejar is a fork of the original that is behind in commits from the original. Why these versions?

mwharrison: Good catch!
This is the blog post I read to get started with this (written in 2020). https://medium.com/@sahsantoshh/salesforce-integration-with-ruby-on-rails-for-real-time-changes-ff93dde05ba4

I tried to update faye to the latest version but the stream connection kept dropping every 10-20 minutes. There was a github issue that I can't find now that mentioned having to use faye 0.8.9 with the salesforce streaming API. 
So, I don't know - but I do know this version is required for it to work because the newer one didn't.

I'll checkout the cookiejar, which might be okay to use from the original package.

edwoodward (approved):

edwoodward:
@mwharrison I've approved this, but some documentation in the wiki would be nice when you have time.

Gemfile
mwharrison: Actually, the version on rubygems of cookiejar (0.3.3 from 2016) is nearly a year older than the version being used here https://github.com/MissionCapital/cookiejar/commit/68eecc8415338c6e9d636c4f7e4d2e13547e20a2 - from 2017).

edwoodward: Sorry I missed that.

mwharrison:
> @mwharrison I've approved this, but some documentation in the wiki would be nice when you have time.

Added a page in the wiki for this: https://github.com/openstax/accounts/wiki/Streaming-API
I'll add more once the daemon service is setup, I think that'll be the part that has the most magic in it. :)

PR #987, opened 2021-09-07
Changes for lead refactoring/Slalom refactoring

mwharrison:


app/services/salesforce_subscriber.rb
```diff
@@ -13,27 +13,29 @@ def initialize
...
+      else
+        Rails.logger.error('failed to create push topic: ' + CONTACT_PUSH_TOPIC_NAME)
+        Sentry.capture_message('failed to create push topic: ' + CONTACT_PUSH_TOPIC_NAME)
+        raise
+      end
     end
   rescue Restforce::ErrorCode::DuplicateValue
-    delete_contact_topics
-    create_contact_push_topic
+    Rails.logger.debug('Push topic duplicate found.')
```
Dantemss, line 38: Might wanna call `retry` here (I think) in case the topic got created in between the 2 calls. And it's a good idea to have a max_retry...

app/routines/update_user_salesforce_info.rb
```diff
@@ -258,9 +258,7 @@ def cache_contact_and_school_data_in_user!(contact, school, user)
                               contact.faculty_verified}'' on contact #{contact.id}"
                             end
 
-      # TODO: We can read school_type and school_location from the cached School records instead,
-      # but better wait 1 additional release to let the Schools be cached and linked
-      user.school_type = case contact.school_type
+      user.school_type = case school&.type
```
Dantemss, line 261: Not sure if it's a problem, but there could be a case where we are running this but the school is not cached yet? There's even a log warning for that case.

app/services/salesforce_subscriber.rb
mwharrison: Didn't know about the `retry` command - cool! Added.

app/routines/update_user_salesforce_info.rb
mwharrison: I reverted this back because you are correct and there is no real benefit to reading this from the cached school over the contact that I can see.

PR #995, opened 2021-10-07
Look up school_location and school_type in User model

edwoodward:


app/models/user.rb
```diff
@@ -192,6 +192,30 @@ def most_accurate_school_country
     school.present? ? school.country : 'United States'
   end
 
+  def school_location
+    return UserHelper.convert_to_user_location(school.location) if school.present? && school.has_attribute?("location")
```
Dantemss, line 196: `has_attribute?` will return false if, for example, the column doesn't exist or if the school was loaded with `School.select(...)` and the select did not include location, leading to inconsistent behavior. I suggest using something like `!school.location.blank?` which would throw an error instead in those cases.

app/models/user.rb
```diff
@@ -192,6 +192,30 @@ def most_accurate_school_country
...
+
+    lookedup_school = School.where(name: self_reported_school) if self_reported_school.present?
+    return UserHelper.convert_to_user_location(lookedup_school.to_a[0]['location']) if lookedup_school.present?
+
+    super if super.present?
+  end
+
+  def school_type
+    return UserHelper.convert_to_user_type(school.type) if school.present? && school.has_attribute?("type")
```
Dantemss, line 205: Same as above.

spec/models/user_spec.rb
```diff
@@ -63,6 +63,7 @@
     user = FactoryBot.create :user, username: " user "
     expect(user.username).to eq "user"
 
+    school = FactoryBot.create :school, name: "Rice University"
```
Dantemss, line 66: Could also test that the user gets the location and type from the school here.

app/models/user.rb
```diff
@@ -192,6 +192,30 @@ def most_accurate_school_country
     school.present? ? school.country : 'United States'
   end
 
+  def school_location
+    return UserHelper.convert_to_user_location(school.location) if school.present? && school.has_attribute?("location")
+
+    lookedup_school = School.where(name: self_reported_school) if self_reported_school.present?
```
Dantemss, line 198: Maybe lookedup_school could be a memoized method so we can load it only once if we need both the school type and the school location.

app/models/user.rb
edwoodward: @Dantemss I was using school.location.blank?, but a test was failing because it was a school with only an id.

spec/models/user_spec.rb
edwoodward: Fixed

app/models/user.rb
edwoodward: Fixed

app/models/user.rb
edwoodward: Fixed

PR #996, opened 2021-10-11
Update users whose Lead is not Converted

edwoodward:


Gemfile
```diff
@@ -125,7 +125,7 @@ gem 'will_paginate'
 gem 'chronic'
 
 # Salesforce
-gem 'openstax_salesforce', github: 'openstax/openstax_salesforce', ref: 'ba17e6de2ae8e0aaecb140a2a9be0a3df64f91d7'
+gem 'openstax_salesforce', github: 'openstax/openstax_salesforce', ref: '2db69c3ae133e5f6c3aa6edc377add291d144125'
```
mwharrison, line 128: Can you merge [this PR](https://github.com/openstax/openstax_salesforce/pull/51) and use the SHA from master before merging this? (don't forget to `bundle update`)

app/helpers/user_helper.rb
```diff
@@ -0,0 +1,40 @@
+module UserHelper
+  COLLEGE_TYPES = [
+    'College/University (4)',
+    'Technical/Community College (2)',
+    'Career School/For-Profit (2)'
+  ].freeze
+  HIGH_SCHOOL_TYPES = ['High School'].freeze
```
mwharrison, line 7: what is the purpose of these single-item arrays?

Can the below case statements just be:

```
    when *COLLEGE_TYPES
      :college
    when 'High School'
      :high_school
    when 'K-12 School'
      :k12_school
    when 'Home School'
      :home_school
```

or do you expect these types might grow one day?

what I'm suggesting might be really un-ruby like, not sure :)

app/routines/update_user_salesforce_lead_info.rb
```diff
@@ -0,0 +1,24 @@
+class UpdateUserSalesforceLeadInfo
+  def self.call
+    new.call
+  end
+
+  def call
+    leads ||= OpenStax::Salesforce::Remote::Lead.where('status != ?','Converted').select(:id, :grant_tutor_access).to_a
```
mwharrison, line 7: Screen Shot 2021-10-13 at 1 10 47 PM

Just so you are aware, I ran this query on production and it returns over 500k rows. So this will probably be VERY slow and memory-consuming.

PR #1011, opened 2021-11-04
Changes to support new deployment

Dantemss:
- No longer generate secrets.yml from secrets.yml.example
- Committed secrets.yml with ENV vars
- Committed production database.yml with ENV vars
- Committed sentry and scout initializers
- Updated puma
- Added sd_notify, delayed_job heartbeat and worker pool


Dantemss:
The migration test will keep failing because of the ruby version upgrade (some of the gems before the upgrade are not compatible with ruby 2.7 so the pre-migration step fails when it tries to install the old gems in ruby 2.7) so we'll have to ignore that one.

mwharrison (approved):

PR #1016, opened 2021-12-01
[hotfix] revert back to therubyracer 

mwharrison:
`552ca52a51e4fa3810946bd753a7a70ed13295c9` was deployed to production to fix an error with mini_racer

this still might be needed because we didn't see any issues on testing environments with low load. If you see any V8 errors on testing environments, will need to migrate back to therubyracer gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem 'compass-rails'
 gem 'coffee-rails', '5.0.0'
 
 # JavaScript asset compiler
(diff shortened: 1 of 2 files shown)


Dantemss (approved):

Dantemss (approved):

Dantemss:
This can probably be closed, the new deployment PR uses a different version of mini_racer that should work...

PR #1026, opened 2022-01-26
Require student count for each adopted book

RoyEJohnson:
it submits


RoyEJohnson:
Four test are failing.
The first one is that choosing "Other educational staff" for the role when signing up _should_ expose a blank to fill in for what that role is.

mwharrison (approved):

PR #1028, opened 2022-01-27
Cleanup sheerid response handling

mwharrison:
This removes a lot of duplicated code. There is now one profile form - there used to be a separate one for CS verification that had nearly identical code. 
This also simplifies the lead creation and SheerID webhook process.

Still a long ways to go to make this great - but feels like this is a good start.


edwoodward (dismissed):

mwharrison:
Screen Shot 2022-01-28 at 11 15 03 AM
We still might have an issue here if the SheerID webhook returns before the user completes their profile (so, instant verification).

app/handlers/newflow/educator_signup/complete_profile.rb
```diff
@@ -98,45 +121,69 @@ def format_books_for_salesforce_string(books)
...
         end
 
-        if signup_params.educator_specific_role.strip.downcase  == INSTRUCTOR &&
-          signup_params.num_students_per_semester_taught.blank?
-
+        if role  == INSTRUCTOR && signup_params.num_students_per_semester_taught.blank?
           param_error(:num_students_per_semester_taught, :num_students_must_be_entered)
         end
+
+        # if (!@did_use_sheerid || role != OTHER) && !signup_params.school_issued_email.blank?
```
edwoodward, line 157: I assume this commented out code is here for a reason.

spec/features/newflow/educators/educator_signup_flow_spec.rb
```diff
@@ -60,9 +60,9 @@ module Newflow
           select_educator_role('other')
           fill_in('Other (please specify)', with: 'President')
           click_on('Continue')
-          expect(page.current_path).to eq(signup_done_path)
-          click_on('Finish')
-          expect(page.current_url).to eq(external_app_url)
+          #expect(page.current_path).to eq(signup_done_path)
```
edwoodward, line 63: Should this be commented out? There is another section below.

app/handlers/newflow/educator_signup/complete_profile.rb
mwharrison: It is, the email field on the CS version of the form is not being validated and I need to revisit it.

edwoodward (approved):

spec/features/newflow/educators/educator_signup_flow_spec.rb
mwharrison: I think this was leftover from debugging - uncommented

app/handlers/newflow/educator_signup/complete_profile.rb
mwharrison: This has been fixed. Here is the commit: https://github.com/openstax/accounts/pull/1028/commits/88c987a7264458351ac7359d3cc995a95ca02b70

PR #1029, opened 2022-02-02
refactor email code if student inputs same email already on account

mwharrison:
When a student went to convert to an instructor, if they typed the same email they used at signup there was an error - this error was also hidden because it happened before we passed the form they were on to the handler.


edwoodward (approved):

mwharrison:
removed the `is_verification_stale` check because this is now unused and was causing intermittent test failures

PR #1030, opened 2022-02-02
Fix users not uploaded to sf during bug

mwharrison:


mwharrison:
@edwoodward this is ready for review - it does include some skipped tests but I'd like to make another PR for the big change away from using newflow and this code has been tested manually a number of different ways, so I feel confident it is an improvement over what we have now.

I will be getting other apps to test as well though before we release (exp. tutor/rex).

app/controllers/application_controller.rb
```diff
@@ -16,6 +14,7 @@ def disable_fine_print
   end
 
   def check_if_admin
+    return true if !Rails.env.production?
```
edwoodward, line 17: Does this allow anyone to be admin? I see there is a check for Blazer using this, but no where else.

edwoodward (approved):

app/controllers/application_controller.rb
mwharrison: yes but only in non-prod environments (so only locally - since everything else runs as ENV=production)

mwharrison: and yes, it's only used for Blazer auth checking

PR #1033, opened 2022-02-25
refactor newflow to not exist

mwvolo:


RoyEJohnson:
On the login form, the error state seems to be served up with the page. On Dev, it serves without error, then if you try to submit, it serves the version with the error. You can reload the page, and the error state remains. But if you go Back, the no-error page is there.

app/views/login/login_form.html.erb
```diff
@@ -2,18 +2,22 @@
...
-        <%= newflow_login_signup_card(
+        <%= login_signup_card(
                     classes: "login-page",
                     header: (I18n.t(:"login_signup_form.login_page_header")),
                     banners: @banners,
                     show_exit_icon: true) do %>
-            <% lev_form_for :login_form, url: newflow_login_path do |f| %>
-                <% fh = NewflowFormHelper::Newflow.new(f: f, context: self, errors: @handler_result&.errors) %>
+            <% lev_form_for :login_form, url: login_path do |f| %>
+                <% fh = FormHelper::One.new(f: f, context: self, errors: @handler_result&.errors) %>
```
RoyEJohnson, line 20: The `errors` parameter here is what triggers the pink box and error message on the form.

app/controllers/application_controller.rb
```diff
@@ -1,15 +1,18 @@
 class ApplicationController < ActionController::Base
+  include ApplicationHelper
+
   layout 'application'
```
RoyEJohnson, line 4: This still wants the original views/layouts/application.html.erb

PR #1036, opened 2022-03-02
Code cleanup

RoyEJohnson:
A less ambitious first pass at integrating newflow into the whole.
I don't think anything is broken.


RoyEJohnson:
Broke this up into multiple smaller PRs.

PR #1039, opened 2022-03-03
add openstax user account to salesforce

mwharrison:


app/routines/add_account_to_salesforce.rb
```diff
@@ -0,0 +1,27 @@
...
+      sf_ox_account.account_id = user.id
+      sf_ox_account.account_role = user.role.capitalize,
+      sf_ox_account.salesforce_contact_id = user&.salesforce_contact_id,
+      sf_ox_account.salesforce_lead_id = user&.salesforce_lead_id,
+      sf_ox_account.signup_date = user.created_at.strftime("%Y-%m-%dT%H:%M:%S%z")
+      sf_ox_account.account_environment = Rails.application.secrets.environment_name
+
+      if sf_ox_account.save!
+        user.salesforce_ox_account_id = sf_ox_account.id
+      else
```
Dantemss, line 19: It's not a real ActiveRecord object so I'm not 100% sure but I think `save!` will throw so the else will never run

app/routines/add_account_to_salesforce.rb
```diff
@@ -0,0 +1,27 @@
...
+        Sentry.capture_message(sf_ox_account.errors) unless sf_ox_account.save!
```
Dantemss, line 20: No point in calling `save!` again

app/routines/update_user_contact_info.rb
```diff
@@ -46,11 +46,14 @@ def call
...
-        event_type: :user_contact_id_updated_from_salesforce,
-        event_data: { contact_id: sf_contact.id }
-      )
+      if user.salesforce_contact_id.blank?
+        user.salesforce_contact_id = sf_contact.id
+      elsif user.salesforce_contact_id != sf_contact.id
+        user.salesforce_contact_id = sf_contact.id
+        warn("Updating a contact ID (#{sf_contact.id}) on an account (#{user.id}) which is different than what was previously linked.")
+      else
+        warn("Unable to update contact ID (#{sf_contact.id}) on an account (#{user.id}). Check user account security log for details.")
```
Dantemss, line 55: I think this code runs for all users who have SF contacts, every time, even if there is nothing to update, so this warn will be printed a lot and not be useful.

app/routines/update_user_contact_info.rb
```diff
@@ -46,11 +46,14 @@ def call
 
       user.salesforce_contact_id = sf_contact.id
 
-      SecurityLog.create!(
-        user: user,
-        event_type: :user_contact_id_updated_from_salesforce,
-        event_data: { contact_id: sf_contact.id }
-      )
```
Dantemss, line 53: It's probably fine to save this when the contact is actually updated, just don't save it when there are no changes.

db/migrate/20220303200215_remove_support_identifier.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveSupportIdentifier < ActiveRecord::Migration[5.2]
+  def change
+    if column_exists? :users, :support_identifier
+      remove_column :users, :support_identifier
+    end
```
Dantemss, line 5: Not sure how this if logic interacts with Rails trying to automatically generate the reverse migration for the `change` method.

I would suggest just using `remove_column` without the `if`, remove_column by itself without a column type is not reversible.

Dantemss (changes requested):
See comments.
Also it doesn't matter much, but I think some of the commits will disappear if you do a `git rebase main`, there's stuff from other PRs in the diff.

app/routines/add_account_to_salesforce.rb
mwharrison: fixed - and moved this to the (seemingly) more standard way of running background job - using the 'jobs' folder

app/routines/update_user_contact_info.rb
mwharrison: Changed this to only log when the id changes. I expect a lot of this code can be simplified once we get these accounts syncing with Salesforce.

db/migrate/20220303200215_remove_support_identifier.rb
mwharrison: fixed, I probably just had this issue locally - should have reset my db first.

Dantemss (approved):

app/jobs/add_account_to_salesforce_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    )
+
+    sf_ox_account.account_uuid = user.uuid
+    sf_ox_account.account_role = user.role.capitalize,
+    sf_ox_account.salesforce_contact_id = user&.salesforce_contact_id,
+    sf_ox_account.salesforce_lead_id = user&.salesforce_lead_id,
+    sf_ox_account.signup_date = user.created_at.strftime("%Y-%m-%dT%H:%M:%S%z")
+    sf_ox_account.account_environment = Rails.application.secrets.environment_name
+
+    sf_ox_account.save
```
Dantemss, line 17: Maybe you want `save!` here to avoid updating the user if the SF record fails to save.

PR #1060, opened 2022-04-05
Integrate newflow controllers

RoyEJohnson:


app/controllers/password_management_controller.rb
```diff
@@ -0,0 +1,106 @@
+class PasswordManagementController < BaseController
+  include LoginSignupHelper
+
+  fine_print_skip :general_terms_of_use, :privacy_policy, only: [
+    :forgot_password_form, :send_reset_password_email, :reset_password_email_sent
+  ]
+
+  before_action :newflow_authenticate_user!, only: [:create_password, :educator_sheerid_form]
```
mwharrison, line 8: we need to replace `newflow_authenticate_user` with `authenticate_user`, which I don't think is being used anymore (unlike `authenticate_admin`). This can be a future PR though.

config/routes.rb
```diff
@@ -33,14 +33,14 @@
     get 'i/signout', action: :logout, as: :newflow_logout
   end
 
-  scope controller: 'newflow/signup' do
+  scope controller: 'signup' do
```
mwharrison, line 36: I think this will allow us to do a blanket match on `i` urls and redirect them with params with a nice function-isk thing!

mwharrison (approved):

PR #1061, opened 2022-04-05
redirect sheer id with params

mwharrison:


app/helpers/application_helper.rb
```diff
@@ -156,7 +152,7 @@ def standard_field(options={})
   end
 
   def is_real_production_site?
-    request.host == 'accounts.openstax.org' || 'openstax.org'
+    request.host == 'accounts.openstax.org'
```
RoyEJohnson, line 155: Is this the host, even if the URL is openstax.org/accounts ?

RoyEJohnson (approved):
One question.

app/helpers/application_helper.rb
mwharrison: it is - but it was causing it do it on all the test environments too (which was kind of annoying).. probably should figure out how to check the host _and path_

PR #1080, opened 2022-06-01
Preauth and migrations

mwvolo:
Breaking up PR#1070 for easier review and spec updating. 

Starting with the migrations that have already been run on dev and QA.

This includes:
- removing the newflow feature flags
- the rails-db-settings, banners
- messages
- salesforce streaming
- updating the default state for new users from needs_profile => unverified
- removing email sent and needs sync from the user
- adding renewal eligible to support the new renewal form


app/handlers/signup_start.rb
```diff
@@ -23,43 +23,6 @@ def handle
...
-
-
-    outputs.pre_auth_state = new_pre_auth_state
-    deliver_validation_email # Send the pin
-  end
-
-  def deliver_validation_email
-    SignupConfirmationMailer.instructions(
-      pre_auth_state: outputs.pre_auth_state
-    ).deliver_later
```
jpslav, line 62: Because pre auth states are being created today on production, we should make sure that they (and their associated code) really really aren't needed.

app/handlers/signup_verify_email.rb
```diff
@@ -15,12 +15,7 @@ def authorized?
   end
 
   def handle
-    run(ConfirmByPin, contact_info: options[:pre_auth_state], pin: pin_params.pin)
-    # lms students do not receive passwords so the account needs to be created now
-    if options[:pre_auth_state].signed_student?
-      run(SignupExternalStudent, pre_auth_state: options[:pre_auth_state], already_verified: true)
-      options[:session].sign_in!(outputs.user)
```
jpslav, line 22: Tutor choosing to not use the external student stuff doesn't necessarily mean we should get rid of it.  I know I agreed with the opposite earlier, but having second thoughts.  If we have code that works and tests for it and an expectation that it will be used in the near term, should we consider keeping it?

config/app.yml
```diff
@@ -1,112 +0,0 @@
...
-# require 'active_support'; include ActiveSupport::Inflector
-# books = File.read('/tmp/books.txt').split("\n")
-# books.sort.each{|bt| (b,t)=bt.split("\t"); puts underscore(b.gsub(/([\W|\s]+)/, '_'))+":\n  title: #{t}\n  sf: #{b}\n" }; 1
-
-defaults: &defaults
-  push_salesforce_lead_enabled: false
-  user_info_error_emails_enabled: false
-  send_google_analytics: false
-  google_analytics_code: UA-73668038-2
-  google_tag_manager_code: GTM-W6N7PB
```
jpslav, line 13: If rails-settings still needs to be used, adding comment here that maybe need to keep this content here?

db/migrate/20220502215806_change_default_state_for_user.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDefaultStateForUser < ActiveRecord::Migration[5.2]
+  def change
+    change_column_default(:users, :state, :unverified)
```
jpslav, line 3: What's up with this change? `"needs_profile"` looks like it is the starting state, is it not? 

```ruby
class User < ApplicationRecord

  VALID_STATES = [
    TEMP = 'temp', # deprecated but still could exist for old accounts
    NEW_SOCIAL = 'new_social',
    UNCLAIMED = 'unclaimed',
    NEEDS_PROFILE = 'needs_profile', # has yet to fill out their user info
    ACTIVATED = 'activated', # means their user info is in place and the email is verified
    UNVERIFIED = 'unverified', # means their user info is in place but the email is not yet verified
  ].freeze
```

Should there be other changes that come with this?

db/migrate/20220303200215_remove_support_identifier.rb
```diff
@@ -1,5 +1,5 @@
 class RemoveSupportIdentifier < ActiveRecord::Migration[5.2]
   def change
-    remove_column :users, :support_identifier
+    remove_column :users, :support_identifier, :citext
```
jpslav, line 3: just checking - we for sure want to remove this? (thought maybe you said we didn't)

db/migrate/20220511185320_add_renewal_eligible_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddRenewalEligibleToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column(:users, :renewal_eligible, :boolean)
```
jpslav, line 3: Why is this field in Accounts?

spec/controllers/faculty_access_controller_spec.rb
```diff
@@ -1,33 +0,0 @@
...
-  end
-
-  context 'when the educator feature flag is ON' do
-    before do
-      Settings::FeatureFlags.educator_feature_flag = true
-    end
-
-    it 'redirects to the SheerID form' do
-      get(:apply)
-      expect(response).to redirect_to(educator_sheerid_form_path)
```
jpslav, line 30: This behavior still exists on production, btw.  Keep test?

spec/routines/user_from_pre_auth_state_spec.rb
```diff
@@ -1,29 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe UserFromPreAuthState, type: :routine do
```
jpslav, line 3: If really deleting PreAuth states, then should delete the routine this spec is testing

Gemfile
```diff
@@ -147,12 +147,6 @@ gem 'awesome_print'
 
 gem 'whenever', require: false
 
-# Admin toggles
-gem 'rails-settings-ui'
```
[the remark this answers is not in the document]
mwvolo: The path has been removed as well as admin templates.

I did forget to remove the locals. I can do that.

All these settings have moved to the[ secrets.yml file](https://github.com/openstax/accounts/pull/1080/files#diff-64b9f8f7be8230dfdad3489d4281d95973194a83948805e6d2b805925b934525R155).

Gemfile
```diff
@@ -200,9 +190,6 @@ group :development, :test do
   #
   # Call 'debugger' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', require: false
-  # Debug in VS Code
-  gem 'ruby-debug-ide', require: false
-  gem 'debase', require: false
```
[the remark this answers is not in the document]
mwvolo: I don't know of anyone using VS code on BIT and removing them is just one less dependency to worry about (that was my thinking anyways).

I can leave them in if you think it's worth keeping.

db/migrate/20220502215806_change_default_state_for_user.rb
mwvolo: I'm basing this change on this comment: https://github.com/openstax/accounts/blob/a6e9cae6eb24c506674a574dff93d00dca593d84/app/routines/ensure_unverified_user.rb#L1

db/migrate/20220303200215_remove_support_identifier.rb
mwvolo: we do - this was supposed to be an identifier so we could hide the uuid, I believe? It's certainly not being used by anyone - checked with Dante and CS. Attaching relevant conversation.

Screen Shot 2022-06-01 at 17 06 06

db/migrate/20220511185320_add_renewal_eligible_to_user.rb
mwvolo: Accounts is already providing the users contact id from Salesforce and doing the heavy job of syncing with contacts - so instead of adding this to the CMS.. I thought it was best to add it here.

It should eventually be moved to the salesforce-api - along with the contact id, faculty status, etc.

spec/controllers/faculty_access_controller_spec.rb
mwvolo: I think we need to remove these [routes](https://github.com/openstax/accounts/blob/main/config/routes.rb#L172-L175) or just add a match/redirect in routes. Honestly, I can't imagine why anyone/thing would be hitting the /apply route after two years of this being released.

app/handlers/signup_verify_email.rb
mwvolo: In the final iteration of this refactor, which is currently on dev, you can direct a student directly to the student signup form.
https://dev.openstax.org/accounts/signup/start?role=student

Perhaps this is slightly different than what this code is doing though? Or just need to have it accept an email param to prefill the student signup email for ease?

app/handlers/signup_start.rb
mwvolo: This handler is only being called here (which is currently not how users signup for an account): https://github.com/openstax/accounts/blob/ca7aa5445eaa91d78aaa6e05ab6b4582b228b8c7/app/controllers/legacy/signup_controller.rb#L27


SignupConfirmationMailer is [not sent by anything else](https://github.com/openstax/accounts/search?q=SignupConfirmationMailer) in the code.

mwvolo: This is the rabbit hole of PR#1070 😂 

This file needs to be removed.. it's confusing.

app/handlers/sessions_create.rb
```diff
@@ -224,14 +174,6 @@ def authentication_user
     @authentication_user ||= authentication.user
   end
 
-  def signing_up?
-    pre_auth_state.present?
```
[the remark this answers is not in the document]
mwvolo: HA!

This only place this class is being called is in a spec!
https://github.com/openstax/accounts/search?q=SessionsCreate

Gemfile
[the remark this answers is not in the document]
jpslav: ok to kill.

app/handlers/sessions_create.rb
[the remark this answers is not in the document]
jpslav: That is HEAD of main.  Aren't we supposed to be looking at 12.1.3 (what is deployed)? This handler still used from the legacy controller which I thought was still being used?

https://github.com/openstax/accounts/blob/v12.1.3/app/controllers/legacy/sessions_controller.rb#L89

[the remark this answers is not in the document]
mwvolo: Sorry - yes. You are correct (and likely a cause of a bug!)

Also, might be worth looking at the updated version:
https://github.com/openstax/accounts/blob/rejected-leads-add-to-release/app/controllers/sessions_controller.rb

app/handlers/signup_verify_email.rb
jpslav: the LMS sends a signed email address and name to Accounts, which we trust without needing the student to reverify it.  Leads to an accelerate Account creation, which is a nice benefit for users using an LMS.  That's what this was doing.  

But, as Dante said, Tutor stopped using this because of a Tutor bug, so we could kill it and just remember to come back for it later.  You can resolve this if you want to.

db/migrate/20220502215806_change_default_state_for_user.rb
jpslav: wfm

db/migrate/20220511185320_add_renewal_eligible_to_user.rb
jpslav: ah ok that makes sense.

.gitignore
```diff
@@ -20,13 +20,7 @@
...
 
 # Ignore public precompiled assets
-/public/assets
-
-# Ignore precompiled static error pages
-/public/404.html
-/public/422.html
-/public/500.html
-/public/503.html
+/public
```
Dantemss, line 23: I think you wanna revert this one because there are things in /public that we do want in the repo like the favicon.ico, robots.txt and the fonts

app/handlers/sessions_create.rb
```diff
@@ -59,7 +57,7 @@ def get_status
 
     if logging_in?
       handle_during_login
-    elsif signing_up?
+    elsif signing_in?
```
Dantemss, line 60: Shouldn't this be signing_up?
To me, signing_in is the same as logging_in

app/handlers/sessions_create.rb
```diff
@@ -224,12 +212,8 @@ def authentication_user
...
   end
 
-  def signing_up?
-    pre_auth_state.present?
-  end
-
-  def pre_auth_state
-    options[:pre_auth_state]
+  def signing_in?
+    buybug
```
Dantemss, line 216: buybug?

config/secrets.yml
```diff
@@ -246,9 +269,20 @@ production:
     security_token: <%= ENV['SALESFORCE_SECURITY_TOKEN'] %>
     consumer_key: <%= ENV['SALESFORCE_CONSUMER_KEY'] %>
     consumer_secret: <%= ENV['SALESFORCE_CONSUMER_SECRET'] %>
+    push_lead_enabled: <%= ENV['SALESFORCE_PUSH_LEAD_ENABLED'] || true %>
```
Dantemss, line 272: It is impossible to set this value to false, since it has `|| true`.

Also, this applies to all the boolean ENV variables: ENV variables are always null or string so it may be a good idea to cast them to boolean using the activerecord boolean type caster thing. Alternatively, you can do this: `ENV['SALESFORCE_PUSH_LEAD_ENABLED'] != 'false'` if you want the default value to be true or `ENV['SALESFORCE_PUSH_LEAD_ENABLED'] == 'true'` if you want the default value to be false.

config/secrets.yml
```diff
@@ -246,9 +269,20 @@ production:
...
+    sync_accounts_enabled: <%= ENV['SALESFORCE_SYNC_ACCOUNTS_TO_SALESFORCE'] || true %>
+    sync_contacts_enabled: <%= ENV['SALESFORCE_SYNC_CONTACT_ENABLED'] || true %>
+    number_of_days_contacts_modified: <%= ENV['SALESFORCE_NUMBER_OF_DAYS_MODIFIED'] || 7 %>
```
Dantemss, line 275: Might wanna use `.to_i` or similar to cast the string to an integer (or use '7' as the default value instead and handle as a string).

Dantemss (changes requested):

db/migrate/20220428191146_create_settings_again.rb
```diff
@@ -0,0 +1,17 @@
+class CreateSettingsAgain < ActiveRecord::Migration[5.2]
+  def self.up
+    drop_table(:settings) if table_exists?(:settings)
+
+    create_table :settings do |t|
```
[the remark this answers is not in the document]
jpslav: The creation of the table and the destroying of it were in the previous PR that was already deployed, so we were keeping them here to be consistent with migrations that have happened.

db/migrate/20220601160740_drop_settings_table.rb
```diff
@@ -0,0 +1,5 @@
+class DropSettingsTable < ActiveRecord::Migration[5.2]
+  def change
+    drop_table(:settings) if table_exists?(:settings)
```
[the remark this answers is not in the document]
mwvolo: Ha, I know - I was trying to keep the migrations the same as what had already been deployed. In the other PR I removed the old migration - which I feel like might be a no-no. So I had to rename the existing migration and make a new one to drop the table. Sorry for the messy migrations.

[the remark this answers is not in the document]
mwvolo: I don't want to break things, so I'm probably going overboard on this. If there is a better way let me know. Basically, don't want to have to revert migrations that had been deployed to dev/qa.

config/secrets.yml
jpslav: Good catch Dante.  At one point I thought we had some utilities for helping with env vars.  Just found [Nenv](https://github.com/e2/nenv) which seems kinda cool too.

app/routines/confirm_by_pin.rb
```diff
@@ -4,15 +4,8 @@ class ConfirmByPin
...
-    value_method = case contact_info
-    when ContactInfo
-      :value
-    when PreAuthState
-      :contact_info_value
-    end
-
     SequentialFailure.confirm_by_pin
-                     .find_or_initialize_by(reference: contact_info.send(value_method)).tap do |sf|
+                     .find_or_initialize_by(reference: contact_info.send(:value)).tap do |sf|
```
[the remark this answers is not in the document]
jpslav: Yes, `.send(...)` was used before because the method name was in the `value_method` variable.

app/handlers/sessions_create.rb
jpslav: :-)

`byebug` 

But of course, we don't want it in here anyway :-)

app/handlers/sessions_create.rb
jpslav: agree - signing_up should stick around probably, tho I get from below you were testing getting here.

mwvolo: honestly, this file is not used and we are removing PreAuth, which is what triggered the signing_up. I'll change it back.. but there will be a change to the method you might not like too much until we get this file out of here 😅

spec/routines/user_from_pre_auth_state_spec.rb
mwvolo: Deleting this routine is causing a somewhat cascading effect with other routines. I'm going to push the changes before I start messing with specs (which will likely be a mess)

config/secrets.yml
mwvolo: changed - I should have noticed all your casting on other secrets 🤦🏼‍♂️

app/handlers/sessions_create.rb
```diff
@@ -30,7 +30,6 @@ class SessionsCreate
   uses_routine TransferAuthentications
   uses_routine TransferOmniauthData
   uses_routine ActivateUnclaimedUser
-  uses_routine TransferPreAuthState
```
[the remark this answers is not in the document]
mwvolo: done

app/handlers/sessions_create.rb
```diff
@@ -225,11 +213,7 @@ def authentication_user
   end
 
   def signing_up?
-    pre_auth_state.present?
-  end
-
-  def pre_auth_state
-    options[:pre_auth_state]
+    nil
```
mwvolo, line 216: hmm, instead of `nil`, maybe I should throw a Sentry message here so we can verify nobody is still getting through this path?

app/handlers/signup_verify_email.rb
mwvolo: I'm going to resolve because if we want to do this again, I think we'll need to rearchitect the way we are passing the information around. It also looks like we are more interested in moving towards the iframe login/signup over signed params from conversations with Tom.

config/secrets.yml
jpslav: Secrets aren't nil or a string, but environment variables are read as strings.

app/handlers/sessions_create.rb
jpslav: sure

.gitignore
mwvolo: these were on my local filesystem 🤔 
I can remove this change and not check them in.. 

OH - I think it's because I was locally trying to debug the compile_asset command and built them with the production environment 🤦🏼 - I'll remove.

config/secrets.yml
mwvolo: had to revert this back - it is actually being handled as a string in the code already

db/migrate/20220531172911_add_index_to_contact_info_value.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToContactInfoValue < ActiveRecord::Migration[5.2]
+  def change
+    add_index :contact_infos, [:value, :type], unique: true
```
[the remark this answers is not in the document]
mwvolo: removed this migration from this PR, so going to resolve this - we'll cross this bride in a future PR (probably the rubocop one.. since that's what started this)

app/handlers/sessions_create.rb
mwvolo: done

mwvolo:
@jpslav @Dantemss still having some issues with tests (some of the Capybara errors are particularly confusing... some of them are working locally on my machine 🤨)

Anyways, probably good to have you look over again before I spend too much time on the specs.

config/secrets.yml
```diff
@@ -212,18 +234,24 @@ test:
...
-  cms_api_url: https://cms-dev.openstax.org/apps/cms/api/
+  cms_api_url: https://dev.openstax.org/apps/cms/api/
   sentry:
       dsn: <%= ENV['SENTRY_DSN'] || 'invalid' %>
 
+  rescue_from:
+    raise_exceptions: <%= ActiveModel::Type::Boolean.new.cast ENV.fetch('RESCUE_FROM_RAISE_EXCEPTIONS', false) %>
+    raise_background_exceptions: <%= ActiveModel::Type::Boolean.new.cast ENV.fetch('RESCUE_FROM_RAISE_BACKGROUND_EXCEPTIONS', false) %>
+
 production:
```
jpslav, line 248: is this used in production @Dantemss ? (vs pulling from parameter store)

jpslav (changes requested):

config/secrets.yml
mwvolo: even if not, won't the rescue from config complain if they are not set?

Dantemss: Yeah for tutor-deployment and bit-deployment I decided to keep the production secrets template in this file as is standard and add those ENV variable substitutions for the production secrets. The deployment script writes a .env file from the parameter store secrets instead, that way I can do proper type casting in secrets.yml and the deployment itself is not ruby-specific.

mwvolo: @jpslav please unresolve this comment if this doesn't address your concern

Dantemss:
> @jpslav @Dantemss still having some issues with tests (some of the Capybara errors are particularly confusing... some of them are working locally on my machine 🤨)
> 
> Anyways, probably good to have you look over again before I spend too much time on the specs.

If they work on your machine but not on GH actions, usually it's some kind of race condition due to the test not waiting, although most capybara methods that return things from the page should automatically wait for the page to render...

mwvolo:
@Dantemss @jpslav ready for another 👀 
I promise the following PRs will be much less complex 😅

app/views/layouts/_google_analytics.html.erb
```diff
@@ -1,11 +1,11 @@
-<% if Rails.env.production? and Settings::GoogleAnalytics.send_google_analytics == true and !Settings::GoogleAnalytics.google_analytics_code.blank?  %>
+<% if Rails.env.production? and Settings::GoogleAnalytics.send_google_analytics == true and !Rails.application.secrets[:google_analytics_code].blank?  %>
```
mwvolo: I was thinking... "why didn't I catch this locally"... `if Rails.env.production?` 🤦🏼 😂

config/secrets.yml
Dantemss: I would change it to "7" just to make it clear it works for strings like the ENV var and to make sure both sides of the `||` are the same type. The YAML template will substitute the string in and it'll look like:

number_of_days_contacts_modified: 7

so then that gets automatically converted to a number...

spec/handlers/sessions_create_shared_examples.rb
```diff
@@ -19,20 +18,20 @@
...
+    #   it "returns mismatched_authentication status and does not log in" do
+    #     result = nil
+    #     expect do
+    #       result = handle(
+    #         request: MockOmniauthRequest.new(authentication.provider, authentication.uid, {})
+    #       )
+    #     end.not_to change { user_state.signed_in? }
+    #     expect(result.outputs.status).to eq :mismatched_authentication
+    #   end
+    # end
```
Dantemss, line 34: Maybe this is important to test?

spec/handlers/sessions_create_shared_examples.rb
```diff
@@ -58,91 +57,67 @@
...
+    #     expect(user_state).to be_signed_in
+    #     user = user_state.current_user
+    #     authentication = user.authentications.reload.first
+    #     expect(authentication.provider).to eq "facebook"
+    #     expect(authentication.uid).to eq "zuckerberg"
+    #     expect(user.contact_infos.size).to eq 1
+    #     expect(user.contact_infos.first.value).to eq "bob@bob.com"
+    #     expect(user.contact_infos.first).to be_verified
+    #   end
+    # end
```
Dantemss, line 115: Seems like all the social auth signup tests got removed

spec/handlers/signup_password_spec.rb
```diff
@@ -1,77 +1,71 @@
...
-  context "when the user is already logged in with a non-anonymous user" do
-    it "freaks out" do
-      user = create_user 'user'
-
-      expect{
-        described_class.handle(params: {signup: {}}, caller: user)
-      }.to raise_error(Lev::SecurityTransgression)
-    end
-  end
+  ## MV: This handler was replaced with the create_password handler and set_password routine.
```
Dantemss, line 4: Delete the file?

Dantemss (approved):
I had some comments but nothing that should hold back the PR I don't think

.github/workflows/migrations.yml
```diff
@@ -66,14 +58,6 @@ jobs:
...
           git checkout --force -
 
-      # Retrieve gem cache for PR merge commit
-      - uses: actions/cache@v2
-        with:
-          path: vendor/bundle
-          key: ${{ runner.os }}-gems-pr-${{ hashFiles('**/Gemfile.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-gems-pr-
-
```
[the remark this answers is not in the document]
mwvolo:
```
- uses: ruby/setup-ruby@v1
  with:
    bundler-cache: true
```
Do I need to keep any of those `restore-key` declarations?

spec/handlers/sessions_create_shared_examples.rb
mwvolo: I think this is replaced with this spec: https://github.com/openstax/accounts/blob/main/spec/handlers/oauth_callback_spec.rb

But let me reenable and see what happens (think it's going to fail on that pesky `signing_up` method).

spec/handlers/sessions_create_shared_examples.rb
mwvolo: again, this is testing old social auth.. notice no mentions of newflow - which everything is now using newflowfacebook and newflowgoogle

spec/handlers/signup_password_spec.rb
mwvolo: going to keep it here for reference.. will remove before final refactor PR though

spec/handlers/sessions_create_shared_examples.rb
mwvolo: it does, and this is in the `signing_up` context, so that's why.. going to keep here for reference though

mwvolo: the `logging_in` specs are still running

PR #1093, opened 2022-07-06
refactor base

mwvolo:
Moves the /profile PUT update to the other controller with the rest of the profile.

This will be refactored into the profile_controller in PR#1084


Dantemss:
I think this is fine, but is there another fix needed in the FE? Or does this controller take care of it somehow?

mwvolo:
I thought it would take care of it, looks like I need to update the post path here too: https://github.com/openstax/accounts/blob/main/app/views/base/_edit-js.html.erb#L6

mwvolo:
I'm going to merge after I can build and test in front of cloudfront

config/routes.rb
```diff
@@ -3,14 +3,14 @@
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
   root to: 'static_pages#home'
 
-  match 'i/signup/(*path)' => redirect { |_,request| "signup/#{request.params[:path]}?#{request.params.except('path').to_query}" }, via: :all
+  get 'i/signup/(*path)' => redirect { |_,request| "signup/#{request.params[:path]}?#{request.params.except('path').to_query}" }, via: :get
```
RoyEJohnson, line 6: get via get seems redundant. I thought you only needed a via clause if you were using match.

RoyEJohnson (approved):
Nothing jumped out at me as really wrong. One comment about which I'm unsure.

config/routes.rb
Dantemss: I missed this but yeah, I think `via` only does anything if you use `match`.

mwvolo: @Dantemss and you said to only use these for GET requests, right? So just remove `via:`?

app/handlers/verify_email_by_pin.rb
```diff
@@ -16,7 +16,7 @@ def authorized?
   end
 
   def handle
-    result = ConfirmByPin.call(contact_info: options[:email_address], pin: confirm_params.pin)
+    result = ConfirmByPin.call(contact_info: options[:email_address])
```
Dantemss, line 19: This fails without the pin

app/routines/confirm_by_pin.rb
```diff
@@ -1,7 +1,7 @@
  class ConfirmByPin
   lev_routine
 
-  uses_routine ConfirmContactInfo
+  uses_routine VerifyEmailByPin
```
Dantemss, line 4: VerifyEmailByPin is a handler so calling it doesn't work properly

Dantemss (changes requested):
The oauth providers got changed so we need a migration to rename them in the database
And remove duplicates in case someone had both the old and the new ones

PR #1098, opened 2022-07-19
Minimal changes to combine newflow and old flow

Dantemss:
Still need to fix specs and combine styles/layout and OAuth providers


app/controllers/faculty_access_controller.rb
```diff
@@ -1,7 +1,7 @@
 class FacultyAccessController < ApplicationController
```
mwvolo, line 1: @Dantemss not sure there is still a path to this controller - I think the faculty access application process is handled in `educator_signup_controller`

app/mailers/sign_in_help_mailer.rb
```diff
@@ -1,4 +1,5 @@
 class SignInHelpMailer < ApplicationMailer
+  include Rails.application.routes.url_helpers
```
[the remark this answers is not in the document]
Dantemss: I'm not 100% sure mailers need this, but generally non-controller files that need access to the url helpers

Dantemss:
The only failure I still get locally is the specs that admins get logged out after 30 min, but only when running multiple specs in a row, oddly enough.

config/routes.rb
```diff
@@ -76,7 +62,7 @@
 
     # Step 3
     get 'signup/educator/apply', action: :educator_sheerid_form, as: :educator_sheerid_form
-    post 'i/sheerid/webhook', action: :sheerid_webhook, as: :sheerid_webhook
+    post 'sheerid/webhook', action: :sheerid_webhook, as: :sheerid_webhook
```
mwvolo, line 65: the `/i/` in this post path is hard-coded in the sheer id platform and might be difficult to change (no, it is... I check with them).
Is there some nginx magic to handle this?

spec/support/feature_helpers.rb
```diff
@@ -245,10 +252,12 @@ def complete_login_password_screen(password)
 end
 
 def complete_reset_password_screen(password=nil)
+  expect(page.current_path).to eq(password_reset_path)
   password ||= 'Passw0rd!'
-  fill_in (t :"identities.set.password"), with: password
-  fill_in (t :"identities.set.confirm_password"), with: password
+  fill_in 'set_password_password', with: password
```
mwvolo, line 257: just so I know - what are you using as the selector here? The form field name?

mwvolo (approved):

spec/support/feature_helpers.rb
Dantemss: The form field id

config/routes.rb
Dantemss: We can add a rewrite I think

PR #1099, opened 2022-07-20
Remove newflow social

mwvolo:


db/data/20220720142258_change_newflow_social_to_facebook_google.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class ChangeNewflowSocialToFacebookGoogle < ActiveRecord::Migration[5.2]
```
Dantemss, line 3: We may need `disable_ddl_transaction!` in the class depending on how many records there are in prod. If there are too many, the transaction will cause issues due to locking a bunch of rows

db/data/20220720142258_change_newflow_social_to_facebook_google.rb
```diff
@@ -0,0 +1,19 @@
...
+  def up
+    Authentication.where(provider: 'facebooknewflow').each do |auth|
+      auth.provider = 'facebook'
+      auth.save
+    end
+
+    Authentication.where(provider: 'googlenewflow').each do |auth|
+      auth.provider = 'google_oauth2'
+      auth.save
+    end
```
Dantemss, line 13: Maybe we can add another migration in the future that runs these updates again and then does a `Authentication.where(provider: ['facebooknewflow', 'googlenewflow']).delete_all` at the end. In a separate release, so the backend code is already updated and not using these old providers.

spec/features/user_login_signup_with_social_network_spec.rb
```diff
@@ -1,242 +0,0 @@
-require 'rails_helper'
-
-feature 'User logs in or signs up with a social network', js: true do
```
Dantemss, line 3: Is it expected that this whole file is gone? I feel like it could be made to work?

spec/features/require_recent_sign_in_to_change_authentications_spec.rb
```diff
@@ -69,45 +69,20 @@
     end
   end
 
-  # scenario 'bad password on reauthentication' do
```
Dantemss, line 72: Is it worth restoring this spec too?

db/data/20220720142258_change_newflow_social_to_facebook_google.rb
mwvolo: added

spec/features/user_login_signup_with_social_network_spec.rb
mwvolo: readded and passing 🎉

spec/features/require_recent_sign_in_to_change_authentications_spec.rb
mwvolo: readded and passing

Dantemss (approved):

PR #1106, opened 2022-07-26
Refactor controllers

Dantemss:
Pull in refactored controllers, routines, handlers, routes


app/assets/javascripts/application/ui.js
```diff
@@ -0,0 +1,32 @@
+(function() {
```
mwvolo, line 1: do we need both `application/ui.js` and `application/accounts-ui.js`?

app/assets/javascripts/multiselect/index.js
```diff
@@ -1,2 +1,2 @@
-//= require knockout
+//= require knockout.js
```
mwvolo, line 1: when does something need .js on the end vs not?

app/assets/javascripts/profile/index.js
```diff
@@ -1,16 +1,18 @@
-//= require bootstrap-editable
-//= require ../vendor/underscore
+//= require jquery
+//= require jquery_ujs
+//= require bootstrap-sprockets
+//= require underscore.js
+//= require x-editable-bootstrap.js
```
mwvolo, line 5: any reason for the change to bootstrap-editable then back to x-editable-bootstrap?

mwvolo:
ah, I started reviewing - but I think the merge of your other PR created conflicts, so I'll wait until those are resolved @Dantemss

app/assets/javascripts/multiselect/index.js
Dantemss: I don't think it needs this? Likely the PR I pulled it from just had it

Dantemss:
The conflicts are unexpected, since it was based on that PR... oh maybe it's because of the squash merge...

app/assets/javascripts/profile/index.js
Dantemss: I think this change is gone after the rebase

app/assets/javascripts/application/ui.js
Dantemss: They do seem very similar, I imagine they could be combined... I might leave that one for you though :)

Dantemss:
Maybe the sheerid webhook thing should be in some sort of API controller, since it's basically an API call...

Dantemss:
@mwvolo conflicts fixed

Dantemss:
Oh guess I should say the educator/student signup route changes might cause some errors if someone is in the middle of that process when deployed.

Maybe we need to add nginx rules for those. Or do it early morning or something.

app/controllers/login_controller.rb
```diff
@@ -23,9 +25,9 @@ def login
           save_unverified_user(user.id)
 
           if user.student?
```
mwvolo, line 27: no need to check user role here? just `redirect_to(verify_email_by_pin_form_path)`

app/handlers/create_password.rb
```diff
@@ -1,26 +1,34 @@
+# Changes the user's password if they have one, otherwise creates a password (an `Identity`).
 class CreatePassword
   lev_handler
-  uses_routine ::SetPassword
+  uses_routine SetPassword
```
mwvolo, line 4: why remove `::` from here but not line 23? Does `::` do something special?

config/initializers/assets.rb
```diff
@@ -8,6 +8,7 @@
 Rails.application.config.assets.precompile += %w(
```
mwvolo, line 8: do we still need this file with https://github.com/openstax/accounts/blob/refactor_controllers/app/assets/config/manifest.js?

config/routes.rb
```diff
@@ -7,125 +6,111 @@
...
-
   # routes for access via an iframe
   scope 'remote', controller: 'remote' do
     get 'iframe'
     get 'notify_logout', as: 'iframe_after_logout'
   end
 
-  scope controller: 'users' do
+  scope controller: :users do
     put 'profile', action: :update
```
mwvolo, line 110: I think as part of another PR, this should either move to the profile_controller or the profile "show/edit" view should go the users controllers.

mwvolo (approved):
Left a few comments/questions - but overall excellent 👏🏼 💯 
Thanks so much for getting this into a working state, @Dantemss!

app/handlers/create_password.rb
Dantemss: The `::` ensures the SetPassword being called is the top-level one, not SomeModule::SetPassword.
It's not necessary unless you have multiple SetPassword or think you may have them in the future. I can probably remove all of them.

config/initializers/assets.rb
Dantemss: The manifest is used by sprockets 4+ but we still have sprockets 3 in the Gemfile

PR #1109, opened 2022-07-28
Create users with an external_id and return access_tokens

Dantemss:
To be used by the LTI gateway that will provide LTI integration for Rex.
If you create a user with just an external_id, they have no username or email and cannot login.
We'll use that to make sure users can only login via their LMS.


app/routines/find_or_create_user.rb
```diff
@@ -18,19 +18,33 @@ class FindOrCreateUnclaimedUser
   protected
 
   def exec(options)
-    unless options[:email] || options[:username]
-      fatal_error(code: :invalid_input, message: (I18n.t :"routines.find_or_create_unclaimed_user.must_provide_email_or_username"))
```
Dantemss, line 22: I felt this message was for our own internal use and should not be translated

app/representers/api/v1/find_or_create_user_representer.rb
```diff
@@ -12,30 +12,36 @@ class FindOrCreateUserRepresenter < Roar::Decorator
...
              writeable: true,
              schema_info: {
-               required: true,
-               description: "Email address to search by or assign to newly created user"
+               description: "External ID to search by or assign to newly created user"
              }
 
-    property :already_verified,
-             type: :boolean,
+    property :username,
```
mwvolo, line 23: do we / should we continue to have usernames in the API? Is someone still using these?

Dantemss: Tutor uses them when creating preview course students

app/routines/find_or_create_user.rb
```diff
@@ -1,12 +1,12 @@
...
+# Given either an external_id, username, or email address:
 #   attempt to find a user with that attribute.
 #   If the user is found, return the user
-#     Otherwise create a new user with the email and username given,
-#     set it's state to "unclaimed" and return that record
+#     Otherwise create a new user with the given external_id, username, or email,
+#     set it's state to "external" or "unclaimed" and return that record
 
-class FindOrCreateUnclaimedUser
+class FindOrCreateUser
```
mwvolo, line 9: what the reason for find_user and find_or_create_user? Seems they are somewhat duplicated?

mwvolo: oh, I guess this is the routines used when using the find or create api calls...

config/locales/en.yml
```diff
@@ -306,9 +306,6 @@ en:
     confirm_by_code:
       unable_to_verify_address: Unable to confirm email address
 
-    find_or_create_unclaimed_user:
```
mwvolo, line 309: very glad to see these unclaimed users gone.. these were a real point of confusion and don't think they were being used properly...

Dantemss: Well only the translation is gone because I feel like we don't need this :) The unclaimed users themselves are still used by the Tutor preview course feature. You can probably remove them in Fall of 2023

Dantemss:
Tom pointed out that for highlights etc to be able to tell that the token is valid, it needs to be our actual SSO cookie instead of a random doorkeeper-generated token, so they can decode it and read the contents. So I'm setting this back to draft for now.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -218,8 +218,8 @@
   end
 
   context "update" do
-    it "should let User update his own User" do
-      api_put :update, user_2_token, body: {first_name: "Jerry", last_name: "Mouse"}
+    it "should let User update their own name" do
```
mwvolo, line 221: 💯

mwvolo (approved):

PR #1111, opened 2022-08-01
Doorkeeper SSO

Dantemss:
Use SSO cookies as Doorkeeper tokens
Generate a Doorkeeper token when calling find-or-create-user
Also fixed some test warnings


lib/sso_cookie_jar.rb
```diff
@@ -1,76 +1,30 @@
...
-    @cookie_name = cookie_secrets[:name]
-    @cookie_options = cookie_secrets[:options]
-  end
+  secrets = Rails.application.secrets.sso[:cookie]
+  @@cookie_name = secrets[:name]
+  @@cookie_options = secrets[:options]
 
   def subject
-    cookie = self[@cookie_name]
+    cookie = self[@@cookie_name]
```
mwvolo, line 9: Learning something new everyday working with you on this! 
https://stackoverflow.com/questions/5890118/what-does-variable-mean-in-ruby/5890199#5890199

mwvolo (approved):
This is awesome as usual!
Does merging this affect the SSO cookie, as you stated in this thread? 
https://openstax.slack.com/archives/C6NRD0CCQ/p1659370560917219

lib/sso_cookie_jar.rb
Dantemss: Normally class variables are bad if you are going to modify them (they are not thread-safe and propagate to all descendants I think) but here we don't modify it and I just need them to be available in the instance methods so they are fine

Dantemss:
> This is awesome as usual! Does merging this affect the SSO cookie, as you stated in this thread? https://openstax.slack.com/archives/C6NRD0CCQ/p1659370560917219

The cookies themselves should not be affected. Doorkeeper tokens will now look like the cookies and can be decoded by apps that know the secrets, but otherwise should work the same (Doorkeeper works by remembering the tokens though, so you cannot use an arbitrary SSO cookie as a token, only one that was emitted as a token).

PR #1115, opened 2022-08-03
Fetch books from salesforce

mwvolo:
The majority of the issues we see from Accounts signups <-> Salesforce is the book names not matching the Salesforce name. So this should prevent the human error issue where the Salesforce name must be correct in the CMS for things not to blow up. 💥 
We also have a lot more books in the CMS than we actually track adoptions for. Only having the books we track on the signup form helps simplify things.


Dantemss:
Hm I guess I need some context why this is fetched from Salesforce and not the CMS or the ABL

Dantemss:
The code itself looks fine, I think

mwvolo:
What is the ABL? 🤔 
Sorry, I should have added this to the descriptions, will do that now for quick future reference.

The majority of the issues we see from Accounts signups <-> Salesforce is the book names not matching the Salesforce name. So this should prevent the human error issue where the Salesforce name must be correct in the CMS for things not to blow up. 💥 
We also have a lot more books in the CMS than we actually track adoptions for. Only having the books we track on the signup form helps simplify things.

app/routines/update_books_from_salesforce.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    books_updated = 0
+    sf_books      = OpenStax::Salesforce::Remote::Book.order(:Id)
+
+    sf_books.each do |sf_book|
+      book                 = Book.find_or_initialize_by(salesforce_id: sf_book.id)
+      book.salesforce_name = sf_book.name
+      book.official_name   = sf_book.official_name
+
+      books_updated        += 1 if book.save!
```
Dantemss, line 22: `book.save!` just throws on failure
if you want a boolean, you want `book.save` instead

mwvolo: Makes sense - thanks! Fixed.

Dantemss (approved):

PR #1134, opened 2022-08-19
Educator pending fix

mwvolo:


mwvolo:
@Dantemss I know this is a little large - but I am trying to make sense of all these states/faculty statues/ roles... and trying to simplify.
Would mind giving this a look over and see if I'm heading down a crazy path?

app/controllers/login_controller.rb
```diff
@@ -1,81 +1,64 @@
...
-        end
-
+        security_log(:sign_in_failed, { reason: @handler_result.errors.first.code, email: email })
         render :login_form
       }
     )
   end
 
   def logout
+    Sentry.set_user({}) if Rails.env.production?
```
Dantemss, line 46: Again, I think this may have to be a before_action

app/controllers/signup_controller.rb
```diff
@@ -35,36 +30,48 @@ def signup_form
...
+    @signing_up_user = get_login_state
     render :email_verification_form
   end
 
   def verify_email_by_pin_post
+    @signing_up_user = get_login_state
     handle_with(
       VerifyEmailByPin,
-      email_address: unverified_user.email_addresses.first,
+      email_address: current_user.email_addresses.first,
```
Dantemss, line 69: current_user can be nil or AnonymousUser here due to `skip_before_action :authenticate_user!`

app/controllers/signup_controller.rb
```diff
@@ -103,29 +109,36 @@ def verify_email_by_code
...
+  end
 
   def signup_done
     security_log(:sign_up_successful, form_name: action_name)
-    redirect_back if current_user.is_tutor_user?
+    if current_user.source_application&.name&.downcase&.include?('tutor')
+      redirect_back
+    else
+      redirect_to(profile_path)
+    end
```
Dantemss, line 141: I know the code did this already, but feels wrong to send them to the profile page rather than where they were before... for example, if they came here by signing up for highlights, shouldn't they go back to Rex after signup?

app/controllers/social_auth_controller.rb
```diff
@@ -1,81 +1,60 @@
...
+          unless user.activated? # needs social confirmation page for info
             @first_name = user.first_name
-            @last_name = user.last_name
-            @email = @handler_result.outputs.email
+            @last_name  = user.last_name
+            @email      = @handler_result.outputs.email
             security_log(:student_social_sign_up, user: user, authentication_id: authentication.id)
-            # must confirm their social info on signup
-            render :confirm_social_info_form and return # TODO: if possible, update the route/path to reflect that this page is being rendered
+            redirect_to confirm_oauth_info_path and return
```
Dantemss, line 27: This may have to be a before_action too? Otherwise they can escape since the session is already set

app/handlers/oauth_callback.rb
```diff
@@ -27,37 +27,39 @@ def authorized?
...
     end
 
-    outputs.user = outputs.authentication.user
+    @authentication = Authentication.find_or_initialize_by(provider: @oauth_data.provider, uid: @oauth_data.uid)
+    user = User.find_by(id: @authentication.user_id)
+    # User found with the given authentication. We will log them in.
+    # byebug
+    if user.present?
+      outputs.user = user
+      handle_while_logged_in(@authentication.user)
```
Dantemss, line 45: This method should be called if this handler is used by a user that is already logged in (session exists), so that they can add a new social auth or something, but that's not what's happening here.

app/models/identity.rb
```diff
@@ -14,50 +13,9 @@ class Identity < OmniAuth::Identity::Models::ActiveRecord
...
-    if is_password_digest_ssha?
-      authenticate_with_ssha unencrypted_password
-    else
-      super
-    end
-  end
-
   # Returns true if the user is due for resetting their password
   def password_expired?
+    return if DEFAULT_PASSWORD_EXPIRATION_PERIOD.nil?
```
Dantemss, line 18: I could see us wanting someone to reset their password even if we currently don't make passwords expired in general

app/models/user.rb
```diff
@@ -24,14 +22,17 @@ class User < ApplicationRecord
...
 
+  # TODO: Thinking we can migrate to having faculty status == state - instead of having both
   VALID_FACULTY_STATUSES = [
     NO_FACULTY_INFO = 'no_faculty_info',
     PENDING_FACULTY = 'pending_faculty',
     CONFIRMED_FACULTY = 'confirmed_faculty',
     REJECTED_FACULTY = 'rejected_faculty',
     PENDING_SHEERID = 'pending_sheerid',
     REJECTED_BY_SHEERID = 'rejected_by_sheerid',
-    INCOMPLETE_SIGNUP = 'incomplete_signup'
```
Dantemss, line 34: Since this is used to create an enum, users that had incomplete_signup status before will automatically be updated to needs_email_verified which may not be what we want.

Dantemss (changes requested):

app/models/user.rb
mwvolo: We haven't released the incomplete_signup code yet 😅 - so hopefully not an issue?

app/controllers/application_controller.rb
```diff
@@ -28,16 +25,39 @@ def check_if_password_expired
     redirect_to(password_reset_path)
   end
 
-  def return_url_specified_and_allowed?
-    # This returns true if `save_redirect` actually saved the URL
-    params[:r] && params[:r] == stored_url
+  def check_if_signup_complete
```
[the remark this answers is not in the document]
mwvolo: good observation, fixed

app/controllers/identities_controller.rb
```diff
@@ -37,8 +37,7 @@ def send_password_email(kind:)
     # Can be reached before logged in (can't remember password) or when logged
     # in and asked to reauthenticate and can't remember password.
 
-    user = signed_in? ? current_user :
-                        User.find_by(id: get_login_state[:matching_user_ids].try(:first))
+    user = signed_in? ? current_user : nil
```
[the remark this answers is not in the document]
mwvolo: I think this is the password reset route? https://github.com/openstax/accounts/blob/abf0fc8d0a26a7a167b54d9b99774de150abaaf2/config/routes.rb#L85

app/controllers/login_controller.rb
```diff
@@ -1,81 +1,64 @@
...
-          save_unverified_user(user.id)
-
-          redirect_to(verify_email_by_pin_form_path)
-
-          return
+        if Rails.env.production?
+          Sentry.configure_scope do |scope|
+            scope.set_tags(user_role: user.role.humanize)
+            scope.set_user(uuid: user.uuid)
+          end
```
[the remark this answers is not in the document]
mwvolo: yeah, I'll move this work to another PR - removing it for now and I'll keep this in mind
Probably needs to be in user_session_management too

app/controllers/login_controller.rb
```diff
@@ -1,81 +1,64 @@
...
         end
 
-        sign_in!(user, security_log_data: {'email': @handler_result.outputs.email})
+        sign_in!(user, security_log_data: { 'email': @handler_result.outputs.email} )
 
-        if current_user.student? || decorated_user.can_do?('redirect_back_upon_login')
-          redirect_back # back to `r`edirect parameter. See `before_action :save_redirect`.
-        else
-          redirect_to(decorated_user.next_step)
```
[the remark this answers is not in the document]
mwvolo: This should handle that: https://github.com/openstax/accounts/pull/1134/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R47

It's called here now: https://github.com/openstax/accounts/pull/1134/files#diff-5d401230d249c0833d3c04a53f1aeab6c223010a996fe82296635e88a10c8cc7R26

app/controllers/signup_controller.rb
mwvolo: this is just a small rewrite, removing a somewhat confusing method to the (original code)[https://github.com/openstax/accounts/blob/main/app/controllers/signup_controller.rb#L128]

app/controllers/signup_controller.rb
```diff
@@ -35,61 +30,68 @@ def signup_form
   def signup_post
     handle_with(
       SignupForm,
+      # TODO: what is all this stuff getting set for?
       contracts_required: !contracts_not_required,
       client_app: get_client_app,
       is_bri_book: is_bri_book_adopter?,
       success: lambda {
-        save_unverified_user(@handler_result.outputs.user.id)
```
mwvolo, line 42: might need to make sure this isn't needed for something else - but I don't think it is

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,10 @@
 class ApplicationController < ActionController::Base
-  layout 'application'
+  include ApplicationHelper
+
+  layout 'newflow_layout'
 
   before_action :authenticate_user!
-  before_action :complete_signup_profile
```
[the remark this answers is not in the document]
Dantemss: Right, they can escape if they go to a page that is not the profile (most people don't end up in the profile after logging in anyway).

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,10 @@
 class ApplicationController < ActionController::Base
-  layout 'application'
+  include ApplicationHelper
+
+  layout 'newflow_layout'
```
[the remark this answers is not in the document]
mwvolo: this was already happening in the base controller that this is replacing - so the only affected pages would be like error pages (that we're quite ugly..) - so yes, we'll need to check them all

app/controllers/identities_controller.rb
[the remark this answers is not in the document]
Dantemss: I feel like you have 2 controllers that do the same thing. Maybe send_reset and send_add here are not used anymore. They have routes but I don't see a link to the routes.
Maybe the entire IdentitiesController is unused?

app/controllers/signup_controller.rb
Dantemss: Still a problem, I think

app/models/identity.rb
mwvolo: So just remove this return statement?

app/controllers/signup_controller.rb
Dantemss: I think this `if` is never true here because there's already a before_action that redirects them away if they are a tutor user: `skip_signup_done_for_tutor_users`

You can probably remove the `if`

mwvolo: it would have been nice to do this after the tutor sunset - ha!
fixed

app/controllers/signup_controller.rb
```diff
@@ -137,24 +150,19 @@ def skip_to_student_sign_up
...
+    unless current_user.is_anonymous?
+      @total_steps ||= params[:role] == 'student' ? 2 : 4
      end
   end
 
   def skip_signup_done_for_tutor_users
-    return unless current_user.is_tutor_user?
-
+    return unless current_user.source_application&.name&.downcase&.include?('tutor')
     redirect_back(fallback_location: signup_done_path)
```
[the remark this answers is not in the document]
mwvolo: makes sense - removed the `fallback_location`

app/controllers/application_controller.rb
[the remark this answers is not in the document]
Dantemss: Or maybe only Tutor users don't end up in the profile page? Anyway it would affect Tutor users at least.

app/models/identity.rb
Dantemss: Yes but check that we don't have a bunch of expired people in the DB already or something :)

app/controllers/application_controller.rb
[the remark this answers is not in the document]
Dantemss: Might as well rename it to application layout then but that can wait

app/models/identity.rb
mwvolo: Screen Shot 2022-09-07 at 13 25 28
Looks like we have about 35k :/

mwvolo: Screen Shot 2022-09-07 at 13 26 28
Ah - but we only care if they are **going** to expire, yeah?

app/controllers/identities_controller.rb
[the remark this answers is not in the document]
mwvolo: I'm going to try and see if I can get the identities controller back in play then - Roy is working on spiking out a FE for accounts that would depend on those APIs for confirming contact info - and I'm guessing more thought was put into the Ident controller over the password management controller.

app/handlers/oauth_callback.rb
mwvolo: I see - seems like just removing this and letting the social_auth_controller handle the sign_in while outputting the user and authentication is the thing to do here.. let me know if that's not right.

PR #1154, opened 2023-03-09
Contracts signing for Assignable

Dantemss:
For: https://github.com/openstax/unified/issues/2267 and https://github.com/openstax/unified/issues/2268
We decided Accounts should handle users signing Assignable contracts

- Added signed contract info to the user API as private data
- New embeddable route to sign latest contracts by name


app/controllers/api/v1/users_controller.rb
```diff
@@ -222,6 +222,22 @@ def find_or_create
...
+    Creates an external id for a given user.
+
+    Both external_id and user_id must be supplied.
+
+    Only trusted apps may call this API. All others will receive a 403 error.
+
+    If the external_id does not exist, it is created and 201 is returned.
+
+    #{json_schema(Api::V1::ExternalIdRepresenter, include: [:readable, :writable])}
+  EOS
```
Dantemss, line 240: This was added in a previous PR but I forgot to add a description

Dantemss:
It would be nice to include this as well to prevent the contracts page from timing out: https://github.com/lml/fine_print/pull/40

mwvolo (approved):

PR #1195, opened 2023-08-10
Application-specific roles

Dantemss:
For: https://github.com/openstax/unified/issues/2525

- Added roles to application_users, display roles when returning user info
- Admin screen to set roles

Gnarly jquery UI


mwvolo (approved):

app/representers/api/v1/application_user_application_representer.rb
```diff
@@ -15,10 +16,19 @@ class ApplicationRepresenter < Roar::Decorator
              type: String,
              readable: true,
              writeable: false,
+             getter: ->(*) { application.name },
              schema_info: {
                required: true,
                description: "The name of this Application"
              }
 
+    collection :roles,
```
mwvolo, line 25: Just a comment, not needing a change - but any reason to restrict the roles to things like `admin`, `editor`, etc? Just making sure we're okay with anything being in the role field.

Dantemss: hm mainly the card didn't ask for any restrictions and I imagine the roles could be different for each app

jivey (approved):
Ideally we'd split out the JS but since this is for admin only it seems fine.

PR #1196, opened 2023-08-17
update past adopter to be using_openstax for renewal popup

mwvolo:
The renewal popup (the only thing I am aware of looking at this field) is showing folks the popup if their adoption status is `true`. We want the renewal popup to show for these folks, so updating this to reflect that.

### DIFF
--- a/app/routines/update_user_contact_info.rb
+++ b/app/routines/update_user_contact_info.rb
@@ -14,7 +14,7 @@ class UpdateUserContactInfo
   ADOPTION_STATUSES = {
     "Current Adopter" => true,
     "Future Adopter" => true,
-    "Past Adopter" => false,
+    "Past Adopter" => true,
     "Not Adopter" => false
   }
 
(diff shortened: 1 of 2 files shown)


mwvolo:
Specs were failing, so I set the version of Chromedriver to highest available version from here: https://chromedriver.storage.googleapis.com/index.html

Dantemss (commented):
Are you sure marketing doesn't use this to send emails? Also I have a fix for chromedriver on my app roles PR

mwvolo:
Marketing does not use this specific field in the API, no.
This field is controlled in a Salesforce flow, based on if someone has a current adoption with us. 
It is then updated here to be a boolean based on what is in Salesforce (it doesn't flow from accounts to any marketing platform). 
https://github.com/openstax/accounts/blob/main/app/routines/update_user_contact_info.rb#L102

mwvolo:
weird, I see you just updated the webdrivers gem... I couldn't seem to get that working locally but this can wait until your other PR is merged and I'll rebase

PR #1203, opened 2023-09-21
skip terms and privacy policy for external credentials form

jivey:
This is probably what we need to make the external credentials form render, but needs to be tested on dev.


Dantemss (approved):

jivey:
@Dantemss I added a couple small changes to remove the tabs (and add a border so it doesn't look weird) and say "step 1 of 1" because there is no step 2

PR #1204, opened 2023-09-25
Added external_ids to Find(OrCreate)UserRepresenter, allow users to be found by UUID

Dantemss:
Assignable doesn't currently store the external ids. We want to be able to go from Account UUID to external id so we can find the correct LMS user when syncing scores manually.

### DIFF
--- a/app/representers/api/v1/find_user_representer.rb
+++ b/app/representers/api/v1/find_user_representer.rb
@@ -10,7 +10,7 @@ module Api::V1
     property :uuid,
              type: String,
              readable: true,
-             writeable: false
+             writeable: true
 
     property :support_identifier,
              type: String,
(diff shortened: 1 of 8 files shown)


jivey (approved):
Looks good to me, may want to update the controller doc for `uuid` option: https://github.com/openstax/accounts/pull/1204/files#diff-ef20d40ddd2dd2a254d26fd4c59655cbed6b541c5f908b81125d09848c1dfeb6R159

TomWoodward (approved):

Dantemss:
I might also add it to the search API so we can do it for multiple users at once

app/representers/api/v1/find_user_representer.rb
```diff
@@ -10,7 +10,7 @@ class FindUserRepresenter < Roar::Decorator
     property :uuid,
              type: String,
              readable: true,
-             writeable: false
+             writeable: true
```
mwvolo, line 13: what's the rationale behind changing this to be `writable: true`?

mwvolo (approved):

PR #1205, opened 2023-09-29
Fixed wrong index and title for applications added in the admin view

Dantemss:
For: https://github.com/openstax/unified/issues/2525

Fixes some issues with the admin UI generating incorrect titles and indices for added applications

### DIFF
--- a/app/views/admin/users/_form.html.erb
+++ b/app/views/admin/users/_form.html.erb
@@ -24,6 +24,26 @@
         </a>
       </li>
     `);
+
+    const applicationOnChange = () => {
+      const enabled = $('#add-application-select').val() && $(`#application-users-list input[type=\"hidden\"][name*=\"application_id\"][value=\"${$('#add-application-select').val()}\"]`).length === 0;
+
+      $('#add-application').prop('disabled', !enabled);
+    };
+
+    const applicationOnClick = (event) => {
+      event.preventDefault();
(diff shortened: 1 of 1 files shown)


app/views/admin/users/_form.html.erb
```diff
@@ -24,6 +24,26 @@
         </a>
       </li>
     `);
+
+    const applicationOnChange = () => {
+      const enabled = $('#add-application-select').val() && $(`#application-users-list input[type=\"hidden\"][name*=\"application_id\"][value=\"${$('#add-application-select').val()}\"]`).length === 0;
```
Dantemss, line 29: Don't let them add the same application twice.

app/views/admin/users/_form.html.erb
```diff
@@ -24,6 +24,26 @@
...
+
+      $('#add-application').prop('disabled', !enabled);
+    };
+
+    const applicationOnClick = (event) => {
+      event.preventDefault();
+      const nextIndex = Math.max(-1, ...Object.values($('#application-users-list input')).map(i => i.name?.match(/^application_users\[(\d+)\]/)).filter(m => m).map(m => parseInt(m[1], 10))) + 1;
```
Dantemss, line 36: I was counting the number of applications before but in case they repeatedly add and delete applications it's better to take the max index + 1. Gaps seem to be fine.

app/views/admin/users/_form.html.erb
```diff
@@ -24,6 +24,26 @@
...
+      renderApplicationUser(
+        $('#add-application-select').val(),
+        $('#add-application-select option:selected').text(),
```
Dantemss, line 39: This selector was wrong before.

TomWoodward (approved):

PR #1207, opened 2023-10-13
add adopter status to user as string

mwvolo:


Dantemss (commented):
I'd say approved but I think update_adopter_status may cause OOM in production. Do we have to load the contacts in chunks? UpdateUserContactInfo only loads contacts modified in the last day for example.

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -0,0 +1,36 @@
...
+        user_by_salesforce_id = User.where(
+          salesforce_contact_id: contacts.map(&:id)
+        ).index_by(&:salesforce_contact_id)
+
+        updated_contacts = contacts.map do |contact|
+          user = user_by_salesforce_id[contact.id]
+          user.adopter_status = contact.adoption_status
+          user.save!
+        end
+      rescue StandardError => se
```
Dantemss, line 29: If this rescues an error, you'll skip up to 250 contact (the entire batch) which is not great. May be less of a problem if we make this resumable. See below.

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -0,0 +1,36 @@
...
+      begin
+        user_by_salesforce_id = User.where(
+          salesforce_contact_id: contacts.map(&:id)
+        ).index_by(&:salesforce_contact_id)
+
+        updated_contacts = contacts.map do |contact|
+          user = user_by_salesforce_id[contact.id]
+          user.adopter_status = contact.adoption_status
+          user.save!
+        end
```
Dantemss, line 28: I'd wrap each batch inside a separate transaction. I think that will improve performance (rather than using autocommit to automatically create a transaction for each save!).

Dantemss (commented):
If this task is just for the initial migration and since the change is mainly happening to the local data, I'd reverse it: load 250 users from the local DB that have `adopter_status: nil` first, then look up their contacts only. This way if you need to stop and resume the process it'll look only at the users that are still missing their adopter_status.

lib/tasks/accounts/update_adopter_status.rake
mwvolo: I think this is how it's done but let me know if this doesn't look like the correct way to do this.

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -0,0 +1,36 @@
...
+      last_id = users.last&.id
+
+      begin
+        contacts = OpenStax::Salesforce::Remote::Contact.select(
+          :id,
+          :adoption_status,
+          :accounts_uuid
+        )
+        .where(id: users.map(&:salesforce_contact_id))
+        .index_by(&:id)
```
mwvolo, line 18: I'm a little fuzzy on if I did this correctly, will this work properly with line 22 to map the correct contact to the user?

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -4,33 +4,33 @@ namespace :accounts do
...
-        :adoption_status,
-        :accounts_uuid
-      )
-      .where("Accounts_UUID__c != null")
-      .order(:Id)
-      .limit(250)
-      contacts = contacts.where("Id > '#{last_id}'") unless last_id.nil?
-      contacts = contacts.to_a
-      last_id = contacts.last&.id
+      users = User.where(adopter_status: nil).order(:id).limit(250)
```
Dantemss, line 7: I would say `User.where(adopter_status: nil).where.not(salesforce_contact_id: nil).limit(250)` here (order(:id) is also fine but optional in this case).

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -4,33 +4,33 @@ namespace :accounts do
...
+      users = users.where("id > #{last_id}") unless last_id.nil?
```
Dantemss, line 8: With the above change you shouldn't need this anymore (or to track last_id). You just find 250 of them and update so adopter_status is not nil and they won't be found in later iterations.

lib/tasks/accounts/update_adopter_status.rake
Dantemss: You need to remove `.to_a` in line 19 below if you want a hashmap.

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -4,33 +4,33 @@ namespace :accounts do
...
-        updated_contacts = contacts.map do |contact|
-          user = user_by_salesforce_id[contact.id]
+        updated_users = users.map do |user|
+          contact = contacts[user.salesforce_contact_id]
           user.adopter_status = contact.adoption_status
-          user.save!
+        end
+        
+        updated_users.transaction do
+          updated_users.save!
```
Dantemss, line 27: I think you need `updated_users.each(&:save!)`

lib/tasks/accounts/update_adopter_status.rake
Dantemss: I think you can remove the rescue now. Just let it fail if something unexpected happens. Otherwise it can get stuck in an infinite loop.

If it fails, you can always retry.

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -0,0 +1,30 @@
...
+          contact = contacts[user.salesforce_contact_id]
+          user.adopter_status = contact.adoption_status
+        end
+
+        updated_users.transaction do
+          updated_users.each(&:save!)
+        end
+      rescue StandardError => se
+        Sentry.capture_exception se
+      end
```
Dantemss, line 27: You still need to break out of the loop when there are less than 250 users.

Dantemss (changes requested):
Just 2 more things

lib/tasks/accounts/update_adopter_status.rake
mwvolo: I was wondering why this was running for so long locally! Fixed

lib/tasks/accounts/update_adopter_status.rake
```diff
@@ -0,0 +1,28 @@
...
+    loop do
+      users = User.where(adopter_status: nil).where.not(salesforce_contact_id: nil ).limit(250)
+
+      contacts = OpenStax::Salesforce::Remote::Contact.select(
+        :id,
+        :adoption_status,
+        :accounts_uuid
+      )
+      .where(id: users.map(&:salesforce_contact_id))
+      .index_by(&:id)
```
mwvolo, line 14: @Dantemss it looks like `index_by` is not supported by ActiveForce. I'll see what I can do to fix this.

Dantemss (approved):

PR #1209, opened 2023-10-30
change subject to subject interest on lead creation, fix lead spec

mwvolo:
This change is to reflect the work to unify the way we collect subjects of interest in Salesforce. 
Currently, we have `subject` and `subject_interest`, with `subject_interest` being the more acceptable version of this field because of the way it's configured in Salesforce.
The web is making a similar change, which can be viewed on [this card.](https://github.com/orgs/openstax/projects/14/views/9?filterQuery=&pane=issue&itemId=41667551)

This also readds the lead creation spec and cassette, which had been xit'd out.


Dantemss (approved):
Yeah I'll just trust you on this one :)

PR #1212, opened 2023-11-17
Show privacy notice before redirecting if not signed

mwvolo:
Check if the user has acknowledged a new privacy policy before redirecting. If not, show them the terms to agree to and send them on their way (to profile if not using a redirect param, otherwise, to the app that originated the request).

### DIFF
--- a/app/controllers/newflow/login_controller.rb
+++ b/app/controllers/newflow/login_controller.rb
@@ -12,6 +12,7 @@ module Newflow
     before_action :known_signup_role_redirect, only: :login_form
     before_action :cache_alternate_signup_url, only: :login_form
     before_action :redirect_to_signup_if_go_param_present, only: :login_form
+    before_action :did_sign_privacy_notice, only: :login_form
     before_action :redirect_back, if: -> { signed_in? }, only: :login_form
 
     def login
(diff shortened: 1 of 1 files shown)


Dantemss:
Do we also need to make them sign the TOS?

mwvolo:
Because we have both Assignable and Accounts TOS in Accounts, we only want to force a signature on the Privacy Notice if it changes (prevents confusion if the Assignable TOS changes; we don't want all users to have to agree to that).

Dantemss (approved):

Dantemss:
I meant the Accounts TOS since Assignable users don't go through this (I think) unless they're adding login credentials to their account which is still being worked on.

Dantemss:
Maybe you could use fine_print_require but honestly idk how well that feature is working nowadays.

Dantemss:
Or rather remove `fine_print_skip` for this action.

mwvolo:
I tried to remove `fine_print_skip` for the login view but it didn't work well because it catches on if the user is not signed in yet, so wants them to have signed the contracts before even signing up. Is there a way to do something like:
```
fine_print_skip :general_terms_of_use, :privacy_policy, except: :profile_newflow, :login, if: -> { signed_in? }
```

mwvolo:
In reality, though, the request I got was only to make sure the privacy policy was visible. I think I'd like to keep the scope to that to avoid any unnecessary complications for other teams using accounts.

Dantemss:
What you have is probably fine then

mwvolo:
@Dantemss I made the terms views use the new layout. Can you make sure nothing looks crazy about this?
It looks good locally.
https://github.com/openstax/accounts/pull/1212/commits/ccb3fd3dc81eac345b5d172bcd14025ebaaa07f1

PR #1217, opened 2023-12-07
return to signup if privacy policy not signed, then back to r param

mwvolo:
The privacy notice was only shown on the redirect if the user was already signed in. This interrupts the redirect if they haven't signed it, asks them to agree to the changes, then continues them on to the parameter specified in `r`.

This caused a couple tests to start failing but I was able to test them locally successfully and they worked. They would each pass but not all together (probably something happening with the user) but I wasn't able to figure it out, so I kept one spec and commented out the other two.

### DIFF
--- a/app/controllers/newflow/login_controller.rb
+++ b/app/controllers/newflow/login_controller.rb
@@ -34,7 +34,7 @@ module Newflow
           sign_in!(user, security_log_data: {'email': @handler_result.outputs.email})
 
           if current_user.student? || !current_user.is_newflow? || (edu_newflow_activated? && decorated_user.can_do?('redirect_back_upon_login'))
-            redirect_back(fallback_location: profile_newflow_path) # back to `r`edirect parameter.
+            did_user_sign_recent_privacy_notice? ? redirect_back : redirect_to_sign_privacy_notice
           else
             redirect_to(decorated_user.next_step)
           end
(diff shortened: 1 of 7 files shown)


app/controllers/newflow/login_controller.rb
```diff
@@ -3,17 +3,14 @@ class LoginController < BaseController
...
-
     fine_print_skip :general_terms_of_use, :privacy_policy, except: :profile_newflow
 
     before_action :cache_client_app, only: :login_form
     before_action :known_signup_role_redirect, only: :login_form
     before_action :cache_alternate_signup_url, only: :login_form
     before_action :redirect_to_signup_if_go_param_present, only: :login_form
-    before_action :did_sign_privacy_notice, if: -> { signed_in? }, only: :login_form
-    before_action :redirect_back, if: -> { signed_in? }, only: :login_form
+    before_action :redirect_to_sign_privacy_notice, if: -> { signed_in? && !did_user_sign_recent_privacy_notice? }, only: :login
```
Dantemss, line 12: Shouldn't this be `login_form`? After this change there's no redirect for `login_form` if they are signed in but did not sign the notice

mwvolo: It seemed to work for me with `:login`, wasn't sure what the difference is but both seem to work.

Does this resolve your second point about being no redirect for `login_form` if signed in but didn't sign?

Dantemss: does this still work as you expect? Otherwise you may need both

mwvolo: I see what you mean, yes, I do need both of them for it to work correctly.

Dantemss (approved):

PR #1218, opened 2024-01-04
fix self-reported school name if school not found locally

mwvolo:


mwvolo:
I tested on dev, and the self-reported school is now going to the lead as expected.

Dantemss (approved):

PR #1223, opened 2024-01-18
Check profile complete and SheerID response in setting lead FV status

mwvolo:
The command to create the lead was overwriting the faculty status set by SheerID. Checking to make sure it's not one of the 3 expected values to come from SheerID before changing to incomplete_signup.

### DIFF
--- a/app/routines/newflow/create_or_update_salesforce_lead.rb
+++ b/app/routines/newflow/create_or_update_salesforce_lead.rb
@@ -46,7 +46,7 @@ module Newflow
       end
 
       # The user has not finished signing up, check for confirmed faculty if they were verified by sheerid
-      if !user.is_profile_complete? && !user.faculty_status.include(:confirmed_faculty, :pending_sheerid, :rejected_by_sheerid)
+      if !user.is_profile_complete? && !user.faculty_status.include?(:confirmed_faculty, :pending_sheerid, :rejected_by_sheerid)
         user.faculty_status = :incomplete_signup
       end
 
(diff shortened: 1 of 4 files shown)


app/routines/newflow/create_or_update_salesforce_lead.rb
```diff
@@ -45,8 +45,10 @@ def exec(user:)
         adoption_json = build_book_adoption_json_for_salesforce(user)
       end
 
-      # The user has not finished signing up
-      user.faculty_status = User::INCOMPLETE_SIGNUP unless user.is_profile_complete?
+      # The user has not finished signing up, check for confirmed faculty if they were verified by sheerid
+      if !user.is_profile_complete? && !user.faculty_status == :confirmed_faculty
```
mwvolo, line 49: I was trying to do something like 
```
user.faculty_status.include?(:confirmed_faculty, :pending_sheerid, :rejected_by_sheerid)
```
but clearly that's not the correct ruby syntax, any advise on checking the state within an array? I'm sure I'm just doing something backwards

Dantemss: `faculty_status` is a Rails enum which is slightly weird. You can assign symbols or strings but the method returns a string if I remember right. I think either of these should work:

`user.confirmed_faculty? || user.pending_sheerid? || user.rejected_by_sheerid?`

`['confirmed_faculty', 'pending_sheerid', 'rejected_by_sheerid'].include?(user.faculty_status)`

mwvolo:
@Dantemss found a different way that I think will work best to determine the faculty status we want to get to Salesforce - thank you for the feedback on the includes statement though!

When you get a moment, can you give this a review?

Dantemss (approved):

PR #1225, opened 2024-01-22
Update openstax_salesforce to pull in General Account types from SF

mwvolo:
We have users in Accounts assigned to neither a School or School District that are confirmed as faculty (Tutoring centers, Government organizations, Vocational schools, etc.).
This will allow those users to properly sync with Salesforce, which is currently causing an error.


Dantemss (approved):
Might be a good idea to try to test that they don't lose their status when synced

PR #1231, opened 2024-01-24
Fix redirects for users in signup process

mwvolo:
Fix redirects for users that have not completed signing up.
This should handle the new faculty states (incomplete signup, pending sheerid, and rejected by sheerid).

Also got some old feature specs cleaned up. 📈


app/controllers/newflow/educator_signup_controller.rb
```diff
@@ -179,14 +179,12 @@ def store_if_sheerid_is_unviable_for_user
...
 
     def exit_signup_if_steps_complete
-      return if !current_user.is_newflow?
-
       case true
-      when current_user.is_educator_pending_cs_verification && current_user.pending_faculty?
+      when current_user.is_educator_pending_cs_verification? && (!current_user.confirmed_faculty? || !current_user.rejected_faculty?)
         redirect_to(educator_pending_cs_verification_path)
-      when current_user.is_educator_pending_cs_verification && !current_user.pending_faculty?
+      when (current_user.confirmed_faculty? || current_user.rejected_faculty?) && current_user.is_profile_complete?
```
Dantemss, line 185: This raises questions in my mind about what happens if the user is confirmed or rejected and the profile is not complete 😂

Dantemss (approved):
I'm mostly trusting you, I don't think I could reason about this logic without actively testing Accounts.

app/controllers/newflow/educator_signup_controller.rb
mwvolo: I was trying to merge these two checks, and both of them fundamentally do the same thing... I didn't want to start a cascade of changes through the application based on this, but I might give it one more go (in another PR).

If someone is rejected and non-profile complete, they'd fall into the last statement here and fill out a profile. It wouldn't do them any good, as they are rejected, but also doesn't really harm anything for this pretty small corner case.

PR #1233, opened 2024-01-25
bump sentry, add cron job monitoring

mwvolo:
Sentry has new capabilities to [monitor crons ](https://docs.sentry.io/platforms/ruby/crons/#:~:text=Sentry%20Crons%20allows%20you%20to,prevent%20disruptions%20to%20your%20service.)and other jobs, this upgrades the gem, adds the [delayed-job sdk](https://docs.sentry.io/platforms/ruby/guides/delayed_job/) for Sentry, and sets parameters in the crons we have setup to monitor that they started and stopped as expected.

This also takes advantage of [setting a user id](https://docs.sentry.io/platforms/ruby/enriching-events/identify-user/?original_referrer=https%3A%2F%2Fwww.google.com%2F) to track which user is having an issue when an error report comes in (using `user.id`, so nothing identifiable),

This also upgrades puma, as mentioned in [this dependabot PR](https://github.com/openstax/accounts/pull/1219).

### DIFF
--- a/config/initializers/rescue_from.rb
+++ b/config/initializers/rescue_from.rb
@@ -4,48 +4,24 @@ secrets = Rails.application.secrets
 exception_secrets = secrets.exception
 OpenStax::RescueFrom.configure do |config|
   # Show the default Rails exception debugging page on dev
-  config.raise_exceptions = EnvUtilities.load_boolean(name: 'RAISE',
-                                                      default: Rails.env.development?)
+  config.raise_exceptions = EnvUtilities.load_boolean(name: 'RAISE', default: Rails.env.development?)
 
   config.app_name = 'Accounts'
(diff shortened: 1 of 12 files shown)


mwvolo:
verified on dev

config/initializers/rescue_from.rb
```diff
@@ -34,7 +34,7 @@
     }
     extra.merge!(proxy.extras) if proxy.extras.is_a? Hash
 
-    Sentry.capture_exception(proxy.exception, extra: extra)
+    Sentry.capture_exception(proxy.exception)
```
Dantemss, line 37: I think `extra` here and above is unused now so it could be removed

Dantemss (approved):

PR #1236, opened 2024-01-31
School info in admin, Sentry message on school error during lead creation

mwvolo:
We have a few users that are signing up and getting associated with a school that has likely been merged in Salesforce, or somehow had it's salesforce_id updated. Because there are users associated with it, it doesn't delete when updating schools from Salesforce.
This code will throw a Sentry message for more investigation into what's going on.

I've also added some administrative abilities to see the school associated with the user.

### DIFF
--- a/app/routines/newflow/create_or_update_salesforce_lead.rb
+++ b/app/routines/newflow/create_or_update_salesforce_lead.rb
@@ -115,7 +115,9 @@ module Newflow
       try
         save_lead(user, lead)
       except Restforce::ErrorCode::InsufficientAccessOnCrossReferenceEntity
+          Sentry.capture_message("Invalid school (#{user.school.salesforce_id}) for user (#{user.id})")
           UpdateSchoolSalesforceInfo.cleanup_merged_schools(user.school)
+          save_lead(user, lead)
 
       outputs.lead = lead
       outputs.user = user
--- a/app/routines/newflow/create_or_update_salesforce_lead.rb
+++ b/app/routines/newflow/create_or_update_salesforce_lead.rb
@@ -112,14 +112,24 @@ module Newflow
         event_data: { lead_data: lead }
       )
 
-      if lead.save
+      try
+        save_lead(user, lead)
+      except Restforce::ErrorCode::InsufficientAccessOnCrossReferenceEntity
+          UpdateSchoolSalesforceInfo.cleanup_merged_schools(user.school)
+
+      outputs.lead = lead
+      outputs.user = user
+    end
+
+    def save_lead(user, lead)
         user.salesforce_lead_id = lead.id
         if user.save
           SecurityLog.create!(
-            user: user,
+            user:       user,
             event_type: :created_salesforce_lead,
             event_data: { lead_id: lead.id.to_s }
           )
+          user
         else
           SecurityLog.create!(
             user: user,
@@ -130,23 +140,8 @@ module Newflow
             }
           )
           Sentry.capture_message("User #{user.id} was not successfully saved with lead #{lead.id}")
+          raise
         end
-        outputs.lead = lead
-      else
-        message = "#{self.class.name} error creating SF lead! #{lead.inspect}; User: #{user.id}; Error: #{lead.errors.full_messages}"
-
-        Sentry.capture_message(message)
-
-        SecurityLog.create!(
-          user: user,
-          event_type: :salesforce_error,
-          event_data: {
-            message: message
-          }
-        )
-      end
-
-      outputs.user = user
     end
 
     def build_book_adoption_json_for_salesforce(user)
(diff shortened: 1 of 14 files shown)


app/routines/update_school_salesforce_info.rb
```diff
@@ -80,6 +80,30 @@ def call
...
+        Sentry.capture_message("Possible merged school not found. Original: #{merged_school.salesforce_id} New: #{merged_school.updated_salesforce_id}")
+      else
+        updated_school = School.find_or_create_by(salesforce_id: sf_school.id)
+        SF_TO_DB_CACHE_COLUMNS_MAP.each do |sf_column, db_column|
+          updated_school.public_send "#{db_column}=", sf_school.public_send(sf_column)
+        end
+        updated_school.save!
+
+        merged_school.users.update_all(school: updated_school) if merged_school.users.any?
+        merged_school.destroy!
```
Dantemss, line 103: I'd call reload before destroy!, just in case... otherwise it could try to dissociate the users.

app/routines/update_user_contact_info.rb
```diff
@@ -121,6 +121,8 @@ def call
       user.grant_tutor_access = sf_contact.grant_tutor_access
 
       if school.nil? && !sf_school.nil?
+        # Add the possible updated school id to the record, we check for this in UpdateSchoolSalesforceInfo
+        user.school.updated_salesforce_id = sf_school.id
```
Dantemss, line 125: user.school is maybe not always set?

Dantemss: Also this code would only be relevant when the school is in SF but not in Accounts, which hopefully only happens sometimes in the first sync after CS adds a school in SF.

Dantemss (changes requested):
I'm thinking instead of this entire PR all you need is maybe another check in update_user_contact_info.rb for this additional case:
`if !school.nil? && !sf_school.nil?`
In this case you want to check that school.salesforce_id == sf_school.id and if not, then update the user's school?

Dantemss:
Nvm, what you actually what is compare `user.school` with `school` somewhere. `school` in update_user_contact_info.rb is being loaded from Salesforce (via the school id in the contact and the salesforce_id in School).

Dantemss:
Alternatively, change the `user.school` association (and `school.users`) to always via the salesforce keys and remove/ignore the `user.school_id` key in Accounts. You can set the `primary_key` and `foreign_key` options in the `belongs_to` and `has_many` associations to accomplish this. You'd have to check nothing else is using `user.school_id` directly though.

mwvolo:
You certainly helped me see that I was on the wrong path in trying to fix the error. The error is popping up when the lead gets created first, which is where it should be caught - then update the users affected so it doesn't impact future syncs with user/school combo.

See if you think I still need to modify the foreign key to accomplish what I am trying to do. There might be another place in the application that I should check if the school is 'stale'.

mwvolo:
@Dantemss decided to just throw a message to Sentry for now, this also adds some admin info that will be helpful.
I'll address cleaning up stale schools in another PR.

app/routines/newflow/create_or_update_salesforce_lead.rb
```diff
@@ -121,31 +121,19 @@ def exec(user:)
             event_data: { lead_id: lead.id.to_s }
           )
         else
+          if lead.errors.messages.inspect.include? == 'INSUFFICIENT_ACCESS_ON_CROSS_REFERENCE_ENTITY'
+            Sentry.capture_message("Invalid school (#{user.school.salesforce_id}) for user (#{user.id})")
```
Dantemss, line 125: It's not clear to me that's what the message means but if you are sure from what you've seen in SF go for it.

Dantemss (commented):
I think we can still have the case where `sf_contact.school_id != user.school.salesforce_id`. If user.school is supposed to reflect what's in Salesforce, I think we just need to set `user.school` in `app/routines/update_user_contact_info.rb` to match whatever is in the contact? Basically where it says `school =` just change to `user.school =`

Dantemss:
Nvm, I see this line https://github.com/openstax/accounts/blob/main/app/routines/update_user_contact_info.rb#L132 is supposed to do that

Dantemss (approved):

app/routines/newflow/create_or_update_salesforce_lead.rb
mwvolo: This took a bit of command line testing to figure out what SF is spitting out as an error - but yeah, this is the one's that need catching with the ID of that school so it can at least be manually cleaned up (maybe this can move to the admin or some automated process in the future).

mwvolo:
> Nvm, I see this line https://github.com/openstax/accounts/blob/main/app/routines/update_user_contact_info.rb#L132 is supposed to do that

Ah, this might be where I need to put something to update all the users with that old school id!
Again, future PR. Thanks for all the pointers!

PR #1253, opened 2024-08-21
DISO-450 Recreate the doorkeeper token when the user is updated

Dantemss:
Need for the Assignable consent stuff because we read the consent by decoding the SSO token/cookie


mwvolo (approved):

Dantemss:
We'll need this on staging before for tomorrow's Assignable staging deployment

PR #1254, opened 2024-08-22
send tracking parameters with lead creation

mwvolo:
1 of 2 PRs (maybe) to get tracking parameters set for leads in salesforce.
This helps us track where adoptions are coming from.

2nd PR would be to include the full url from the user landing on accounts, which will give us more detail on where account signups are occuring.


Dantemss (approved):
Seems fine though the url won't really be correct for dev/staging

PR #1255, opened 2024-08-23
re-add support_identifier

mwvolo:
and this is why we need qa....

seems this [broke](https://openstax.sentry.io/issues/5108908364/?environment=production&project=5772471&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=30d&stream_index=6) a few things, so adding it back in

[not sure if this is also related](https://openstax.sentry.io/issues/5593702379/?environment=production&project=5772471&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=30d&stream_index=2)

### DIFF
--- a/app/representers/api/v1/user_representer.rb
+++ b/app/representers/api/v1/user_representer.rb
@@ -50,6 +50,11 @@ module Api::V1
              readable: true,
              writeable: false
 
+    property :support_identifier,
+             type: String,
+             readable: true,
+             writeable: false
+
     property :consent_preferences,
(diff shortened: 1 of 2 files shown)


Dantemss (approved):
What did it break, for future reference?
Also, the linked error seems to indicate something is calling `.trusted?()` with nil as an argument?

PR #1256, opened 2024-08-26
limit student count to 2,000

mwvolo:
To match our existing adoption form, and due to a recent error in class size reporting, this will limit the count of students to 1,000 per book they select.


mwvolo:
Hold on merging this until we have [Anthony's decision](https://openstax.slack.com/archives/C6LPT04BV/p1724708757036149?thread_ts=1724707280.190649&cid=C6LPT04BV).

Dantemss (approved):

mwvolo:
Because we have noticed some legitimate student reports over 1k, moving to 2k.

PR #1259, opened 2024-09-26
CORE-484 Ensure access tokens returned to Assignable have a minimum duration of 12 hours

Dantemss:
To prevent random expired session errors.


Dantemss:
Side note, I can no longer run Accounts locally due to OS X updates and libv8. Maybe it's time to upgrade to at least Ruby 3.0.

mwvolo (approved):

PR #1261, opened 2024-10-07
Add a way to see if a user is an assignable adopter and if school is integrated

mwvolo:
For [DATA-133](https://openstax.atlassian.net/browse/DATA-133?atlOrigin=eyJpIjoiNjc3ODI4ODJkMDMxNDdiYTkwZjgxMzJjODQ4YTEyY2QiLCJwIjoiaiJ9)

[DATA-133]: https://openstax.atlassian.net/browse/DATA-133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mwvolo:
Dante, if this is going to mess with your work to update to Ruby 3/Rails 6, it's fine to wait - not a rush.

app/routines/newflow/create_or_update_salesforce_lead.rb
```diff
@@ -61,9 +61,9 @@ def exec(user:)
 
 
       if user.salesforce_lead_id
-        lead = OpenStax::Salesforce::Remote::Lead.find_by(email: user.best_email_address_for_salesforce)
+        lead = OpenStax::Salesforce::Remote::Lead.find_by(email: "#{user.best_email_address_for_salesforce}")
```
mwvolo, line 64: @Dantemss do you think this will help with [this issue](https://openstax.sentry.io/issues/5414102669/?environment=production&project=5772471&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=3)? Would `to_s` achieve the same?
Seems to be exploding the the email address has a `'` in it.

Dantemss: `best_email_address_for_salesforce` already returns a string or nil, so I'm not sure that would help here.

app/representers/api/v1/user_representer.rb
```diff
@@ -193,6 +193,14 @@ class UserRepresenter < Roar::Decorator
                if: ->(user_options:, **) { user_options.try(:fetch, :include_private_data, false) },
                getter: ->(represented:, **) { represented.external_ids.map(&:external_id) }
 
+    property :assignable_school_integrated,
+               if: ->(user_options:, **) { user_options.try(:fetch, :include_private_data, false) },
+               getter: ->(*) { self.school.has_assignable_contacts unless self.school.nil?}
```
Dantemss, line 198: You can probably do `self.school&.has_assignable_contacts`

Dantemss: I worry about adding these things to this representer though, as the representer is used to make the SSO cookie, which runs into size issues often...

app/routines/update_user_contact_info.rb
```diff
@@ -123,6 +123,8 @@ def call
       if school.nil? && !sf_school.nil?
         users_without_cached_school += 1
         Sentry.capture_message("User #{user.id} has a school that is in SF but not cached yet #{sf_school.id}")
+      elsif sf_contact.school_id != user.school.salesforce_id
+        Sentry.capture_message("User #{user.id} has a mismatched school id (Salesforce: #{sf_contact.school_id}, Accounts: #{user.school.salesforce_id})")
```
Dantemss, line 127: The previous behavior was `user.school = school` which is in the `else` below. We should probably still do this right? The other fields like `user.school_type` and `user.school_location` were already updated above.

mwvolo: This is here to identify a problem that is happening when schools are merged in Salesforce.
I went through and fixed a bunch of them, and I'm not sure how frequent it happens (so logging for now).
The sync code [deletes the school properly](https://github.com/openstax/accounts/blob/main/app/routines/update_school_salesforce_info.rb#L29-L41) if it has no users, but if there are users and duplicate schools, it creates both - and the users are never moved from the merged school (hope that makes sense).

app/representers/api/v1/user_representer.rb
mwvolo: Heck... most of this stuff probably isn't needed in the SSO cookie...

app/routines/update_user_contact_info.rb
Dantemss: What I'm saying is... because this overrides the stuff in the `else`, the user's school will not get updated here but some school-related fields were updated above so everything will be half-updated.

mwvolo: Ah, gotcha. Thanks for pointing that out. 
See what you think of the change I made, I think it'll fix the broken records (and log for now, because I'm interested in how frequently this is happening)

app/routines/newflow/create_or_update_salesforce_lead.rb
mwvolo: reverted this, sounds like we need to look into fixes elsewhere
[for reference, slack message](https://openstax.slack.com/archives/C0R8GMF1S/p1729528884970799)

app/routines/update_user_contact_info.rb
```diff
@@ -101,6 +101,13 @@ def call
                            :other_school_type
                          end
 
+      if school != user.school.salesforce_id
```
Dantemss, line 104: This may blow up if `user.school` is nil. If it doesn't blow up, it will always trigger because school is an ActiveRecord object and salesforce_id is a string but see below first

app/routines/update_user_contact_info.rb
```diff
@@ -101,6 +101,13 @@ def call
...
+        user.school = School.find_by(salesforce_id: user.school.salesforce_id)
```
Dantemss, line 105: This is a no-op but it hits the database, which is not great.
`user.school` should already return the thing you are setting it to.
I'm thinking what was here before (the `else` part of the if statement) would already fix mismatched schools. You could put a check inside the `else` to log if a mismatch happened (like actually inside the `else`, not `elsif`).

It won't fix the mismatch if `school.nil?` which will get logged by the other thing.
This case can maybe be greatly reduced by ensuring this routine always runs right after the routine that syncs Schools from Salesforce.

Dantemss (changes requested):

Dantemss:
If you are getting too many users with mismatched schools for 1 sync cycle, maybe the thing to do is combine the 2 half-hour cron tasks, ensuring the school sync always runs before the user sync.

Another option would be changing the user -> school relation to use the salesforce_id instead of the Rails id (probably make a new column in User for this if you do that). That way we wouldn't care about the state of the school sync when updating the users.

Gemfile.lock
```diff
@@ -514,7 +517,7 @@ GEM
     omniauth-twitter (1.4.0)
       omniauth-oauth (~> 1.1)
       rack
-    openstax_active_force (1.1.0)
+    openstax_active_force (1.1.1)
```
mwvolo, line 520: @Dantemss I went ahead and upgraded this as part of this PR.
This is all that's needed, right? No update to `openstax_salesforce` required?

app/routines/update_user_contact_info.rb
mwvolo: I see what you are saying now - it took a little walkthrough of the code. I'm going to take this out - it sounds like I need to rethink the problem we saw (which might have been from a legacy process).
I'll investigate more.

mwvolo:
I moved the school sync to the same cron as the contact, so they run in order. Not sure if the 10-past file is still needed, or good to have around for other jobs if it's part of a deployment change.

app/representers/api/v1/user_representer.rb
mwvolo: All the inline comments on the files got deleted from the diff when I updated the branch from main... but I'm pretty sure this is the last remaining issue. 
How do I keep this in the API but not bloat the cookie?

db/migrate/20241007155255_add_assignable_flag_to_school.rb
```diff
@@ -0,0 +1,5 @@
+class AddAssignableFlagToSchool < ActiveRecord::Migration[5.2]
+  def change
+    add_column :schools, :has_assignable_contacts, :boolean
```
Dantemss, line 3: I feel like this table is small enough that you could probably set `null: false, default: false`

lib/tasks/cron/5-past-half-hour.rake
```diff
@@ -2,7 +2,8 @@ namespace :cron do
   task '5-past-half-hour': :log_to_stdout do
     Rails.logger.debug 'Starting 5-past-half-hour cron'
 
-    Rails.logger.info "UpdateUserContactInfo.call"
+    Rails.logger.info "Starting Salesforce Contact and School Sync"
+    OpenStax::RescueFrom.this { UpdateSchoolSalesforceInfo.call }
     OpenStax::RescueFrom.this { UpdateUserContactInfo.call }
```
Dantemss, line 7: It may be a good idea to check how long these take in prod actually, to make sure combined they don't exceed 30 min

Dantemss (approved):

lib/tasks/cron/5-past-half-hour.rake
mwvolo: Sometimes they do - does that mean they are going to start running again? I thought we had something that would prevent duplicate crons from starting if the previous wasn't complete.

mwvolo: Oh, you know, I think I fixed what was causing these to run so long. Looks like they are down to sub-minute runs now. Probably good.
https://openstax.sentry.io/crons/accounts/update-user-contact-info/?environment=production

db/migrate/20241007155255_add_assignable_flag_to_school.rb
Dantemss: I know I just said the above but now I'm thinking... is Salesforce guaranteed to return a boolean here? Just need to make sure it won't try setting this to nil.

mwvolo: If it's cast as a boolean [here](https://github.com/openstax/openstax_salesforce/blob/main/lib/openstax/salesforce/remote/school.rb#L19), shouldn't this always return true/false?

lib/tasks/cron/5-past-half-hour.rake
mwvolo: yeah, and it looks like it's handling whatever salesforce throws at it through active_force https://github.com/openstax/active_force/blob/3210375b8ea95b576c876de87f1e02fcdc0139d0/lib/active_force/mapping.rb#L63

db/migrate/20241007155255_add_assignable_flag_to_school.rb
Dantemss: No idea 😂 have to check active_force...
Seems like it always returns a boolean: https://github.com/openstax/active_force/blob/3210375b8ea95b576c876de87f1e02fcdc0139d0/lib/active_force/mapping.rb#L64

mwvolo: LOL, I looked at the same thing.. put the comment on the wrong thread 🤦🏼 
https://github.com/openstax/accounts/pull/1261#discussion_r1815361600

PR #1262, opened 2024-10-08
allow user deletion from admin

mwvolo:
To incorporate a better customer support process and have mechanisms to comply with privacy requirements. 
This adds a button to the edit user page, which allows a user account to be deleted.


Dantemss:
I'm thinking this will fail to delete any activated users, since it delegates to https://github.com/openstax/accounts/blob/main/app/routines/destroy_user.rb

You also need to think what you want to do with all the records that have foreign keys to user, like external_id.

It may be easier to just erase the user name, password and contact infos instead (and username but that's old)?

Dantemss:
I think that's what GitHub does anyway. The reason why you become "ghost" when you are deleted here.

mwvolo:
Are you sure this is delegated to that routine? I'm not seeing the connection, but maybe it's some rails magic I don't understand. It looks like it's only using that routine on merge_unclaimed_user and transfer_authentications.

Emptying out the identifying info seems reasonable to me, if you think that is a better way. It's what CS is doing now, so building an automated way to do that is what I'm trying to accomplish.

mwvolo:
FWIW, it did work locally on an activated student and instructor account.

Dantemss:
You have to go through all the `has_many/has_one` in User and ensure all of them are handled.
`dependent: :destroy` will destroy the associated record and may be necessary for some validations
By default it may try to leave the record as-is or nullify the foreign key (not sure which). Need to make sure the foreign key constraint, if any, allows that. And also probably add `optional: true` to the `belongs_to :user` in the other record.

app/models/application_user.rb
```diff
@@ -1,7 +1,7 @@
 class ApplicationUser < ApplicationRecord
   belongs_to :application, class_name: 'Doorkeeper::Application',
                            inverse_of: :application_users
-  belongs_to :user, inverse_of: :application_users
```
Dantemss, line 4: This should probably not be optional, since the DB still requires it.

app/models/authentication.rb
```diff
@@ -1,5 +1,7 @@
+include UserSessionManagement
+
 class Authentication < ApplicationRecord
-  belongs_to :user, inverse_of: :authentications
+  belongs_to :user, inverse_of: :authentications, optional: true
```
Dantemss, line 4: Oddly enough, optional is correct here because the column is nullable in the DB.

app/models/authentication.rb
```diff
@@ -20,6 +22,8 @@ def display_name
   protected
 
   def check_not_last
+    return if user.is_deleted?
```
Dantemss, line 25: Given the above, you may want `user.nil? || user.is_deleted?` or something similar.

app/routines/soft_delete_user.rb
```diff
@@ -0,0 +1,33 @@
...
+    return if user.nil?
+
+    # Make sure object up to date, esp before dependent destroy stuff kicks in
+    user.reload
+
+    user.is_deleted = true
+    user.save!
+
+    user.external_ids.destroy_all
+    user.external_uuids.destroy_all
```
Dantemss, line 17: I didn't even realize we had both. Maybe `external_uuids` can be deleted. I think they just stored the UUID that Tutor used?

db/migrate/20241009175157_add_delete_flag_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeleteFlagToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :is_deleted, :boolean
```
Dantemss, line 3: I would make this `null: false, default: false` just to prevent having 3 possible values.

Dantemss: Actually ignore what I said because adding a column with a default rewrites the entire table and we have a ton of users.

app/routines/soft_delete_user.rb
mwvolo: I'm sure we could have an entire PR dedicated to remove tutor-specific features 😂

Dantemss (approved):

PR #1265, opened 2024-10-18
Playwright

Dantemss:
- Moved from Selenium to Playwright
- Updated GH actions OS versions
- Reorganized .gitignore and removed stuff that seems gone
- Updated README

If you have something local that should remain in .gitignore let me know


Dantemss:
Well from the test failure I see this didn't really work. Maybe we need is to try blocking GTM or other external scripts.

PR #1268, opened 2024-10-21
Fix error when calling the users API with no parameters

Dantemss:
I noticed this error when testing. Maybe it doesn't matter much in the end, but we might as well handle it.


mwvolo (approved):
Nice find!

mwvolo:
I popped into Sentry and realized I need to move some changes to the Salesforce sandbox... I'll clean those up.

PR #1269, opened 2024-10-23
More user API token creation fixes

Dantemss:
Turns out we should just be passing in the actual objects to these methods now.


Dantemss:
I went to check if we actually have tests and we do, guess the tests here were not too useful.

mwvolo (approved):

mwvolo:
makes sense why the `to_s`/`to_i` wasn't working :)

PR #1271, opened 2024-10-24
Fix asset precompile

mwvolo:
the [build is failing](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/624276253531/projects/accounts-ami/build/accounts-ami%3A207044c8-90ef-49de-95a4-c7a2e15592ed?region=us-east-1) and I think it's related to the blazer upgrade


mwvolo:
this did the trick, maybe we can move to latest (3.1.0) as we upgrade more
https://us-east-1.console.aws.amazon.com/codesuite/codebuild/624276253531/projects/accounts-ami/build/accounts-ami%3Afcc8f945-1c6d-4e54-961d-3f2bbdae2a1c?region=us-east-1

Dantemss (approved):
Uglifier is already the latest version so not sure what to do there to support blazer 3.

mwvolo:
blazer 3 works, it's blazer 3.1 that seems to blow up
i was thinking it might be more about the rails version we are using - maybe 7 is needed for 3.1

PR #1276, opened 2024-10-31
show spanish books and hide polish from cms api

mwvolo:
For https://openstax.atlassian.net/browse/DATA-147

### DIFF
--- a/lib/fetch_book_data.rb
+++ b/lib/fetch_book_data.rb
@@ -19,7 +19,6 @@ class FetchBookData
 
     items = results.fetch('items', [])
     # See https://github.com/openstax/openstax-cms/blob/main/books/constants.py#L7 for book states
-    # !['retired', 'unlisted'].include?(i['book_state'])
     books = items.reject{ %w[retired unlisted].include?(i['book_state']) }
 
     books_with_subject = []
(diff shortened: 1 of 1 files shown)


lib/fetch_book_data.rb
```diff
@@ -18,8 +18,8 @@ def fetch_titles
     return [] if results.blank?
 
     items = results.fetch('items', [])
-    # All books except for "retired" ones
-    books = items.select{ |i| i['book_state'] != 'retired' }
+    # See https://github.com/openstax/openstax-cms/blob/main/books/constants.py#L7 for book states
+    books = items.select{ |i| i['book_state'] != 'retired' || i['book_state'] != 'unlisted' }
```
Dantemss, line 22: Shouldn't it be `i['book_state'] != 'retired' && i['book_state'] != 'unlisted'`?

mwvolo: I'm trying to make sure it's not shown if it's retired or unlisted (two different states)... wouldn't that only exclude if they are both retired and unlisted, which can't happen .

Dantemss: You want `!(A == B || A == C)` which is equivalent to `(A != B && A != C)`
Or you can just use `!['retired', 'unlisted'].include?(i['book_state'])`
Or use `reject` instead of `select` to remove the negation.

Dantemss: `(A != B || A != C)` where `B != C` is just `true` all the time.

mwvolo: ah, thank you! that would have been confusing (as if I wasn't enough already)... 
`!['retired', 'unlisted'].include?(i['book_state'])` looks like an elegant solution, that's way more readable... I'll go with that

Dantemss (approved):

PR #1280, opened 2024-11-07
send language with lead creation

mwvolo:
If the book is foreign, we track that by appending the language to the book name in brackets - `Calculus Volume 1 [Spanish]`

This breaks that out from the book name and sends it with the adoptionJSON


Dantemss (approved):
Could probably do this in one gsub with 2 capture groups, but it's probably good enough.

PR #1284, opened 2024-11-20
update delayed_job

mwvolo:
A possible fix for emails sometimes going out and sometimes not.

From Dante:
```
this apparently doesn't work anymore: https://github.com/openstax/accounts/blob/main/config/initializers/active_job.rb#L2
Needs to be ActiveJob::Base.queue_adapter = :delayed_job instead
So all jobs are running async in the same thread
```

### DIFF
--- a/config/initializers/active_job.rb
+++ b/config/initializers/active_job.rb
@@ -1,2 +1,2 @@
 # Use delayed_job for background jobs
-ActiveJob::Base.queue_adapter = :delayed_job
+ActiveSupport.on_load(:active_job) { ActiveJob::Base.queue_adapter = :delayed_job }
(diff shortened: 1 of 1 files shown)


Dantemss (approved):

Dantemss:
The config doesn't work because ActiveJob is being loaded before the config is set.

Maybe the best way is actually something like:
`ActiveSupport.on_load(:active_job) { ActiveJob::Base.queue_adapter = :delayed_job }`

Since something else (some gem?) is already loading ActiveJob too early, I guess it doesn't matter much. But I might change to that eventually so we don't contribute to the problem by also loading ActiveJob::Base too early ourselves.

mwvolo:
are you saying that's what should be in the initializer file instead of this?

mwvolo:
like this? https://github.com/openstax/accounts/pull/1284/commits/708fa1e9c09632a4c3ce9f9b81073214a400ce67

Dantemss:
Yeah that looks good, to prevent the initializer itself from loading ActiveJob too early.

PR #1288, opened 2024-12-11
skip displaying the pin code text when creating external users

jivey:
https://openstax.atlassian.net/browse/CORE-542

### DIFF
--- a/app/routines/newflow/create_email_for_user.rb
+++ b/app/routines/newflow/create_email_for_user.rb
@@ -5,7 +5,7 @@ module Newflow
 
     protected ###############
 
-    def exec(email:, user:, is_school_issued: nil)
+    def exec(email:, user:, is_school_issued: nil, show_pin: nil)
       @email = EmailAddress.find_or_create_by(value: email&.downcase, user_id: user.id)
       @email.is_school_issued = is_school_issued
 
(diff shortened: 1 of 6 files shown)


mwvolo (approved):

app/routines/newflow/create_email_for_user.rb
```diff
@@ -5,7 +5,7 @@ class CreateEmailForUser
 
     protected ###############
 
-    def exec(email:, user:, is_school_issued: nil)
+    def exec(email:, user:, is_school_issued: nil, show_pin: nil)
```
Dantemss, line 8: Just to be clear, this will actually pass `nil` show_pin into the method below if not specified, which might be unexpected (but then I guess `nil != false` so it'll show the pin)

Dantemss (approved):

app/routines/newflow/create_email_for_user.rb
mwvolo: that's a good point - maybe it should default to true?

jivey: I don't recall why I thought I should leave it ambiguous here, I agree it's a lot clearer to just default to true; pushed that change.

PR #1290, opened 2025-01-02
Use released versions of lev, openstax_transaction_isolation, openstax_transaction_retry and fine_print

Dantemss:
There are still a few gems loaded from GitHub in the Gemfile, namely:
```
openstax/intl-tel-input-rails
openstax/path_prefixer
openstax/pattern-library
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -122,7 +122,7 @@ gem 'representable'
 gem 'keyword_search'
 
 # ToS/PP management
-gem 'fine_print', github: 'lml/fine_print', ref: '636023f68e95196dffaf295bfad3ad8051c23542'
+gem 'fine_print'
 
 # Send users back to the correct page after login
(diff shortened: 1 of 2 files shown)


mwvolo:
@Dantemss any ideas on why migrations are failing? Is there some kind of github actions cache in play here? 
Seems stage all the other tests pass.

Dantemss:
@mwvolo I noticed the cache action was running really slowly on that day, maybe related to GitHub deprecating cache v2 and/or Ubuntu 20... let's see if the retry works

Dantemss:
On second thought, what is happening is the migration task tries to checkout the gems that were there before this PR to compare and it's trying to checkout the lev revision that was there before and that is failing for whatever reason. I think we can ignore.

mwvolo (approved):

PR #1292, opened 2025-01-29
Fix TOS page jump to bottom. 

P-Gill97:
https://openstax.atlassian.net/browse/CORE-75


P-Gill97:
The issue is due to the browser autofocusing to the checkbox which causes the browser to scroll when the policy is longer than a page. I tried updating the checkbox line here `<%= f.check_box :i_agree, style: "height: 1.3rem"%>` by adding the attribute for `autofocus: false`. It did not work and the page kept jumping down to the bottom. So instead I added an `DOMContentLoaded` event listener `document.addEventListener("DOMContentLoaded", function() {` and set the window scroll to the top of the page using `window.scrollTo(0, 0);`.

Dantemss:
I am worried that this will cause the page to flash different positions when loading?
Also wondering if the second thing mentioned in the card (the button not being visually disabled) has been fixed already or if  it's still an issue.

mwvolo:
@P-Gill97 is this ready to be merged?

mwvolo:
maybe @Dantemss too, since you are on as reviewer 🙂

P-Gill97:
@mwvolo  I am conflicted on this because this is my first time working with Ruby templates and Coffee js. I was able to solve the issues but I don't know if I went down the right path to do it. Any insight on this?

Dantemss (approved):
Well, send it to testing. Ask QA to check if the terms flash the top section before scrolling and if it happens, if that is an issue.

PR #1289, opened 2024-12-23
update colors, add profile buttons, support chat with feature flag

mwvolo:
Hold merging until our Salesforce support chat is upgraded. 
Planned for March 2025.

Tested and working on dev - but may need to review Trusted Hosts and CORS settings in Salesforce as we move between environments.

Screenshot 2025-03-12 at 21 19 10
Screenshot of update profile page + chat with prefilled prechat form (authenticated only)

PR #1293, opened 2025-02-01
Posthog analytics

mwvolo:
To infer more usage without having to ask users (less reliance on self-reported adoption), we want to try [Posthog](https://posthog.com/). This product integrates with Salesforce and S3 (event capture), allowing us to create a CDP that can ingest and map information across multiple systems.
This could also grow into a better replacement of Google Analytics. I've been talking with the team, and they offer non-profit discounts, though the free tier will likely be plenty for Accounts + a few other marketing-based forms, pages, and the help center (where it is already installed).
The [privacy policy](https://posthog.com/privacy) is in alignment with our org goals, and their [documentation](https://posthog.com/docs) and [open-source code](https://github.com/PostHog/posthog) is an outstanding added perk (we could self-host this, but I don't think we have the resources in-house for managing that).

The built-in survey platform looks promising in replacing the very expensive PulseInsights.

This PR adds the ruby library to capture key events on server-side actions. It also contains the Javascript snippet to capture page enter/exit. Both of these identify the user, allowing us to join on tables in other systems for analysis on behavior.

Future todos:
- [ ] Implement [Sentry integration](https://posthog.com/docs/libraries/sentry) to tie errors with users
- [ ] Use Posthog feature flags for #1289 so we can easily turn on/off chat and do slow rollouts to determine support load.


app/controllers/newflow/login_controller.rb
```diff
@@ -18,7 +18,6 @@ def login
         success: lambda {
           clear_signup_state
           user = @handler_result.outputs.user
-          Sentry.set_user(id: user.id) if Rails.env.production?
```
Dantemss, line 21: Removing this because nothing here throws? Or some other reason?

app/controllers/newflow/social_auth_controller.rb
```diff
@@ -63,7 +66,7 @@ def oauth_callback
                 sign_up: view_context.link_to(I18n.t(:"login_signup_form.sign_up"), newflow_signup_path)
               )
             )
-          when :authentication_taken || :taken
+          when :authentication_taken || 'taken'
```
Dantemss, line 69: Seems odd that one would be a string and the other a symbol?

app/views/layouts/application.html.erb
```diff
@@ -17,18 +17,16 @@
...
+    <% end %>
 
+    <%= stylesheet_link_tag "application" %>
     <%= javascript_include_tag "application" %>
     <%= csrf_meta_tags %>
 
     <%= yield :head if content_for?(:head) %>
 
-    <% if !Rails.env.test? %>
-    <script src="https://cdn.optimizely.com/js/1345615444.js"></script>
```
Dantemss, line 29: So should we remove this if we find it in other apps?

config/secrets.yml
```diff
@@ -119,6 +119,8 @@ development:
   sentry:
       dsn: <%= ENV['SENTRY_DSN'] || 'invalid' %>
 
+  posthog_project_api_key: <%= ENV['POSTHOG_PROJECT_API_KEY'] || 'phc_n9UdrQ7MA5hF21CqmbGWoSxM1LfzxpcCxcjUBGfcdrE' %>
```
Dantemss, line 122: Well, just make sure this key is not valid for anything, as this repo is public.

Dantemss (approved):
Well, looks OK to me. I feel like maybe the security log calls could be combined with this, but not sure.

app/controllers/newflow/login_controller.rb
mwvolo: [just moved, not removed](https://github.com/openstax/accounts/pull/1293/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R39)

app/controllers/newflow/social_auth_controller.rb
mwvolo: testing this, because it's still failing, but the error code is definitely 'taken', so hoping a string will cover the error code

config/secrets.yml
mwvolo: this is the test key (and is available in the javascript, anyway); there is a separate production key in the parameter store
they key is write only, so it being a test key is okay in this instance (subject to change if someone starts hammering our posthog instance of course) - this just makes it easier to change

PR #1294, opened 2025-02-08
CORE-753 Make FindOrCreateApplicationGroups and FindOrCreateApplicationUser more resilient to concurrent calls

Dantemss:
By doing an `INSERT ... ON CONFLICT DO NOTHING` and then loading the pre-existing records after, these routines should in theory become more resilient to concurrent calls.


Dantemss:
I think all we can do for this is some regression testing

Dantemss:
I just realized I'm missing a `.preload(:group)` that was there before. Let me think about it for a sec...

Dantemss:
@mwvolo I added a bunch of preloading to a bunch of representers... hopefully I got all the N+1? It's hard to know though

Dantemss:
Sorry I realized one more thing @mwvolo
The code as-is would cause the primary key for ApplicationUser and ApplicationGroup to increment rapidly (even though most of them would be unused), because every time the `INSERT ... ON CONFLICT DO NOTHING` query runs, it increments the id sequence. So we could either prevent most of this increment by adding an extra check for the existence of the record (before the insert would run), or alternatively, we should change the column type for the `id` of ApplicationUsers and ApplicationGroups to bigint so it doesn't run out after 2 billion. Not sure which solution you'd prefer. Changing the datatype may cause some downtime for large tables though.

mwvolo:
What do you think about changing the primary key to a UUID?
How much downtime are we talking about? Currently, we have 13 ApplicationGroups and ~388k ApplicationUsers

Dantemss:
I think uuid should work fine as well. Downtime I imagine would be an hour or more, but we can maybe avoid it by doing the migration in parts. In the first release:

1. Add a nullable uuid column named uuid without a default value
2. Add a default value to generate a random uuid
3. Populate the column in the background
4. Make it not nullable

Then in another release:

1. Drop the old primary key constraint and add a new one in a migration
2. Use the new column as the primary key in the app code

mwvolo:
@Dantemss that sounds good - yeah, let's make sure this even works as expected before we make massive db changes

Dantemss:
@mwvolo added a commit to do this

mwvolo (approved):

PR #1299, opened 2025-03-13
reintroduce rails-settings-ui

mwvolo:
I made a fork of rails-settings-ui and got it working for Ruby 3.
Not necessary, but would help with the setting being used to show the chat widget (so we can easily disable if support load is too high).


Dantemss (approved):
🤞 that everything works

PR #1300, opened 2025-03-19
refactor sheerid webhook code, do not trigger on confirmed

mwvolo:
Major change is to prevent sheerid from overriding a confirmed status.

Took the liberty to try and refactor some of this code while here to make it more readable, handle error better, and increase test coverage.


Dantemss (commented):
So `handle_user_verification` will now ignore subsequent (duplicate?) hooks, but `update_user_with_verification_data` and `process_verification_step` still change the user info and faculty status. Is that expected?

PR #1302, opened 2025-03-26
CORE-763 Return 201 for duplicate external ids

Dantemss:
https://openstax.atlassian.net/browse/CORE-763

This is to solve issues with calls to this API in Assignable due to multiple tabs being open or whatever

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -242,7 +242,7 @@ class Api::V1::UsersController < Api::V1::ApiController
   # external_id
   ###############################################################
 
-  api :POST, '/user/external_id', 'Creates an external id for a given user.'
+  api :POST, '/user/external-ids', 'Creates an external id for a given user.'
   description <<-EOS
     Creates an external id for a given user.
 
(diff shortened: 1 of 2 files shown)


Dantemss:
Interestingly I'm getting failures running tests locally (even on main) but these don't seem to happen here.

app/controllers/api/v1/users_controller.rb
```diff
@@ -242,20 +242,34 @@ def find_or_create
   # external_id
   ###############################################################
 
-  api :POST, '/user/external_id', 'Creates an external id for a given user.'
+  api :POST, '/user/external-ids', 'Creates an external id for a given user.'
```
mwvolo, line 245: are you not concerned about changing the url for these? is Assignable the only thing using this endpoint?

mwvolo (approved):
approving, with one question/concern (but okay to merge if not a problem)

app/controllers/api/v1/users_controller.rb
mwvolo: Ah, gotcha ✅

PR #1631, opened 2025-05-06
CORE 890: use updated intl-tel-input

RoyEJohnson:
[CORE-890]
This branch uses the updated master branch of [intl-tel-input-rails](https://github.com/openstax/intl-tel-input-rails)

[CORE-890]: https://openstax.atlassian.net/browse/CORE-890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,7 +53,6 @@ gem 'coffee-rails'
 gem 'mini_racer'
 
 # JavaScript asset compressor
-gem 'uglifier'
 
 # Nicely-styled static error pages
 gem 'error_page_assets'
@@ -315,3 +314,5 @@ group :production do
   # Consistent logging
   gem 'lograge'
 end
+
+gem "terser", "~> 1.2"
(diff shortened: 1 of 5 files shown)


Dantemss:
Changing Uglifier to terser is a huge change. It probably requires the entire frontend to be regression-tested. Does that sound good?

RoyEJohnson:
Ugh. I'd rather not, but uglifier craps out on the build due to (I think) the optional-chaining operator (even though supposedly it supports it).
I've backed that commit out and you can see the issue [here](https://github.com/openstax/accounts/actions/runs/14869623914/job/41754827548?pr=1631#step:7:20).

RoyEJohnson:
I got the tip to replace Uglifier with terser [here](https://stackoverflow.com/a/76538920).
I'm currently rewriting all the optional-chaining statements, which I think is a different bad alternative.

Dantemss:
I didn't mean to discourage you, just point out that we'll need some good testing

Dantemss (approved):

PR #1645, opened 2025-05-15
CORE-986: Fix color contrast of signup heading

RoyEJohnson:
[CORE-986]
[CORE-1047]
New color contrast ratio is 5.
Also moved the disabled-submit styling to CSS
Also added some JS to control the disabled state of the Continue button on login.

For 1047, new color contrast ratio is at least 4.5.

[CORE-986]: https://openstax.atlassian.net/browse/CORE-986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-1047]: https://openstax.atlassian.net/browse/CORE-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/newflow/login/login_form.html.erb
+++ b/app/views/newflow/login/login_form.html.erb
@@ -93,7 +93,8 @@
                     <%= f.submit I18n.t(:"login_signup_form.continue_button"),
                                 disable_with: I18n.t(:"login_signup_form.continue_button"),
                                 class: 'primary',
-                                data: { ga_category: 'Login', ga_action: 'Click', ga_label: 'Email' }
+                                data: { ga_category: 'Login', ga_action: 'Click', ga_label: 'Email' },
+                                disabled: 'disabled'
                     %>
                 </div>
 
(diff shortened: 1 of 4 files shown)


app/views/newflow/login/login_form.html.erb
```diff
@@ -93,7 +93,8 @@
                     <%= f.submit I18n.t(:"login_signup_form.continue_button"),
                                 disable_with: I18n.t(:"login_signup_form.continue_button"),
```
mwvolo, line 94: does this need both the `disable_with` and the newly added `disabled` in the data?

RoyEJohnson: I'm actually not sure what disable_with does, but it looks like it provides a message to display when the item is disabled, whereas the disabled setting makes it disabled.

PR #1648, opened 2025-06-02
fixing specs from broken gh action failure

mwvolo:
Need to update specs to work with the accessibility changes on accounts (that were failing silently in GH actions).

- Removes a lot of specs that have been `xit` or `xscenario` for a long time for clarity.
- Remove _some_ failing specs that I feel are tested in other parts of the application (and in general, not a huge fan of feature tests).


mwvolo:
There seems to be an intermittent failure here, I think it has to do with Capybara.

PR #1704, opened 2026-03-19
CORE-1736: use ExternalIdRepresenter decorator for external_ids in UserRepresenter

jomcarvajal:
Related with: [CORE-1736]

[CORE-1736]: https://openstax.atlassian.net/browse/CORE-1736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/representers/api/v1/external_id_representer.rb
+++ b/app/representers/api/v1/external_id_representer.rb
@@ -4,7 +4,7 @@ module Api::V1
 
     property :user_id,
              type: Integer,
-             readable: true,
+             readable: false,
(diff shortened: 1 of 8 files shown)


mwvolo (approved):
This seems okay to me, most our data from accounts comes through glue jobs, not the api... and like you said, if someone is pinging the api with a user id, that should already be known.

PR #1709, opened 2026-04-17
Add role specific copy capability

mwvolo:
This pull request updates the educator signup flow to support audience-specific copy, particularly for K-12 educators, and introduces helper methods and tests to manage and verify this behavior. The main changes include adding logic to select copy based on the user's school type, updating the view to use this logic throughout, and ensuring correctness with comprehensive tests.

**Audience-specific copy support:**

* Added `educator_copy_audience` and `educator_copy` helper methods in `EducatorSignupHelper` to select copy based on the user's `school_type`, allowing for K-12-specific text overrides in the educator profile form.
* Updated the educator profile form view (`educator_profile_form.html.erb`) to use the new `educator_copy` helper for all user-facing strings, enabling dynamic audience-based copy throughout the form.
* Added a K-12-specific translation for the instructor role ("K-12 Teacher") in the English locale file.

**Testing and validation:**

* Added comprehensive helper specs for `educator_copy_audience` and `educator_copy` to ensure correct copy selection and fallback behavior.
* Added controller specs for the educator profile form to verify that the correct role label ("K-12 Teacher" or "Instructor") is shown based on the user's school type.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -75,6 +75,21 @@ en:
     school_issued_email_is_taken: Email already in use
     k12:
       instructor: K-12 Teacher
+      administrator: School or district administrator
+      other: Other school staff
+      complete_page_header: Complete your teacher profile
+      school_name: School or district name
+      describe_role: Which best describes your role at your school?
+      how_textbooks_chosen: How are textbooks chosen at your school or district?
+      chosen_by_instructor: Teachers choose textbooks for their classes
+      chosen_by_coordinator: School or district coordinator chooses the textbook
+      using_primary_textbook: Using an OpenStax book as my primary textbook
+      using_recommending_openstax: Recommending the book — my students use a different one
+      books_used: Which OpenStax book(s) are you using?
+      books_of_interest: Which OpenStax book(s) are you interested in?
+      include_ta_sections: Include all students across your classes
+      students_per_semester: Students per school year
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -73,6 +73,23 @@ en:
     school_issued_email_must_be_entered: Please enter your school-issued email address
     school_issued_email_is_invalid: Please enter a valid email address
     school_issued_email_is_taken: Email already in use
```
Dantemss, line 85: So... do you need to nest the old entries under "default" or is "default" special somehow?

Also, there are some old entries that are not reflected in k12, though maybe they are unused.

Dantemss (commented):
See comment above

config/locales/en.yml
mwvolo: this is really on purpose, because while we only have one other audience right now (k12), there could be a future where we want to override these for other audiences (like researchers), the default falls back [here](https://github.com/openstax/accounts/pull/1709/changes/BASE..e341d079171cdbc2770f03379e389838d1a0b94f#diff-6a061902692f13203917977b7aaf9eecc3877320869f34d1b2ecce2f7245426fR36)

to your point about k12 only having some keys, they are only overrides where applicable - this + not having default nested helps from having to repeat everything twice, if it's not different

Dantemss (approved):

PR #1721, opened 2026-06-12
School name autocomplete for student and educator signup

mwvolo:
For: https://openstax.atlassian.net/browse/DATA-300

## Summary

Replaces the two school-name fields in signup with one shared, accessible autocomplete that produces clean data:

- **Student signup** previously used a native HTML5 `<datalist>` populated from the openstax.org CMS endpoint — browser-styled dark dropdown, names only, and whatever text was in the box saved as free text (`school_id` never set).
- **Educator profile (step 4)** was a plain free-text field (shown when SheerID fails/is unavailable or on the CS form).

### What changed

- **New `School.search` scope** on the local Salesforce-synced `schools` table: case-insensitive substring match + trigram typo tolerance (uses the existing `gist_trgm_ops` index), prefix matches ranked first, max 10 results.
- **New unauthenticated `GET /i/schools?q=` endpoint** (`Newflow::SchoolsController`) returning `{id, name, city, state}`. The cross-origin CMS dependency is removed entirely.
- **Shared ARIA combobox component** (`newflow/school_autocomplete.js`, no new dependencies): debounced fetch with abort, keyboard navigation, suggestions show **name + city, state** to disambiguate same-named schools, and an always-present `Use "<typed text>"` footer row so schools not in our list can still be entered.
- **Clean data path:** picking a suggestion sets a hidden `signup[school_id]`; both handlers link the real `School` record and store its canonical name in `self_reported_school`. Editing after picking clears the link. Free text behaves exactly as before. The educator CS-form fuzzy match is skipped when a real pick was made; SheerID flows are untouched.
- **Graceful degradation:** if the endpoint is down, the field just acts as free text — signup is never blocked.

### Tests

- Model specs for `School.search` (ranking, typo tolerance, min length, cap)
- Controller specs for the endpoint (JSON shape, blank/short queries, logged-out access)
- Handler specs for valid / blank / nonexistent `school_id` on both flows, including fuzzy-match skip
- 3 selenium (`js: true`) feature specs driving the real dropdown: pick from list, use-as-entered, edit-clears-link


app/handlers/newflow/educator_signup/complete_profile.rb
```diff
@@ -71,6 +71,11 @@ def handle
         total_students = calculate_total_students
         return if errors?
 
+selected_school = School.find_by(id: signup_params.school_id) if signup_params.school_id.present?
+# If the school field was used but no valid id was provided, clear any existing link.
+if signup_params.school_name.present? || signup_params.school_id.present?
+  @user.school = selected_school
+end
```
Dantemss: Indentation issue?

Dantemss: Also the clearing functionality is maybe a bit hidden... you'd have to type a non-empty non-whitespace string and not select anything? Probably

Dantemss (commented):
Other than the stuff I pointed out the rest seems fine

PR #1728, opened 2026-07-18
Bump patched gem versions for open Dependabot security alerts

mwvolo:
## Summary

Bumps gems to their patched versions to close out open Dependabot security alerts, plus one locale fix required by the doorkeeper bump.

Net diff against `main` is three files: `Gemfile`, `Gemfile.lock`, and `config/locales/doorkeeper.en.yml`.

| Gem | Before | After | Fixes |
|---|---|---|---|
| rails (actionmailer/actionpack/actiontext/actionview/activerecord/activestorage/activesupport) | 6.1.7.8 | 6.1.7.10 | activestorage: unsafe transform methods (critical), path traversal, DoS via range requests, content-type bypass; activesupport: DoS in number helpers, ReDoS in `number_to_delimited`, XSS in `SafeBuffer#%`; activerecord: ANSI escape injection in logging; actionpack/actionview/actionmailer/actiontext: ReDoS and XSS issues |
| rack | 2.2.14 | 2.2.23 | multipart parser memory-exhaustion DoS (multiple CVEs), `Rack::Directory` traversal/XSS, `Rack::Static` prefix bypass, `Rack::Sendfile` regex injection, query parser `params_limit` bypass, quadratic-complexity Accept-Encoding DoS, Content-Length mismatch |
| nokogiri | 1.18.9 | **1.18.10** (1.19.x not reachable on Ruby 3.1 — see note) | vendored libxml2 bump |
| oj | 3.16.6 | 3.17.4 | multiple use-after-free bugs, integer overflow, heap/stack buffer overflows in `Oj.dump` and `Oj::Parser` |
| jwt | 2.9.3 | 2.10.3 | empty-key HMAC signature bypass (CVE-2026-45363) |
| faraday | 1.10.4 | 1.10.6 | SSRF via protocol-relative URL host override, stack exhaustion DoS via deeply nested query params |
| concurrent-ruby | 1.3.6 | 1.3.7 | `AtomicReference#update` livelock on NaN, `ReentrantReadWriteLock` overflow/corruption bugs |
| bcrypt | 3.1.20 | 3.1.22 | integer overflow causing zero key-strengthening iterations at cost=31 on JRuby |
| rexml | 3.3.9 | 3.4.4 | DoS parsing malformed XML |
| net-imap | 0.5.7 | 0.5.15 | command-injection and DoS issues, STARTTLS stripping (IMAP isn't used here — pure transitive dep via `mail`) |
| websocket-driver | 0.7.7 | 0.8.2 | memory exhaustion in header/protocol parsing (via ActionCable, which isn't mounted) |
| puma | 6.4.3 | 7.2.1 | PROXY protocol v1 parsing DoS (not applicable — `config/puma.rb` doesn't use `proxy_protocol`); bumped anyway per Dependabot ask |
| **doorkeeper** | **5.7.1** | **5.9.3** | **Not a security bump — pulled in by the resolver. Called out separately below because it is a direct dependency, not incidental.** |

Also bumped by the resolver, low-risk: `jquery-rails` 4.6.0 → 4.6.1, `font-awesome-rails` 4.7.0.8 → 4.7.0.9, `rails-i18n` 7.0.9 → 7.0.10, `meta_request` 0.8.3 → 0.8.6, `i18n-tasks` 1.0.14 → 1.1.2, `lograge` 0.14.0 → 0.15.0, `marcel` 1.0.4 → 1.2.1, `active_attr` 0.17.0 → 0.17.2, `mini_portile2` 2.8.8 → 2.8.9, plus new transitive dep `prism` 1.9.0.

`omniauth` (1.9.2) and `css_parser` (1.22.0) are unchanged — handled separately.

## doorkeeper 5.7.1 → 5.9.3 needs a callout

`gem 'doorkeeper'` is a **direct, unpinned dependency** (`Gemfile:70`) and is the OAuth provider for every OpenStax product, so this bump was reviewed rather than waved through. It is safe, with two behavior changes worth knowing:

**1. Consent-screen copy — fixed in this PR.** 5.9.1 ([#1784](https://github.com/doorkeeper-gem/doorkeeper/pull/1784)) moved a hardcoded colon out of `authorizations/new.html.erb` and into the gem's default `en.yml` string. We don't override that view, but we do override that locale key, and our copy predated the change. Left alone, the OAuth consent screen would render "This application will be able to" with no colon above the scope list. Fixed in `config/locales/doorkeeper.en.yml`.

**2. Token introspection omits `exp` for non-expiring tokens.** 5.9.1 ([#1818](https://github.com/doorkeeper-gem/doorkeeper/pull/1818)) changed `token_introspection.rb` to `response[:exp] = @token.expires_at.to_i if @token.expires_at`. This app sets `access_token_expires_in nil`, so every token is non-expiring: `POST /oauth/introspect` previously returned `"exp": 0` and now omits the key. No action needed — the old value was itself wrong (epoch zero reads as "expired 1970-01-01"), so any resource server reading it for expiry would already have been rejecting every token Accounts issues. The new behavior is what RFC 7662 §2.2 calls for.

Checked and clear: no new migrations or schema columns anywhere in the 5.7.1→5.9.3 range; `Doorkeeper::ApplicationsController` is byte-identical between the two versions, so `main`'s new role-audit screens (#1741) are unaffected; every option set in `config/initializers/doorkeeper.rb` is still valid API; the `TokensController` alias patch in `doorkeeper_models.rb` still resolves; the refresh-token locking changes only apply when `revoke_previous_refresh_token_on_use` is true, which we never set.

## nokogiri note

The requested target was 1.19.4, but **nokogiri >= 1.19.0 dropped Ruby 3.1 support** and this app is pinned to Ruby 3.1.6. 1.18.10 is the newest installable version, and per nokogiri's changelog 1.19.0 is "otherwise functionally identical to v1.18.10" for CRuby — the CVEs fixed in 1.18.9/1.18.10 (CVE-2025-6021, CVE-2025-6170, CVE-2025-49794/49795/49796) are covered. The additional CRuby fixes in 1.19.1–1.19.4 (XPathContext use-after-free, `NodeSet#[]` OOB read, XSLT transform memory leak, CSS tokenizer ReDoS, `initialize_copy_with_args` invalid read) are **not** included and won't be reachable until this app moves to Ruby >= 3.2. Added an explicit `gem 'nokogiri', '~> 1.18.10'` pin with a comment so it's easy to revisit.

## puma note

6.4.3 → 7.2.1 is a major bump. Puma 7's `required_ruby_version` is `>= 3.0`, so Ruby 3.1.6 is fine. Puma 7 renamed several callback hooks (`on_worker_boot` → `before_worker_boot`, etc.), but `config/puma.rb` only uses `before_fork` plus `threads`/`workers`/`preload_app!`/`stdout_redirect`/`worker_timeout`/`wait_for_less_busy_worker`/`plugin :tmp_restart` — all confirmed present in `puma-7.2.1/lib/puma/dsl.rb`. `puma_worker_killer` reaches puma only through `ObjectSpace.each_object(Puma::Cluster)` and declares `puma >= 2.7` with no upper bound. Verified by booting `puma -C config/puma.rb` in clustered mode: both workers came up and served real responses.

Pinned to `~> 7.2` rather than left unconstrained, since a plain `bundle update puma` pulls puma 8.0.2 — a bigger jump not covered by this pass.

One cosmetic change: puma 7 now warns that `before_fork` only runs in cluster mode. `NUM_WORKERS` defaults to `Etc.nprocessors`, so production runs clustered and PumaWorkerKiller still starts; the warning only appears in single mode (e.g. during specs). Behavior is unchanged from puma 6.

## jwt note

`omniauth-google-oauth2` 0.8.2 verifies Google's `id_token` via `::JWT.decode(access_token['id_token'], nil, false)` — nil key, verification disabled. The 2.10.3 CVE fix rejects nil/empty HMAC keys *when signing and verifying*, which that call doesn't do. Confirmed by decoding a Google-shaped RS256 `id_token` through the identical call under 2.10.3: decodes fine, while `JWT.encode(payload, nil, 'HS256')` correctly raises. Accounts' own SSO cookies use the separate `json-jwt` gem (`lib/sso_cookie.rb`), not this one.

## Scope change

Reverted `b3f084f` ("Fix login token uniqueness: retry on collision in `refresh_login_token`"), which was added mid-PR and is unrelated to dependency bumps. The premise didn't hold up: `User` already validates `login_token` uniqueness (`app/models/user.rb:123`), both production callers save through validations so a collision would surface as a validation error rather than `RecordNotUnique`, and the token is 128 bits of `SecureRandom`. The check-then-write loop also left open the TOCTOU window it aimed to close, with no `RecordNotUnique` rescue anywhere in `app/` or `lib/`. If we want defense-in-depth here it belongs in its own PR with a rescue rather than a pre-check.

## Test plan

- [x] Merged `origin/main` — clean, and semantically clean: a follow-up `bundle install` produced zero further lockfile changes, and `main`'s Salesforce client-credentials work survived intact (`openstax_salesforce` 9.0.0, `restforce` 8.0.1)
- [x] `bundle exec rspec` on the merged branch: **1249 examples, 0 failures, 2 pending** (both pre-existing `xit` specs in `spec/handlers/newflow/educator_signup/sheerid_webhook_spec.rb`)
- [x] Confirmed via lockfile diff that every listed gem is at/above target (except nokogiri, documented above) and that `omniauth`/`css_parser` are unchanged
- [x] `db/schema.rb` at `2026_08_19_130000` with no pending migrations
- [x] Clustered puma boot test (2 workers, real HTTP responses)
- [x] Google `id_token` decode verified directly under jwt 2.10.3
- [x] Confirmed no multipart/file-upload endpoints and only flat `.permit()` allow-lists, so rack's new query/multipart limits aren't reachable; ActiveStorage (`marcel`) and ActionCable (`websocket-driver`) are both unused


mwvolo:
@copilot resolve the merge conflicts in this pull request

mwvolo:
@copilot can we update ruby to remove the block with nokogiri?

mwvolo:
### `tests (3.1, 16)` is red on a pre-existing flaky spec, not on this PR

`spec/features/pose_terms_spec.rb:9` ("Terms agrees to terms when signatures not present") fails with:

```
Failure/Error: expect(page).to have_button((t :"terms.pose.agree"), disabled: false)
  expected to find button "Accept and continue" that is not disabled but there were no matches
```

This is not caused by the gem bumps. Evidence:

1. **It has already failed on `main`.** Run [31916944973](https://github.com/openstax/accounts/actions/runs/31916944973) (2026-08-16, commit `4527daf9`) failed on this exact spec with the identical message, with the old gem versions and none of this PR's changes. It passed on a re-run of the same SHA, which is why it went unnoticed.
2. **It fails locally with the old gems.** Running just this spec in a `main` checkout (jquery-rails 4.6.0, rails 6.1.7.8, puma 6.4.3) reproduces the same failure.
3. **The full suite passes on this branch.** `bundle exec rspec` against the merged branch: **1249 examples, 0 failures, 2 pending** (the 2 pending are the pre-existing `xit` specs in `sheerid_webhook_spec.rb`).

The spec is nondeterministic and load-sensitive, which is why CI hits it more reliably than a local single-file run — CI runs `WORKERS=4 parallel:spec`, so four workers compete for CPU. It surfaced on this PR rather than the previous green run because the example count changed (1250 → 1248), which reshuffled how `parallel_tests` splits specs across workers.

Leading hypothesis for the cause, **not yet proven**: the submit button carries `data: { disable_with: (t :".agree") }`, whose label is identical to the button's normal label. After `click_button`, jQuery-UJS disables the button while keeping that same text. The retry loop guards re-entry with `page.has_current_path?(profile_newflow_path, wait: 1)`, so on a slow run it can re-enter before the submitted page has loaded and assert against the stale, UJS-disabled button. There's an existing comment in `app/assets/javascripts/newflow/newflow_ui.js.coffee:41-45` about racing with jQuery-UJS's button disabling, so this area has caused trouble before.

I could not confirm the mechanism: any instrumentation added before the assertion (`evaluate_script`, extra DOM queries) changes the timing enough to make the failure disappear. Worth its own issue/PR rather than a speculative fix here — the likely fix is to wait for the submitted page to actually load before re-entering the retry loop, instead of relying on a 1-second path check.

### What changed in this PR since the last review

- Merged `main`. The lockfile merge is semantically clean — a follow-up `bundle install` produced zero further changes, and `main`'s Salesforce client-credentials work is intact (`openstax_salesforce` 9.0.0, `restforce` 8.0.1).
- Reverted `b3f084f` (the `refresh_login_token` retry loop). It was unrelated to dependency bumps and its premise didn't hold — `User` already validates `login_token` uniqueness at `app/models/user.rb:123`, both production callers save through validations so a collision surfaced as a validation error rather than `RecordNotUnique`, and the token is 128 bits. The check-then-write loop also left open the TOCTOU window it aimed to close. `app/models/user.rb` is now byte-identical to `main`.
- Fixed one real regression from the doorkeeper bump: 5.9.1 moved a hardcoded colon out of `authorizations/new.html.erb` into the gem's default `en.yml`. We override that locale key but not the view, so the OAuth consent screen would have lost the colon above the scope list. Fixed in `config/locales/doorkeeper.en.yml`.
- Rewrote the description: `doorkeeper` 5.7.1 → 5.9.3 now has its own section (it's a direct, unpinned dependency, not an incidental transitive bump), the incorrect `addressable` entry is removed, and `css_parser` is stated at its actual version.

Net diff against `main` is three files: `Gemfile`, `Gemfile.lock`, `config/locales/doorkeeper.en.yml`.

PR #1732, opened 2026-07-18
Retire the legacy auth flow: delete legacy controllers, redirect old URLs to newflow

mwvolo:
## Summary

Stacked on #1731 (Rails upgrade Phase 0). Deletes the entire legacy auth flow (~2,200 lines removed net) now that traffic data confirmed it's noise-level, per Michael's verdict. **No old URL 404s** — every legacy GET becomes a named, query-string-preserving, CloudFront-prefix-aware route redirect to its newflow equivalent, so `login_path`/`profile_path`/`signup_path` helpers and old bookmarks/emails keep working.

## What's deleted

- `app/controllers/legacy/*` (all 5 controllers), their views, and legacy-only handlers (SessionsCreate, SessionsLookupLogin, SessionsReauthenticate, IdentitiesSendPasswordEmail, IdentitiesSetPassword, LogInByToken)
- The legacy omniauth strategies + custom_identity; initializer now registers only `facebooknewflow`/`googlenewflow` (newflow social login was always independent — its strategies pin `path_prefix '/i/auth'`)
- `omniauth-twitter` — **Twitter login is fully removed.** Newflow never offered it, so Twitter-only accounts already had no UI path; they can recover via password reset.
- 40 spec examples whose subjects were deleted (suite goes 1251 → 1211, 0 failures, accounted example-by-example)

## What discovery kept or moved (live non-legacy dependencies)

- `/signup/profile` → new top-level `SignupProfileController` (the global `complete_signup_profile` before_action sends any needs-profile user here, including mid-OAuth-authorize)
- `PUT /profile` → moved to `OtherController#update` (the newflow profile page's inline-edit fields PUT here; gained an explicit `newflow_authenticate_user!`)
- Account-claiming flow (`confirm_unclaimed`) repointed to newflow password forms — which surfaced and fixed a real infinite-redirect bug (`check_if_password_expired` vs the retargeted `password_reset_path`) and a stored-oauth-url gap in `Newflow::PasswordManagementController`
- `omniauth-identity` gem kept (the `Identity` model subclasses it); only the strategy was retired

## Verification

- Full suite 1211 examples / 0 failures / 2 pre-existing pending (run twice)
- Production `assets:precompile` green
- curl redirect matrix for every legacy GET incl. query-string carry-through (`/login?r=%2Ffoo` → `/i/login?r=%2Ffoo`)
- `SIMULATE_CLOUDFRONT=true` boot check; redirect lambda uses `req.script_name`
- Independently reviewed (spec compliance + code quality + whole-diff auth-infrastructure review)

## Reviewer note

The one judgment call worth eyes: removal of Twitter login and the deletion of `Legacy::AuthenticationsController` (verified: the newflow profile page manages login methods exclusively via `/i/auth/:provider` and never referenced it).


mwvolo:
Copilot review triage (commit 223dbbe7, branch rebased onto updated rails-upgrade-prep):

**Both findings fixed:**
- **Mass assignment in `OtherController#user_params` — confirmed real** (inherited verbatim from the deleted legacy controller, but live). The single-string x-editable branch now validates against `ALLOWED_STRING_UPDATE_FIELDS = %w[username]` and returns 400 otherwise. Verified `username` is the only single-field editable that PUTs a plain string to `profile_path` (the composite name editor always submits an object; email edits go to `/contact_infos`). New controller specs assert the 400 path leaves attributes unchanged.
- **Redirect coverage**: added `spec/requests/legacy_redirects_spec.rb` — GET /profile → /i/profile, query-string preservation (`/login?r=%2Ffoo` → `/i/login?r=%2Ffoo`), /signup and /logout redirects, and an authenticated PUT /profile hitting `other#update`.

PR #1739, opened 2026-08-17
Move Salesforce auth to the OAuth client credentials flow

mwvolo:
Salesforce retires the OAuth 2.0 username-password flow in Spring '27 — enforcement **Feb 20, 2027**. That's the flow Accounts uses to authenticate for every lead/contact/school sync, so it stops working on that date.

This moves us to the **client credentials** flow, which is what Salesforce recommends for server-to-server integrations.

## What changed

Almost nothing in app code — the entire auth surface was the initializer plus two methods in `openstax_salesforce`. Restforce picks its auth middleware from whichever options are present, so dropping the username/password is what selects client credentials.

- `openstax_salesforce` 8.3.0 → **9.0.0** ([gem PR](https://github.com/openstax/openstax_salesforce/pull/85))
- Dropped `SALESFORCE_USERNAME`, `SALESFORCE_PASSWORD`, `SALESFORCE_SECURITY_TOKEN` from `.env.example` and all three `config/secrets.yml` sections
- Auth is now the connected app's consumer key + secret alone

`restforce` moves 7.5.0 → 8.0.1 as a consequence (the gem now requires `>= 7.1`, where the client-credentials middleware first appears). Restforce 8.0.0's only breaking change is dropping Ruby 3.0 support; we're on 3.1.6. The lock diff was kept minimal with `bundle update --conservative`.

## ⚠️ Deploy requires two things outside this PR

**1. `SALESFORCE_LOGIN_DOMAIN` must become the org's My Domain** in every environment.

Salesforce **rejects** client-credentials token requests at `login.salesforce.com` / `test.salesforce.com`. The gem raises on either rather than letting you find out via a confusing 400. Production is `openstax.my.salesforce.com`; sandboxes are `openstax--<sandbox>.sandbox.my.salesforce.com`.

Also delete the three retired vars from each environment — a leftover `SALESFORCE_PASSWORD` or `SALESFORCE_SECURITY_TOKEN` with no username now raises deliberately, so a half-migrated config fails loudly instead of silently switching flows.

**2. The connected app needs client credentials enabled + a Run As user.**

In Setup → App Manager → **OpenStax Accounts** → Edit:
- Under *API (Enable OAuth Settings)*, check **Enable Client Credentials Flow**
- Then Manage → Edit Policies → under *Client Credentials Flow*, set the **Run As** user

```
$ RAILS_ENV=test bundle exec rails runner 'puts OpenStax::Salesforce::Client.new.authentication_middleware'
Restforce::Middleware::Authentication::ClientCredential
```

## Rollout

Sandbox first: set the My Domain there, run a sync, then confirm Setup → Login History shows Login Subtype `OAuth Client Credentials` instead of `OAuth Username-Password`. Reversible at any point by putting the three env vars back — 9.0.0 still honors the old flow when a username is set.


Dantemss (approved):
I guess we were already using the client credentials flow so this just drops the username/password/security token config? Makes sense

mwvolo:
yep - this just brings accounts up to speed with how assignable is already doing the auth handshake with sf

PR #1740, opened 2026-08-19
CORE-2701 Removed unused ApplicationGroups, GroupNestings, Group APIs

Dantemss:
https://openstax.atlassian.net/browse/CORE-2701

- All existing ApplicationGroups in production belong to OpenStax Tutor, which is defunct.
- None of the repos actually call the Group APIs. ApplicationGroup syncing was removed from the openstax_accounts ruby gem in 2019.
- GroupNesting was never used.
- Group, GroupMember, GroupOwner are still used internally in Accounts to control application ownership, so they remain.


Dantemss:
Maybe we can migrate the application owners back to single users and remove the groups entirely, but I figured this part of the cleanup was basically free.

mwvolo (approved):
this is very exciting! nice cleanup of unused functionality (which really confuses claude)

PR #1750, opened 2026-09-03
Move SheerID's iframe to a program URL that reports its own height

mwvolo:
## Why

Step 3 loaded a hosted `offers.sheerid.com` page in a fixed-height iframe. That page never reports its content height — `updateHeight` appears nowhere in its bundle — so the frame was pinned at a hardcoded size with dead space under it, and the number had to be re-guessed whenever SheerID changed the form. Two of my own guesses at it were wrong by 8–20rem.

A program verification URL reports it. Given `installType=cdn_inline_iframe` and a `verificationIframeUid`, the form posts `updateHeight` and `newflow/sheerid_iframe.js` applies it, clearing the CSS floor. Measured against the live program, the frame tracks each step: **500 → 650 → 961 → 1057px** as the form grows.

## Why not `sheerid-install.js`

Their installer's entire contribution to auto-height is three lines, reimplemented here. Loading it would mean an **ES-module-only** script (older institutional browsers would get *no* iframe at all, against this project's ES5-safe rule), a **third-party script in the origin that handles passwords**, and it **copies every query param on the page into the third-party URL** — here that would hand SheerID `r`, `sp` and `client_id`.

## What else is in it

| Change | Why |
|---|---|
| Program id → `Settings::SheerId` (+ `SHEERID_PROGRAM_ID` override) | Not a secret: it ships in the iframe URL every visitor loads. The admin-editable `sheer_id_base_url` setting goes away; the orphan row is simply never read (the settings UI builds from `defined_fields`, confirmed in the gem). |
| Prefill → `setViewModel` over postMessage | A program URL **ignores** prefill query params — the form comes up blank. Sending them put the user's email in a third-party URL for nothing. |
| Redirect to step 4 from `ON_VERIFICATION_SUCCESS` | The hosted page carried a per-environment success redirect. One program serves every environment, so its single redirect setting cannot send dev, staging and prod to their own step 4. |
| `allow="camera *; clipboard-read *; clipboard-write *"` | SheerID's installer delegates these; document upload needs the camera and was otherwise denied it silently. Microphone and geolocation deliberately left off. |
| `expect_sheerid_iframe` rewritten | It asserted the heading, submit label, the "can't find your school" hints and Country defaulting to United States — all Program Builder content, and the new program differs on every one. It now checks that the form rendered and that our prefill reached it. |

## Verification

- 390 examples, 0 failures (`spec/features/newflow`, `spec/helpers`, `spec/controllers`)
- Live against program `6a7c5eb82050c35ab0e1f676`: auto-height settles at 1057px and clears the floor; prefill lands; OpenStax logo renders in the frame
- Message handling exercised against the exact payload shapes SheerID sends: nested `updateHeight` applied, mismatched `verificationIframeUid` ignored, flat legacy `{action:'updateHeight'}` honoured, malformed payloads neither throw nor resize, `ON_VERIFICATION_SUCCESS` navigates
- Asset compiles and is in the precompile list

**Not covered by specs: a real end-to-end approval.** Driving SheerID's cross-origin form from Capybara hits their verification rate limits — the spec helper still carries a commented-out `expect(page).not_to have_text('Verification Limit Exceeded')` from a previous attempt. Click through once with first name `APPROVED` when testing this.

## Needs doing outside this PR

- Program Builder for `6a7c5eb82050c35ab0e1f676`: the copy reads *"Verify you're **a educator**"*, the heading is *"Unlock this Educator-Only Offer"*, and **Country is no longer preselected** so the school field starts disabled
- Clear or repoint that program's own redirect URL so it does not race the one here
- Confirm one program is right for every environment, or set `SHEERID_PROGRAM_ID` per environment — it currently renders in TEST MODE


OpenStaxClaude:
@mwvolo — heads up, I pushed three commits to `mv-sheerid-program-url`. Roy asked me to resolve conflicts on #1757, which is stacked on this branch, and the conflicts were all here rather than there. Happy to drop any of it if you'd rather own it.

### 1. `main` merged in (8112417c..1718fb8e)

This branch was stacked on #1748 and #1751. Both squash-merged, so `main` now carries one squashed copy of work this branch still holds unsquashed — and a *later* copy, with the review changes that landed after the split. That was the whole conflict; none of it was SheerID work.

I took `main`'s side for every file in that lineage, because in each case `main`'s is the newer one:

| File | What `main` has that this branch didn't |
|---|---|
| `switch_signup_role.rb`, `signup_controller.rb`, `en.yml` | `user.with_lock`, Salesforce job deferred until after the transaction, `switched_role_notice` flash |
| `update_existing_salesforce_lead.rb` + spec | re-raises in a real delayed job, swallows only when `perform_later` runs inline |
| `sheerid_webhook.rb` | your post-switch guard already, inside the `with_lock` it added, plus the `expiredVerification` filter |
| `create_or_update_salesforce_lead.rb`, `security_log.rb` | your `unless user.student?` guard and `outputs.lead_saved`, plus #1722's Contact sync |

Two bits that weren't mechanical, worth your eyes:

- **`spec/controllers/newflow/signup_controller_spec.rb` merged wrong and clean.** Both sides had added a byte-identical `describe 'POST #switch_role'` block. Git can't tell that's the same change, so it took both and produced two copies with no conflict marker. Took `main`'s file, which also has #1758's `signup_done` capture specs.
- **`CLAUDE.md`** is the union of both sides. The escape-hatch paragraph still said the frame is "pinned at 80rem"; it's the 110rem floor now, so I corrected the number.

Everything this branch actually contributes is untouched — `Settings::SheerId`, the `data-sheerid-*` attributes, `sheerid_iframe.js`, the 110rem floor over `main`'s fixed 100rem, the rewritten `expect_sheerid_iframe`. Net diff against `main` is the same 13 files it was.

### 2. A guard on `goToNextStep` (4fe96aa6, 8112417c) — and CodeQL is still red

Merging `main` in is what first ran CodeQL against this branch; the previous head has no CodeQL check at all. It reports **`js/xss-through-dom` (high)** on `window.location.assign(path)`, where `path` is `data-sheerid-success-path`.

No live bug — that attribute is written from `educator_profile_form_path`. But it is a navigation sink reading a DOM attribute, and `location.assign` will happily run a `javascript:` URL or leave the origin for a protocol-relative one, so I added a guard that fails closed if the path is ever made dynamic.

**It does not silence CodeQL.** I tried a character test first (not modelled as a sanitizer), then an anchored regex and navigating the match rather than the original string — CodeQL still treats a `RegExp.exec` result as tainted, so the alert just moved line. I stopped there rather than keep reshaping your code around a static analyser. As I read it there are two honest options:

1. **Dismiss alert #33 as a false positive** in the Security tab. It is one, and this is your call, not mine.
2. Render the destination as a hidden `<a href>` server-side and `.click()` it, so no string reaches a sink. Clears the alert properly, but puts a hidden anchor in the markup — and given the neighbouring PR is an a11y fix, I'd want it `hidden` rather than `tabindex="-1"`, and I'd rather you chose it than find it.

The guard itself stands on its own merits either way; say the word and I'll revert it.

### Checks

`tests (3.1, 16)` **1383 examples, 0 failures, 2 pendings**. `assets`, `Analyze (ruby)`, `Analyze (javascript)` all pass. `CodeQL` is the one red, per above. Locally I also ran the specs for every file I touched in the resolution — 60 examples, 0 failures.

PR #1757, opened 2026-09-17
Give the SheerID form the focus ring its own stylesheet suppresses

OpenStaxClaude:
Jira: [CORE-909](https://openstax.atlassian.net/browse/CORE-909) — *(SheerID) Indicate keyboard focus for Validate button*

**Draft, and stacked on #1750** — base branch is `mv-sheerid-program-url`, not `main`. #1750 is what gets us onto a program verification URL and the current `@sheerid/jslib`; this sits on top and adds the focus ring. Rebase onto `main` once #1750 lands.

## Why

The audit (WCAG 2.4.7 AA, High) flags three controls on step 3 with no visible focus indicator: the submit button, the document dropzone and "Add file". All three are SheerID's, inside a cross-origin iframe, so no stylesheet of ours can reach them. The usual vendor answer — **Theme > Custom CSS in MySheerID — is not offered on our program**; it exposes a palette and nothing else (Roy checked).

Their form does accept `customCss` over the **same postMessage channel #1750 already uses for prefill**, and renders it into a `<style>` tag of its own, taking it ahead of any customCss the program theme carries:

```js
// @sheerid/jslib@2.287.0
case "setOptions": m(e.data.options); break;
// ...
customCssFromOptions: getOptions().customCss   // `t || e?.customCss` — options win
```

The sanitising pass that strips `body,`, `label,`, `button[type=submit]` and `h1` rules from customCss only runs when `window.self === window.top`. We're in an iframe, so the rules go in verbatim.

So the CSS travels to the form instead of being applied to it, and it lives in our repo where it's reviewable — no vendor console, no third-party script in the origin that handles passwords.

## What it does

`sendFocusStyles()` in `newflow/sheerid_iframe.js` posts `{action: 'setOptions', options: {customCss}}` on `ON_VERIFICATION_READY`, **before** the prefill. That ordering is the whole trick: options are read when the form next renders and then stay set, and the prefill is what triggers that render. It re-sends on `updateHeight` (capped at 3 sends) to cover a visitor with no prefill to send, where the render would otherwise wait for the first keystroke.

The ring is `2px solid #026AA1`, offset 2px — the same treatment the rest of the signup flow uses, so crossing into the frame doesn't change what focus looks like. 5.9:1 against the form's white card and 5.3:1 against its grey dropzone, where 2.4.7 asks for 3:1.

## Measured, not assumed

A real Tab walk into the live frame (program `6a7c5eb8`), computed styles read after `.sid-btn`'s `transition: all .2s` settles. Same harness, same program, with and without the `sendFocusStyles()` calls:

| Control | Without | With |
|---|---|---|
| `sid-country`, `sid-first-name`, `sid-last-name`, `sid-email` | **`outline-style: none`** — no indicator at all | `solid 2px rgb(2, 106, 161)` offset 2px |
| `sid-submit-btn-collect-info` | `auto 1px rgb(16, 16, 16)` | `solid 2px rgb(2, 106, 161)` offset 2px |
| `sid-input-select-button`, `sid-h-btn-link-like`, `sid-privacy-policy`, `sid-faq`, `sid-test-mode-flag__link` | `auto 1px rgb(16, 16, 16)` | `solid 2px rgb(2, 106, 161)` offset 2px |

All **ten** tab stops on the step, not just the one in the ticket title. Two things worth flagging from that table:

- The four text inputs have **no** indicator without this — `.sid-text-input:focus{outline:none}` is still in `@sheerid/jslib@2.287.0`. That's a second 2.4.7 failure on the same screen that the audit didn't call out.
- The migration in #1750 does restore a *browser-default* ring on the button (the old microsite's `.sid-btn{outline:none}` is gone in the current lib). It computes as `auto 1px rgb(16, 16, 16)`, and in a headless screenshot at the edge of the dark button it isn't distinguishable — see below. So the migration alone doesn't close the finding.

Screenshots of the focused submit button, taken from that same harness, tell
the same story: without the fix the button edge looks identical to its
unfocused state; with it there is a blue ring standing 2px clear of the button.
(I can't attach images from here — they're reproducible with the harness
described above, or just Tab to the button on a branch build.)

## Coverage

`expect_sheerid_iframe` already proves our `setViewModel` message landed by
checking the prefill in the frame; the `setOptions` message is the same
contract, so the assertion that the CSS arrived sits next to it. Asserted in JS
because a `<style>` element has no rendered text for Capybara to match.

## Not verified live

The dropzone and "Add file" only exist on the `docUpload` step, which needs a real verification to reach, and driving that from a harness hits SheerID's rate limits. Their classes are named explicitly in the rule (`.sid-dropzone-wrap__dropzone`, `.sid-uploadzone__button`, `.sid-file-list__remove-btn`), and a `.sid-form-wrapper button/[tabindex]:focus-visible` catch-all covers whatever else that step puts in the tab order, with `!important` to beat the inline `outline:none` react-dropzone puts on the dropzone. **Please Tab through the upload step once when testing** (first name `APPROVED`, per #1750) and confirm the ring is there.

## Also worth doing, outside this PR

`.sid-text-input:focus{outline:none}` is shipping to every SheerID customer. Worth a support ticket from whoever owns the relationship — this works around it, it doesn't fix it for anyone else.

[CORE-909]: https://openstax.atlassian.net/browse/CORE-909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OpenStaxClaude:
Checks pass on the latest push: `assets` and `tests (3.1, 16)` (1282 examples, 0 failures).

Worth noting what that second one covers here — `parallel:spec` runs the feature specs too, headless Chrome against the live program, so the new assertion in `expect_sheerid_iframe` is real end-to-end confirmation that the `setOptions` message lands and the form renders our CSS. It agrees with the standalone Tab-walk measurements in the description, from a different driver.

Staying draft until #1750 merges; I'll rebase onto `main` and mark it ready then.

RoyEJohnson (changes requested):
Resolve conflicts

OpenStaxClaude:
@RoyEJohnson Conflicts resolved — this PR and #1750 both report `mergeable` now.

Worth saying where they actually were, because it wasn't this PR. Against its base this branch merged clean the whole time; the conflict was one level down. **#1750 was stacked on #1748 and #1751**, both since squash-merged, so `main` now carries a single squashed copy of work that branch still holds unsquashed — and a later copy of it, with the review changes that landed after the split. Nothing SheerID-related was involved.

So the fix was: merge `main` into `mv-sheerid-program-url`, then merge that up into this branch. I took `main`'s side for every file in the #1748/#1751 lineage, since `main`'s is the newer one in each case (`with_lock` on the role switch, the delayed-job re-raise in `UpdateExistingSalesforceLead`, #1722's Contact sync). The full breakdown is [in a comment on #1750](https://github.com/openstax/accounts/pull/1750#issuecomment-5763309839) and in the merge commit message, since that's where the work landed.

Two things from it that are worth a second pair of eyes:

- **One file merged wrong and merged *clean*.** Both sides of `spec/controllers/newflow/signup_controller_spec.rb` had added a byte-identical `describe 'POST #switch_role'` block. Git can't recognise that as the same change, so it kept both — two copies of four identical examples, no conflict marker, nothing failing. Took `main`'s file. If you only skim one thing here, skim that.
- **`CLAUDE.md`** is the union of both sides, with the escape-hatch paragraph's stale "pinned at 80rem" corrected to the 110rem floor that actually ships.

This PR's own contribution is unchanged and still three files, +110/-2.

### Checks

Green on this PR: `tests (3.1, 16)` at **1383 examples, 0 failures, 2 pendings**, plus `assets`, `CodeQL` and both `Analyze` jobs.

I also ran the specs for every file I touched in the resolution locally — 60 examples, 0 failures. Three examples in `educator_signup_flow_spec.rb` fail in my sandbox, but they fail identically on unmodified `main` there (it's on Ruby 3.2; `Object#=~` and the mailer's kwarg serialization changed since 3.1), so that's the environment, not the merge. CI's 3.1 run passes them.

### One thing I could not close, on #1750

Merging `main` in is what first ran CodeQL against that branch — the previous head has no CodeQL check — and it reports `js/xss-through-dom` (high) on the `location.assign` in `sheerid_iframe.js`. It's a false positive today (the value is `educator_profile_form_path`), and I added a guard so it fails closed if that path is ever made dynamic, but **CodeQL still flags it** and I stopped rather than keep reshaping Michael's code around the analyser. Laid out the two real options for him there. It's red on #1750 only; this PR is clean.

Still draft and still stacked — it merges into `main` cleanly once #1750 lands, no rebase needed. Re-requesting review. The upload-step Tab-through ask from the description still stands (first name `APPROVED`).

